PR #2, opened 2022-03-17
Refactor with Relayer and Chain interfaces. Add working test

agouin:
- Add Relayer interface
- Add Chain interface
- Pull docker images from heighliner and cosmos/relayer
- Implementing test between fresh gaia and osmosis chains for cosmos relayer


jtieri (approved):

jackzampolin (approved):

ibc/Relayer.go
```diff
@@ -0,0 +1,51 @@
+package ibc
```
adizere, line 1: At a high level the relayer interface looks appropriate to work with Hermes.

FYI we have E2E tests that we run between Hermes and 2 gaia instances here:
- https://github.com/informalsystems/ibc-rs/blob/master/ci/e2e.sh, which runs

- https://github.com/informalsystems/ibc-rs/blob/master/e2e/run.py 

These are in Bash/Python, however, but could be insightful to see how Hermes runs.

PR #7, opened 2022-04-05
Reproduce Juno Halt

agouin:
How to run tests:
```bash
# Build CLI
go build

# Chain halts at block 69. Contract states are different on every node
./ibc-test-framework test -s juno:v2.1.0 --src-chain-id juno-1 --src-vals 10 JunoHaltTest 

# Chain does not halt. Contract states are the same on all node
./ibc-test-framework test -s juno:v2.3.0 --src-chain-id juno-1 --src-vals 10 JunoHaltTest 
```


jackzampolin:
This is pretty amazing @agouin, maybe we should move this over into a `trophies` directory where we can keep one off things?

PR #21, opened 2022-04-18
Extract RelayerFactory interface

mark-rushakoff:
This provides more flexibility in supporting different relayer
configurations. It is the first step towards running relayer IBC tests
from a separate project.


agouin:
Looks great!

PR #22, opened 2022-04-18
Extract ChainFactory interface

mark-rushakoff:
And provide two implementations to cover the test and custom commands.

Most importantly, this will allow future tests to choose to start more
than two chains, to eventually cover a single relayer process handling
more than one path.

### DIFF
--- a/test/relayer_test.go
+++ b/test/relayer_test.go
@@ -9,47 +9,40 @@ import (
 
 // These tests are run by CI
 
-func getTestChains(t *testing.T) (ibc.Chain, ibc.Chain) {
-	numValidatorsPerChain := 4
-	numFullNodesPerChain := 1
-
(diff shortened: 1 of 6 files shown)


test/relayer_test.go
```diff
@@ -9,47 +9,40 @@ import (
...
-	numValidatorsPerChain := 4
-	numFullNodesPerChain := 1
-
-	srcChain, err := ibc.GetChain(t.Name(), "gaia", "v6.0.4", "cosmoshub-1004", numValidatorsPerChain, numFullNodesPerChain)
-	require.NoError(t, err)
-	dstChain, err := ibc.GetChain(t.Name(), "osmosis", "v7.0.4", "osmosis-1001", numValidatorsPerChain, numFullNodesPerChain)
-	require.NoError(t, err)
-
-	return srcChain, dstChain
+func getTestChainFactory(t *testing.T) ibc.ChainFactory {
```
agouin, line 12: Looks like `t` is no longer needed here

agouin (approved):
Good stuff!

PR #25, opened 2022-04-19
DRAFT: Support relayertest in existing cmd package

mark-rushakoff:
Move `IBCTestCase.RelayPacketTest` to `relayertest.TestRelayer_RelayPacket`, so that it can be directly invoked through the exported `relayertest.TestRelayer` function. This will allow running `relayertest.TestRelayer` from other unit tests (see change to `test/relayer_test.go` for example).

Retain the existing CLI functionality so that `go run . test RelayPacketTest` still works.

I'm going to add some inline comments to the PR for context.


ibc/relayerfactory.go
```diff
@@ -29,6 +29,10 @@ type builtinRelayerFactory struct {
 	impl RelayerImplementation
 }
 
+func NewBuiltinRelayerFactory(impl RelayerImplementation) RelayerFactory {
```
mark-rushakoff, line 32: This probably should have just been an exported function in a previous PR. Needed for using simple relayers in `relayertest`.

ibc/test_relay.go
```diff
@@ -7,174 +7,6 @@ import (
 	transfertypes "github.com/cosmos/ibc-go/v3/modules/apps/transfer/types"
 )
 
-func (ibc IBCTestCase) RelayPacketTest(testName string, cf ChainFactory, relayerImplementation RelayerImplementation) error {
```
mark-rushakoff, line 10: This function was effectively moved to `relayertest.TestRelayer_RelayPacket`.

ibc/test_setup.go
```diff
@@ -143,23 +143,23 @@ func StartChainsAndRelayerFromFactory(
 	}
 
 	if err := srcChain.Initialize(testName, home, pool, networkID); err != nil {
-		return errResponse(err)
+		return errResponse(fmt.Errorf("failed to initialize source chain: %w", err))
```
mark-rushakoff, line 146: Wrapping these errors was very helpful in debugging some self-inflicted test failures during development.

relayertest/legacy.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"runtime"
+	"strings"
+	"sync"
+
+	"github.com/strangelove-ventures/ibc-test-framework/ibc"
+)
+
+// GetLegacyTestCase looks up the given testCase and returns a function suitable for the existing cmd structure.
+func GetLegacyTestCase(testCase string) (func(string, ibc.ChainFactory, ibc.RelayerImplementation) error, error) {
```
mark-rushakoff, line 14: There is probably a better name for this.

This wraps the `TestRelayer_*` functions in the function signature that the existing commands expect. If we take the pure `go test` path, we won't need this file.

relayertest/legacy.go
```diff
@@ -0,0 +1,96 @@
...
+			name: testName,
+			cf:   cf,
+			rf:   ibc.NewBuiltinRelayerFactory(ri),
+		}
+		return w.run(testFunc)
+	}, nil
+}
+
+// tWrapper satisfies TestingT for use in a non-test binary.
+type tWrapper struct {
```
mark-rushakoff, line 34: I under-documented the `tWrapper` type for this draft. This acts like a `*testing.T` but can be used in a plain, non-test Go package.

relayertest/test.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func sanitizeTestNameForContainer(testName string) string {
+	return strings.ReplaceAll(testName, "/", "_")
+}
+
+// TestingT is a subset of *testing.T,
+// to support calling certain test-like functions
+// from a non-test binary.
+type TestingT interface {
```
mark-rushakoff, line 57: This interface should only be necessary if we continue in this direction to run tests in a non-test main.go. Otherwise we can fall back to a plain `*testing.T` instead.

relayertest/test.go
```diff
@@ -0,0 +1,227 @@
...
+	Name() string
+
+	// We do some diagnostic logging.
+	Logf(format string, args ...interface{})
+
+	// Use require for assertions and failures.
+	require.TestingT
+}
+
+func TestRelayer_RelayPacket(t TestingT, cf ibc.ChainFactory, rf ibc.RelayerFactory) {
```
mark-rushakoff, line 68: This was nearly a straight copy of `(ibc.IBCTestCase).RelayPacketTest`, but converted to work with a `TestingT` instead.

Previous calls to `fmt.Printf` are replaced with calls to `t.Logf`, and previous calls to `return err` are now `require.FailNow(t, err)`.

test/relayer_test.go
```diff
@@ -18,6 +19,10 @@ func getTestChainFactory() ibc.ChainFactory {
 	)
 }
 
+func TestRelayerByRelayerTest(t *testing.T) {
```
mark-rushakoff, line 22: This name is bad, but it was easy to use for filtering.

The structure of this test is identical to how a separate project, e.g. cosmos/relayer, could import and run the tests locally.

The intent is that the other project would pick a builtin chain factory and would provide its own relayer factory. The in-project tests are not intended to replace CI running a full suite of tests against many chains; they are merely a developer convenience.

I have changes to cosmos/relayer on my workstation that follow this pattern, but I'm not ready to push those changes until we update ibc-test-framework to allow it.

mark-rushakoff (commented):
I think I commented on all the parts where context would be particularly helpful.

ibc/test_relay.go
agouin: Why was only this function moved?

cmd/test.go
```diff
@@ -81,9 +82,14 @@ ibc-test-framework test -s osmosis:v7.0.4 -d juno:v2.3.0 -r rly RelayPacketTest
 		var testCases []func(testName string, cf ibc.ChainFactory, relayerImplementation ibc.RelayerImplementation) error
 
 		for _, testCaseString := range strings.Split(testCasesString, ",") {
-			testCase, err := ibc.GetTestCase(testCaseString)
+			// Prefer the new way of getting legacy test cases,
+			// but fall back to the old way for now.
+			testCase, err := relayertest.GetLegacyTestCase(testCaseString)
```
agouin, line 87: I think we should go one way or the other with this (GetLegacyTestCase or GetTestCase). I prefer the reflection so that the mapping does not need to be maintained. But maybe neither are necessary if we ditch the CLI and require the matrix files instead.

ibc/cosmos_relayer.go
```diff
@@ -335,3 +334,16 @@ func (relayer *CosmosRelayer) Bind() []string {
 func (relayer *CosmosRelayer) StopContainer() error {
 	return relayer.pool.Client.StopContainer(relayer.container.ID, uint(time.Second*30))
 }
+
+// condenseHostName truncates the middle of the given name
+// if it is 64 characters or longer.
+//
+// Without this helper, you may see an error like:
+//     API error (500): failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: sethostname: invalid argument: unknown
+func condenseHostName(name string) string {
```
agouin, line 343: awesome

relayertest/legacy.go
agouin: I think pure go test will be better, as long as the matrix file can support the current available chain configuration customization

PR #26, opened 2022-04-19
DRAFT: Add cmd/ibctest to provide a test binary

mark-rushakoff:
This builds on #25 to demonstrate how to provide a `go test`-built binary, as opposed to providing a standalone binary that kind of acts like `go test`.

I personally have a strong preference for this approach, as it would enable us to use the existing utilities around `go test` while still controlling runtime parameters of which relayer and which chains to run in test. For instance, when using the matrix configuration file, we can rely on the default `go test` parallelism to run as many tests as you have CPU cores; or if that proves too taxing on your system, you can provide the `-test.parallel` flag. 

This DRAFT change is being proposed as work on top of #25, but if we choose this route, a fair portion of the code added in #25 would be removed.

(original commit message follows:)

----
There would be a bit of remaining work to get this production-ready.

- Remove the existing cmd package?
- Port the remaining ibc.IBCTest tests to the relayertest package
- Add a Makefile to build the ibctest binary
- Fix the moniker-too-long error instead of working around it
- Support custom chain factories instead of just the builtins

### DIFF
--- a/cmd/ibctest/ibc_test.go
+++ b/cmd/ibctest/ibc_test.go
@@ -1,0 +1,168 @@
+// Command ibctest allows running the relayer tests with command-line configuration.
+package ibctest
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/strangelove-ventures/ibc-test-framework/ibc"
+	"github.com/strangelove-ventures/ibc-test-framework/relayertest"
+)
+
+// The value of the extra flags this test supports.
+var mainFlags struct {
+	MatrixFile string
+}
+
+// The value of the test matrix.
+var testMatrix struct {
+	Relayers []string
+
+	ChainSets [][]ibc.BuiltinChainFactoryEntry
+
+	// TODO: support a slice of ibc.CustomChainFactoryEntry too.
+}
+
+func TestMain(m *testing.M) {
+	addFlags()
+	flag.Parse()
+
+	if err := setUpTestMatrix(); err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to build test matrix: %v\n", err)
+		os.Exit(1)
+	}
+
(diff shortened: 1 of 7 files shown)


cmd/ibctest/ibc_test.go
```diff
@@ -0,0 +1,168 @@
...
+				chainNames := make([]string, len(cs))
+				for i, c := range cs {
+					// TODO: this shouldn't need to truncate bits of the name and version,
+					// but without this, the tests hang on a failed container running gentx, with this error:
+					//    panic: failed to execute message; message index: 0: invalid moniker length; got: 75, max: 70
+					// It would be better to figure out the correct gentx invocation
+					// rather than hiding details from the test name.
+					chainNames[i] = c.Name[:3] + "@" + c.Version[1:]
+				}
+				chainTestName := strings.Join(chainNames, "+")
```
agouin, line 148: with only 2 chains this should be fine, but I have noticed the docker containers can fail to run if the testname is too long since it is used in the hostname of the containers (max length 63), so we may want to truncate this value and add some other randomized string

DavidNix:
> I personally have a strong preference for this approach

+1

I think this is a great idea to produce a test binary to leverage all of Go's builtin test features. 

I pulled this branch down and tried it. Worked great! 

My only question is related to the community. Can we break existing clients in favor of this new, test binary command line interface? I don't think we'd want to support both paths - test binary vs. the old way. So it would be nice to deprecate and delete the old way. 

Great work here.

agouin:
> I personally have a strong preference for this approach

I also agree.

> My only question is related to the community. Can we break existing clients in favor of this new, test binary command line interface? I don't think we'd want to support both paths - test binary vs. the old way. So it would be nice to deprecate and delete the old way.

This framework is brand new, so I am in favor of ditching the existing CLI for this approach to avoid things like the hardcoded mapping of test name to test function.

We can build out the matrix file or test flags to support everything that the current `test custom` command supports and we shouldn't lose any functionality.

PR #28, opened 2022-04-20
Produce ibctest binary built with go test

mark-rushakoff:
This supersedes #25 and #26.

Removes the existing cmd package and top-level main.go, in favor of cmd/ibctest, so that the relayer tests can be packages as an executable built with go test. That binary supports a `-matrix` argument, which points to a JSON file specifying which relayers to test, and which chain configurations to test as well.

There are likely a few remaining rough edges internally, but this is a broad enough change that I think it can be merged now and the rough edges smoothed out as followup work.

There may be some issues with hostnames, as I was locally testing with 1 node and 1 validator, so I will revisit that later today.


test/relayer_test.go
```diff
@@ -4,45 +4,20 @@ import (
...
 )
 
 // These tests are run by CI
 
 func getTestChainFactory() ibc.ChainFactory {
 	return ibc.NewBuiltinChainFactory(
 		[]ibc.BuiltinChainFactoryEntry{
-			{Name: "gaia", Version: "v6.0.4", ChainID: "cosmoshub-1004", NumValidators: 4, NumFullNodes: 1},
-			{Name: "osmosis", Version: "v7.0.4", ChainID: "osmosis-1001", NumValidators: 4, NumFullNodes: 1},
+			{Name: "gaia", Version: "v6.0.4", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
```
agouin, line 15: Can we keep these at 4?

DavidNix:
LGTM! 🚢

agouin (approved):

jackzampolin:
Ran these locally and tests passing v nice!

PR #32, opened 2022-04-20
[DO NOT MERGE] Reorganize into modules

agouin:
Will rebase after the merge of #28


ibctest/test_chains.go
```diff
@@ -0,0 +1,39 @@
...
+	chainConfigMap = make(map[string]ibc.ChainConfig)
+	for _, chainConfig := range chainConfigs {
+		chainConfigMap[chainConfig.Name] = chainConfig
+	}
+}
+
+func GetChain(testName, name, version, chainID string, numValidators, numFullNodes int) (ibc.Chain, error) {
+	chainConfig, exists := chainConfigMap[name]
+	if !exists {
+		return nil, fmt.Errorf("No chain configuration for %s", name)
```
mark-rushakoff, line 28: Error strings should begin with a lowercase letter.

ibctest/test_setup.go
```diff
@@ -39,15 +31,15 @@ const (
 type IBCTestCase struct{}
 
 // uses reflection to get test case
-func GetTestCase(testCase string) (func(testName string, cf ChainFactory, relayerImplementation RelayerImplementation) error, error) {
+func GetTestCase(testCase string) (func(testName string, cf ChainFactory, relayerImplementation ibc.RelayerImplementation) error, error) {
```
mark-rushakoff, line 34: I think this function is unused now; I should have deleted it in a previous change.

mark-rushakoff (approved):
Nice cleanup. 

I would maybe rename the "cosmos" package to "cosmoschain", but since that is the only package name cosmos it is probably fine. It gets more confusing when there are multiple packages with the same name like that. 

The utils package, I would probably rename to docker or dockerutils. Go tends to avoid package names like utils or tools, instead preferring names reflecting more specific purposes.

mark-rushakoff:
I have some changes queued up that depend on this, so I'm merging the PR now. The small naming suggestions are okay to happen separately.

DavidNix (commented):
Nice cleanup! I'm too new to this project and don't have enough context, but sometimes too many packages can bite you back because Go does not allow cyclic imports. Luckily, we can refactor ourselves out of that mess if it becomes a problem.

I really like the separate `chain` packages.

utils/docker.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return net.JoinHostPort(ip, m[0].HostPort)
+}
+
+// RandLowerCaseLetterString returns a lowercase letter string of given length
+func RandLowerCaseLetterString(length int) string {
+	chars := []rune("abcdefghijklmnopqrstuvwxyz")
+	var b strings.Builder
+	for i := 0; i < length; i++ {
+		i, _ := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
```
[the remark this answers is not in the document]
mark-rushakoff: We probably don't -- I wouldn't say I've noticed consistent suffixes, so either it is getting seeded by a third party package, or there is enough nondeterminism in call order that they are different each time. 

If we change this part of the code, we can construct a rune or byte slice and call with `rand.Intn` instead of using a string builder and `math/big`.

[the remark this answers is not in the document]
DavidNix: I agree. The `math/big` was a surprise to me.

ibctest/test_chains.go
DavidNix: A linter can catch this in the future.

utils/docker.go
```diff
@@ -0,0 +1,80 @@
+package utils
```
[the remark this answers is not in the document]
agouin: In the penumbra PR #33 , I added another util file `file.go` for a file `Copy` helper. So maybe it should be `utils/docker` with one file `dockerutils.go` with package name `dockerutils`, and then `utils/file` with one file `fileutils.go` with package name `fileutils`? cc @mark-rushakoff since he brought up the same concern.

utils/docker.go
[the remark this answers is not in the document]
agouin: Feel free to clean this up if necessary. This has been carried forward from it's initial use in horcrux about 9 months ago.

ibctest/test_chains.go
agouin: the `golangci-lint` linter we are using catches the errors.New() capital first letters but not the fmt.Errorf. Maybe we can configure it to, or maybe we should change the linter or add an additional one?

mark-rushakoff: https://staticcheck.io/ has been the canonical linter in my experience.

PR #33, opened 2022-04-21
Add penumbra chain impl

### LINKED ISSUE #24: Add penumbra chain
https://github.com/strangelove-ventures/heighliner/pkgs/container/heighliner%2Fpenumbra
https://github.com/penumbra-zone/penumbra
https://guide.penumbra.zone/main/pd.html

agouin:
Chain start test is functional. Tests that a chain of 4 validators can start and produce blocks
```bash
go test -v -run TestPenumbraChainStart ./...
```

`PenumbraNode` (validator or fullnode) is composed of a `TendermintNode` for consensus and a `PenumbraAppNode` for the ABCI application, using the tendermint abci proxy between the two.

tendermint v0.35.0 had enough differences from the current `ChainNode` struct that I decided to break `TendermintNode` out into its own module.

For penumbra, I am currently creating the full genesis file by marshaling a test `PenumbraGenesisFile`, as opposed to the gentx flow that is used for the cosmos chain.

Since penumbra requires a version for both penumbra and tendermint, I changed the `ChainConfig` struct to have an array of `ChainDockerImage`s instead of a single `Repository` and `Version`. This can also be used in the future for other things like upgrade testing.

Penumbra does not yet have IBC fully functional, so testing with penumbra with the relayer will currently fail. This gives a way to test penumbra IBC as it matures. The tests will need to expand beyond ICS-20 also.

Resolves #24

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -48,6 +48,7 @@ type ChainNode struct {
 	Client       rpcclient.Client
 	Container    *docker.Container
 	TestName     string
+	Image        ibc.ChainDockerImage
 }
 
 // ChainNodes is a collection of ChainNode
@@ -602,7 +603,7 @@ func (tn *ChainNode) CreateNodeContainer() error {
 			Hostname:     tn.HostName(),
 			ExposedPorts: sentryPorts,
 			DNS:          []string{},
-			Image:        fmt.Sprintf("%s:%s", chainCfg.Repository, chainCfg.Version),
+			Image:        fmt.Sprintf("%s:%s", tn.Image.Repository, tn.Image.Version),
 			Labels:       map[string]string{"ibc-test": tn.TestName},
 		},
 		HostConfig: &docker.HostConfig{
@@ -773,7 +774,6 @@ func (tn *ChainNode) NodeJob(ctx context.Context, cmd []string) (int, string, st
 	funcName := strings.Split(caller, ".")
 	container := fmt.Sprintf("%s-%s-%s", tn.Name(), funcName[len(funcName)-1], dockerutil.RandLowerCaseLetterString(3))
 	fmt.Printf("{%s} -> '%s'\n", container, strings.Join(cmd, " "))
-	chainCfg := tn.Chain.Config()
 	cont, err := tn.Pool.Client.CreateContainer(docker.CreateContainerOptions{
 		Name: container,
 		Config: &docker.Config{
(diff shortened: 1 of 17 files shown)


chain/penumbra/penumbra_app_node.go
```diff
@@ -0,0 +1,219 @@
...
+		return []byte{}, utils.HandleNodeJobError(exitCode, stdout, stderr, err)
+	}
+	addresses := strings.Split(stdout, "\n")
+	for _, address := range addresses {
+		fields := strings.Fields(address)
+		if len(fields) < 3 {
+			continue
+		}
+		if fields[1] == keyName {
+			// TODO looks like penumbra address is bech64? need to decode to bytes here
```
redshiftzero, line 86: Penumbra addresses use Bech32m encoding: https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/address.rs#L156 (a bit more detail here on the encoding in case it's useful: https://protocol.penumbra.zone/main/protocol/addresses_keys/addresses.html#address-encodings)

agouin: thanks! looks like I don't need to decode it yet actually to get the chain moving, but updated the TODO message for later when it's needed for IBC testing.

dockerutil/file.go
```diff
@@ -0,0 +1,32 @@
+package dockerutil
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+func Copy(src, dst string) (int64, error) {
```
DavidNix, line 9: nitpick:

Maybe `CopyFile` as an alternative name?

Without digging into the implementation, it was unclear to me what we were copying.

DavidNix (approved):
Other than a couple nitpicks, LGTM. Pretty neat to have a non-cosmos chain now!

chain/cosmos/chain_node.go
```diff
@@ -116,14 +116,19 @@ func (tn *ChainNode) CliContext() client.Context {
...
+
+// hostname of the test node container
+func (tn *ChainNode) HostName() string {
+	return dockerutil.CondenseHostName(tn.Name())
 }
 
 // Dir is the directory where the test node files are stored
 func (tn *ChainNode) Dir() string {
-	return fmt.Sprintf("%s/%s/", tn.Home, tn.Name())
+	return path.Join(tn.Home, tn.Name())
```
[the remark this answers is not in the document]
agouin: good catch. replaced throughout the project.

dockerutil/file.go
agouin: done

PR #34, opened 2022-04-21
Avoid ioutil package

mark-rushakoff:
The io/ioutil package has been deprecated in favor of the existing os
and io packages. There should be no functional change by switching these
usages to the os package.


DavidNix:
👏

PR #37, opened 2022-04-21
refactor: Update gitignore for macos and IDEs

DavidNix:


mark-rushakoff (approved):
I don't like having the wide list of system- and application-specific files in .gitignore -- those belong in [your global .gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).

Having a wide list committed to .gitignore encourages people to expand the list over time, instead of updating their personal global .gitignore. OTOH, GitHub's PR review interface makes it very easy to miss committed binary files.

WDYT about keeping the list, but adding a noticeable comment suggesting that this list is frozen, and linking to the global .gitignore docs?

DavidNix:
> I don't like having the wide list of system- and application-specific files in .gitignore -- those belong in [your global .gitignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).
> 
> Having a wide list committed to .gitignore encourages people to expand the list over time, instead of updating their personal global .gitignore. OTOH, GitHub's PR review interface makes it very easy to miss committed binary files.
> 
> WDYT about keeping the list, but adding a noticeable comment suggesting that this list is frozen, and linking to the global .gitignore docs?

@mark-rushakoff TIL! Thank you. I'm all for using global .gitignore. I'll close this PR.

PR #38, opened 2022-04-21
Make more code aware of *testing.T

mark-rushakoff:
In places where we were passing (*testing.T).Name(), prefer to pass the
plain *testing.T value. This makes the interfaces more amenable to
external definitions that are also used specifically in Go tests.

This allows use of t.Cleanup rather than returning callbacks that need
to be deferred.

When we need to sanitize test names for containers, do it later in the
call chains.


agouin (approved):

agouin:
I had changed this so the tests could be used from both CLI and go test before, but now that we're aligned on go test, good to have this cleanup

PR #39, opened 2022-04-21
Support non-docker relayers

mark-rushakoff:
For non-docker relayers, we need to provide the host-bound rpc and gRPC
addresses. Those addresses cannot be determined until the chain
containers are up and running. The relayer does not have an existing API
to modify a chain's addresses, so delay relayer configuration until
after chains are running.

However, we need certain accounts set up to perform chain genesis, so
generate those keys out of band, provide the account details to genesis,
and then restore the keys into a new relayer setup.


DavidNix (approved):
🎉

PR #40, opened 2022-04-21
refactor: Misc cleanup + unit test coverage

DavidNix:

* Rename `utils` -> `dockerutil` to clarify the intention of that package. Seems like a safe name given every function in there is to help with docker related things. https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common
* Add unit test coverage for above package
* Update Makefile for unit tests

- NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,13 @@
+default: help
+
+.PHONY: help
+help: ## Print this help message
+	@echo "Available make commands:"; grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
 .PHONY: ibctest
-ibctest:
+ibctest: ## Build ibctest binary into ./bin
 	go test -c -o ./bin/ibctest ./cmd/ibctest
+
+.PHONY: test
+test: ## Run unit tests
+	@go test -cover -short -race -timeout=60s $(shell go list ./... | grep -v /cmd/)
--- a/dockerutil/errors.go
+++ b/dockerutil/errors.go
@@ -2,7 +2,7 @@ package dockerutil
 
 import "fmt"
 
-func HandleNodeJobError(exitCode int, stdout, stderr string, err error) error {
+func HandleNodeJobError(exitCode int, _, _ string, err error) error {
 	if err != nil {
 		return err
 	}
--- a/dockerutil/errors.go
+++ b/dockerutil/errors.go
@@ -1,0 +1,13 @@
+package dockerutil
+
+import "fmt"
+
+func HandleNodeJobError(exitCode int, stdout, stderr string, err error) error {
+	if err != nil {
+		return err
+	}
+	if exitCode != 0 {
+		return fmt.Errorf("container returned non-zero error code: %d\n", exitCode)
+	}
+	return nil
+}
--- a/dockerutil/strings.go
+++ b/dockerutil/strings.go
@@ -2,12 +2,13 @@ package dockerutil
 
 import (
 	"fmt"
-	"github.com/ory/dockertest/docker"
 	"math/rand"
 	"net"
 	"os"
 	"regexp"
 	"runtime"
+
+	"github.com/ory/dockertest/docker"
 )
 
 // GetHostPort returns a resource's published port with an address.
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -1,3 +1,13 @@
+default: help
+
+.PHONY: help
+help: ## Print this help message
+	@echo "Available make commands:"; grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
```
DavidNix, line 6: Boilerplate I often add to Makefiles to print a nice help message. 
Screen Shot 2022-04-21 at 4 48 56 PM

dockerutil/strings.go
```diff
@@ -31,25 +28,15 @@ func GetHostPort(cont *docker.Container, portID string) string {
...
 }
 
+var chars = []rune("abcdefghijklmnopqrstuvwxyz")
+
 // RandLowerCaseLetterString returns a lowercase letter string of given length
 func RandLowerCaseLetterString(length int) string {
-	chars := []rune("abcdefghijklmnopqrstuvwxyz")
-	var b strings.Builder
-	for i := 0; i < length; i++ {
-		i, _ := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
```
DavidNix, line 39: This was using `crypto/rand` previously. It wasn't wrong. I just thought `math/rand` was a little cleaner.

DavidNix:
Moving back to draft while I investigate the CI failure

DavidNix:
Please correct me if any of my terminology is off!

test/relayer_test.go
```diff
@@ -20,5 +20,8 @@ func getTestChainFactory() ibctest.ChainFactory {
 }
 
 func TestRelayer(t *testing.T) {
+	if testing.Short() {
+		return
```
mark-rushakoff, line 24: The pattern in the Go standard library is to use t.Skip to mark a short test as skipped, as opposed to a return which counts as a success.

dockerutil/errors.go
```diff
@@ -0,0 +1,13 @@
+package dockerutil
+
+import "fmt"
+
+func HandleNodeJobError(exitCode int, _, _ string, err error) error {
+	if err != nil {
+		return err
+	}
+	if exitCode != 0 {
+		return fmt.Errorf("container returned non-zero error code: %d\n", exitCode)
```
mark-rushakoff, line 10: The two unnamed string arguments are standard output and error, so those should be part of the formatted error message. Or maybe use a new structured error for this, in case we need to introspect the error somewhere else.

dockerutil/strings.go
```diff
@@ -31,40 +29,29 @@ func GetHostPort(cont *docker.Container, portID string) string {
...
-	return b.String()
-}
-
-func HandleNodeJobError(exitCode int, stdout, stderr string, err error) error {
-	if err != nil {
-		return err
-	}
-	if exitCode != 0 {
-		return fmt.Errorf("container returned non-zero error code: %d\n", exitCode)
+	b := make([]rune, length)
```
mark-rushakoff, line 36: Since the lowercase letters are all plain ascii, this and the chars variable could be a byte slice rather than a rune slice.

mark-rushakoff (approved):
A couple small comments but nothing major. I would expect tests to pass with these changes, so maybe just rerun them.

dockerutil/errors.go
DavidNix: This function was copied. I split it into a separate file `errors.go` to separate from the strings helpers. I agree a function with unused args is suspect. In this case, it's a convenience given the return signature in other areas of the codebase. 

It probably warrants a refactor but I would argue out of scope for this PR.

test/relayer_test.go
DavidNix: Ah, good call.

DavidNix:
Changing to "Ready for Review" since @mark-rushakoff confirmed the test failure is intermittent. Hoping a re-run gets it to pass.

dockerutil/errors.go
agouin: The NodeJob method returns stdout and stderr so that command callers can use them if necessary to parse output. Execute-only commands that don't need to parse the output can simply wrap the NodeJob call with this method to get an error return. But I don't think that stderr (or stdout) should be printed here necessarily. I had it printing stderr when err != nil and it made the test very hard to follow since some containers have extremely verbose stderr. But maybe printing stderr for that case is fine now that the long running chain node container output is disabled unless `SHOW_CONTAINER_LOGS` env var is provided. 

> It probably warrants a refactor but I would argue out of scope for this PR.

+1

PR #42, opened 2022-04-22
Omit src and dst chains from RelayerFactory.Build

mark-rushakoff:
First, these are not used by any existing RelayerFactory implementation.
Second, coupling relayer instances to a pair of chains will
overcomplicate upcoming many-chain relayer tests.


DavidNix (approved):
LGTM!

PR #44, opened 2022-04-22
refactor: Use v3 import path for dockertest package

DavidNix:

Adhere to Go best practices. https://go.dev/doc/modules/major-version

Effectivly updates https://github.com/ory/dockertest v3.3.5 -> v3.8.1

* 🤞 v3.8.1 does not break anything. CI tests should tell us. If it breaks, will downgrade to v3.3.5.
* Perhaps there is a reason I'm not aware of as to why we are not using the major version in the import path?


mark-rushakoff (approved):
I would expect this to be a seamless upgrade.

DavidNix:
> I would expect this to be a seamless upgrade.

One would hope!

PR #45, opened 2022-04-22
Fix indexing into alphabet for random characters

mark-rushakoff:
I kept seeing test with suffixes only composed of characters a, b, and
c. This was because it was only using the requested length for indexing
into the alphabet, and those tests were requesting a random string of
length 3.


DavidNix (approved):
Great catch!

PR #48, opened 2022-04-22
Make ibctest aware of relayer capabilities

mark-rushakoff:
Rather than having hardcoded t.Skip calls in tests, allow the
RelayerFactory instance to report which relayer features it supports,
and use that information to decide which tests to mark as skipped.


tools.go
```diff
@@ -0,0 +1,11 @@
+//go:build tools
+
+// This pattern of a file named tools.go, to import dependent tools,
+// comes from the official Go wiki:
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+
+package ibctestframework
+
+import (
+	_ "golang.org/x/tools/cmd/stringer"
```
DavidNix, line 10: +1 for stringer

DavidNix (approved):

relayer/capability.go
```diff
@@ -0,0 +1,28 @@
+package relayer
+
+//go:generate go run golang.org/x/tools/cmd/stringer -type=Capability
```
jtieri, line 3: First time I'm seeing stringer before. This is great!!

jtieri (approved):
This looks great 👍

PR #50, opened 2022-04-25
feat: Test acknolwedgements

DavidNix:

https://github.com/strangelove-ventures/ibc-test-framework/issues/19

TODO

TODO


chain/cosmos/chain_node.go
```diff
@@ -653,7 +653,7 @@ func (tn *ChainNode) CreateNodeContainer() error {
 }
 
 func (tn *ChainNode) StopContainer() error {
-	return tn.Pool.Client.StopContainer(tn.Container.ID, uint(time.Second*30))
```
DavidNix, line 656: Documentation of the 3rd party package says this unit of time is already seconds. 

Not the best package API. I prefer `time.Duration` for time arguments.

chain/cosmos/cosmos_chain.go
```diff
@@ -36,27 +36,6 @@ type CosmosChain struct {
 	ChainNodes    ChainNodes
 }
 
-func NewCosmosChainConfig(name string,
```
DavidNix, line 39: Dead code. Probably worth running a linter to find and delete unused code at some future point.

ibctest/test_setup.go
```diff
@@ -271,13 +271,13 @@ func StartChainsAndRelayerFromFactory(
 
 func WaitForBlocks(srcChain, dstChain ibc.Chain, blocksToWait int64) error {
 	chainsConsecutiveBlocksWaitGroup := errgroup.Group{}
-	chainsConsecutiveBlocksWaitGroup.Go(func() (err error) {
```
DavidNix, line 274: I personally always avoid naked returns. Why? https://levelup.gitconnected.com/go-naked-returns-4e2094b598e6

In past projects, I typically run a linter on CI that prohibts naked returns entirely.

In this case, I'm being pedantic.

chain/cosmos/cosmos_chain.go
agouin: this was used in the `test custom` command of the previous CLI. It would be nice to make this functional in the matrix file

PR #51, opened 2022-04-25
Wait for chains to produce more blocks in relayer tests

mark-rushakoff:
Ten blocks ought to be sufficient, but we are seeing test failures that
appear to stem from internal retries that don't finish quickly. Those
retries will be addressed separately in the relayer repository. In the
meantime, waiting for twenty blocks makes the tests a bit more forgiving
and reduces the frequency of CI retries.


DavidNix (approved):
🚢

PR #53, opened 2022-04-27
Better docker cleanup

agouin:
- Run cleanup before and after
- Fix missing cleanup for rly docker `NodeJob`s
- Use latest images for example matrix. Keep fixed versions for CI

### DIFF
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -108,12 +108,12 @@ func (c *CosmosChain) getRelayerNode() *ChainNode {
 
 // Implements Chain interface
 func (c *CosmosChain) GetRPCAddress() string {
-	return fmt.Sprintf("http://%s:26657", c.getRelayerNode().Name())
+	return fmt.Sprintf("http://%s:26657", dockerutil.CondenseHostName(c.getRelayerNode().Name()))
 }
 
 // Implements Chain interface
 func (c *CosmosChain) GetGRPCAddress() string {
-	return fmt.Sprintf("%s:9090", c.getRelayerNode().Name())
+	return fmt.Sprintf("%s:9090", dockerutil.CondenseHostName(c.getRelayerNode().Name()))
 }
 
 // GetHostRPCAddress returns the address of the RPC server accessible by the host.
--- a/ibctest/test_setup.go
+++ b/ibctest/test_setup.go
@@ -33,7 +33,12 @@ func SetupTestRun(t *testing.T) (context.Context, string, *dockertest.Pool, stri
 	if err != nil {
 		return ctx, "", nil, "", err
 	}
+
+	// schedule cleanup for afterwards
 	t.Cleanup(dockerCleanup(t.Name(), pool))
+
+	// run cleanup for this test first in case previous run was killed (e.g. Ctrl+C)
+	dockerCleanup(t.Name(), pool)()
 
 	home := t.TempDir()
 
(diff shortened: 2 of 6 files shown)


test/relayer_test.go
```diff
@@ -13,8 +13,8 @@ import (
 func getTestChainFactory() ibctest.ChainFactory {
 	return ibctest.NewBuiltinChainFactory(
 		[]ibctest.BuiltinChainFactoryEntry{
-			{Name: "gaia", Version: "v6.0.4", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
-			{Name: "osmosis", Version: "v7.0.4", ChainID: "osmosis-1001", NumValidators: 1, NumFullNodes: 1},
+			{Name: "gaia", Version: "latest", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
```
mark-rushakoff, line 16: I like `latest` for the example matrix file and for the default behavior of the `ibctest` binary we ship, but I don't like the idea that the tests for the same commit of `ibc-test-framework` could change behavior based on an external docker image changing.

The tradeoff, of course, is that we have to update ibc-test-framework whenever those images change. That can be mostly automated with a little effort.

mark-rushakoff (commented):
I would prefer to not use `latest` docker images in the tests that are run as part of testing ibc-test-framework itself, but if anyone on @strangelove-ventures/ibc-test-maintainers wants to override me on that, everything else in the PR looks good to me.

test/relayer_test.go
DavidNix: +1

ibctest/test_setup.go
```diff
@@ -33,8 +33,13 @@ func SetupTestRun(t *testing.T) (context.Context, string, *dockertest.Pool, stri
 	if err != nil {
 		return ctx, "", nil, "", err
 	}
+
+	// schedule cleanup for afterwards
 	t.Cleanup(dockerCleanup(t.Name(), pool))
 
+	// run cleanup for this test first in case previous run was killed (e.g. Ctrl+C)
+	dockerCleanup(t.Name(), pool)()
```
DavidNix, line 41: I'm unsure the implications here but worth a shot. Easy enough to revert.

DavidNix (approved):
🎉 

With the caveat, we may want to consider using a known stable docker tag for the default chain config as Mark suggested.

test/relayer_test.go
agouin: works for me, then it's just one place to maintain instead of 3

ibctest/test_setup.go
agouin: Just means if you killed a running test with Ctrl+C, so cleanup does not happen, then run the test again, it will run the cleanup before starting so that you don't run into `container already exists` errors.

chain/cosmos/cosmos_chain.go
```diff
@@ -108,12 +108,12 @@ func (c *CosmosChain) getRelayerNode() *ChainNode {
 
 // Implements Chain interface
 func (c *CosmosChain) GetRPCAddress() string {
-	return fmt.Sprintf("http://%s:26657", c.getRelayerNode().Name())
+	return fmt.Sprintf("http://%s:26657", dockerutil.CondenseHostName(c.getRelayerNode().Name()))
```
agouin, line 111: I think we might keep running into this when we want the hostname. I'll put up a PR of a suggestion.

PR #54, opened 2022-04-27
WIP: add outline of 'from scratch' test

mark-rushakoff:
A root test that shares chains and eventual relayer client state through
subtests will be helpful in onboarding new chains to IBC, and it will
also detect potential flakiness in any early relayer setup.

This is WIP for now because I did not have time to extract helper
functions and a likely helper struct to contain most of this setup and
expose appropriate data to subtests, but this is illustrative of the
general outline on how I would approach this.

### DIFF
--- a/relayertest/fromscratch.go
+++ b/relayertest/fromscratch.go
@@ -1,0 +1,183 @@
+package relayertest
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
Superseded by #110.

PR #55, opened 2022-04-28
hostname

agouin:
Make sure hostname is used for all node dials


agouin:
Part of me thinks we should just have `Name()` be the hostname, and have it be sanitized/condensed for both the moniker and hostname, because I don't see a big reason for them to be separate, but this continues with the separate approach for moniker/hostname.

PR #57, opened 2022-04-29
refactor: Misc cleanup

DavidNix:


chain/cosmos/cosmos_chain.go
```diff
@@ -228,7 +228,7 @@ func (c *CosmosChain) GetGasFeesInNativeDenom(gasPaid int64) int64 {
 // creates the test node objects required for bootstrapping tests
 func (c *CosmosChain) initializeChainNodes(testName, home string,
 	pool *dockertest.Pool, networkID string) {
-	ChainNodes := []*ChainNode{}
```
DavidNix, line 231: Shadows existing type.

chain/cosmos/chain_node.go
```diff
@@ -625,7 +625,7 @@ func (tn *ChainNode) CreateNodeContainer() error {
 }
 
 func (tn *ChainNode) StopContainer() error {
-	return tn.Pool.Client.StopContainer(tn.Container.ID, uint(time.Second*30))
+	return tn.Pool.Client.StopContainer(tn.Container.ID, 30)
```
DavidNix, line 628: Unit of time is already seconds. This 3rd party package made an interesting choice. For time args, I almost always use `time.Duration`.

ibctest/test_setup.go
```diff
@@ -276,13 +276,13 @@ func StartChainsAndRelayerFromFactory(
 
 func WaitForBlocks(srcChain, dstChain ibc.Chain, blocksToWait int64) error {
 	chainsConsecutiveBlocksWaitGroup := errgroup.Group{}
-	chainsConsecutiveBlocksWaitGroup.Go(func() (err error) {
-		_, err = srcChain.WaitForBlocks(blocksToWait)
-		return
+	chainsConsecutiveBlocksWaitGroup.Go(func() error {
+		_, err := srcChain.WaitForBlocks(blocksToWait)
+		return err
```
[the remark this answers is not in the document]
mark-rushakoff: I'd be okay with rejecting that in a linter. We can opt out of the rule in exceptional circumstances.

mark-rushakoff (approved):
All good changes, thanks for the cleanup.

agouin (approved):
Nice cleanup!

chain/cosmos/chain_node.go
agouin: Great catch. We should fix this at the upstream too. https://github.com/strangelove-ventures/horcrux/pull/76

relayer/rly/cosmos_relayer.go
```diff
@@ -241,10 +241,7 @@ func (relayer *CosmosRelayer) CreateNodeContainer(pathName string) error {
 		return err
 	}
 	relayer.container = cont
-	if err := relayer.pool.Client.StartContainer(relayer.container.ID, nil); err != nil {
-		return err
```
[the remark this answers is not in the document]
DavidNix: +1 for wrapping errors

Off topic: What are your thoughts on https://github.com/pkg/errors? I heavily rely on that package, but I now see it's read-only. 😢 

The new-ish `%w` is fine but I like how Cheney's package captures stack traces and you can optionally print them.

PR #60, opened 2022-05-02
WIP: feat: Add structured logging and optional log to files

DavidNix:

The relayer and chains emit tons of logs which are mixed in with the Go test framwork logs. Thus, it's often difficult to zero-in on test failures.

Multiple chain/relayer combinations, given the parallel tests, may also mix together. 

Additionally, the logs are not structured, making it difficult to parse and search for specifics using tools which expect structured logging. 

The PR adds the option to save chain combination logs to separate files under `~/.ibctest`. The user may still choose stdout/stderr if they wish. 

I made log files the default but am happy to revert that to stderr (the previous default).

* This does not relate to the `testing.Log`s which may be confusing.


log/iface.go
```diff
@@ -0,0 +1,17 @@
+package log
+
+var _ Logger = logger{}
+
+// Logger is a convenience interface
+type Logger interface {
```
DavidNix, line 6: I typically do not like packages defining and exposing exported interfaces. 

Go interfaces let you invert the dependency (as it should be), such that the caller gets to define what it wants. 

However, a `log` package is an exception to the above for the convenience of callers asking for a `log.Logger` vs. forcing them to define an interface each time.

test/relayer_test.go
```diff
@@ -26,5 +27,15 @@ func TestRelayer(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
 	}
-	relayertest.TestRelayer(t, getTestChainFactory(), ibctest.NewBuiltinRelayerFactory(ibc.CosmosRly))
```
DavidNix, line 29: I'm wondering the appetite for deleting this test entirely? It's duplicative. The `cmd/ibctest_test.go` seems like it should be the canonical path.

DavidNix: I'll leave it in this PR. I think we may use it to ensure it's run as part of the CI test suite? As an option, we can workaround that and funnel all e2e tests through the ibctest binary.

chain/cosmos/cosmos_chain.go
```diff
@@ -34,31 +35,8 @@ type CosmosChain struct {
 	numValidators int
 	numFullNodes  int
 	ChainNodes    ChainNodes
-}
 
-func NewCosmosChainConfig(name string,
```
DavidNix, line 39: dead code

test/relayer_test.go
agouin: I'm open to removing it and using the ibctest binary in CI

agouin (approved):
Great enhancement for speeding up root cause analysis from a test failure.

test/relayer_test.go
DavidNix: Great! I'll add ti to the list.

PR #64, opened 2022-05-09
feat: Assert IBC packet acknowledgments

DavidNix:

Attempt to make meaningful assertions against IBC transactions and the followup acknowledgement transaction.

## Addiionally

When `debug` logs activated, prints tendermint block details for cosmos chains.

* Bloats the `ibc.Chain` interface even more. 
* `WaitForBlocks` may be able to move toward `WaitForAck(seq uint64)` or similar. That may be explored in a future PR.

### DIFF
--- a/relayertest/test.go
+++ b/relayertest/test.go
@@ -141,24 +141,31 @@ func sendIBCTransfersFromBothChainsWithTimeout(
 		Amount:  testCoinAmount,
 	}
 
-	var eg errgroup.Group
-	var txHashSrc string
-	var txHashDst string
+	var (
+		eg        errgroup.Group
+		txHashSrc string
+		txHashDst string
+
+		srcChannelID = channels[0].ChannelID
+		dstChannelID = channels[0].Counterparty.ChannelID
+	)
 
 	eg.Go(func() error {
 		var err error
-		txHashSrc, err = srcChain.SendIBCTransfer(ctx, channels[0].ChannelID, srcUser.KeyName, testCoinSrcToDst, timeout)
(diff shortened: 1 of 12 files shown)


jackzampolin:
I like the `WaitForPacket` and `WaitForAck` idea and don't have a problem adding it to the chain interface.

relayertest/test.go
```diff
@@ -292,9 +295,18 @@ func testPacketRelaySuccess(
 
 	require.Equal(t, srcInitialBalance-expectedDifference, srcFinalBalance)
 	require.Equal(t, dstInitialBalance+testCoinAmount, dstFinalBalance)
+
+	seq, err := srcChain.GetPacketSequence(ctx, srcTxHash)
```
agouin, line 299: maybe `GetPacketSequence` can be changed to take a *types.TxResponse instead of a hash? For all the uses of it, it looks like the transaction has already been fetched with `Chain.GetTransaction`, so we can avoid the double query.

agouin (approved):
Looks great! We will want to assert in `testPacketRelayFail` also, but that can probably be handled in a separate PR

relayertest/test.go
DavidNix: The double query is a good point. I was attempting to keep the interface methods as generic as possible without leaking *types.TxResponse because it is tendermint/cosmos specific. It's perhaps pre-optimization on my part thinking about non-cosmos chains. 

But in a future PR, I'll explore using a type we define in our `ibc` package to limit the queries.

agouin: Yes, the `*types.TxResponse` return type for the `Chain.GetTransaction` should eventually be replaced with our own generic type, or possibly removed and replaced with more specific queries if there isn't a good one-size-fits-all transaction schema.

PR #65, opened 2022-05-09
WIP: refactor(logging): Replace zerolog with zap

DavidNix:


chain/cosmos/chain_node.go
```diff
@@ -211,7 +211,7 @@ func (tn *ChainNode) WaitForBlocks(blocks int64) (int64, error) {
 	startingBlock := stat.SyncInfo.LatestBlockHeight
 	mostRecentBlock := startingBlock
 	tn.logger().
-		WithField("initialHeight", startingBlock).
+		With("initialHeight", startingBlock).
```
DavidNix, line 214: Taking a leaf out of zap's book: `With` is shorter. 😄

chain/cosmos/chain_node.go
```diff
@@ -703,7 +702,7 @@ func (tn *ChainNode) StartContainer(ctx context.Context) error {
 	tn.Container = c
 
 	port := dockerutil.GetHostPort(c, rpcPort)
-	tn.logger().WithField("container", tn.Name()).Infof("RPC => %s", port)
+	tn.logger().With("container", tn.Name()).Infof("RPC => %s", port)
```
mark-rushakoff, line 705: Using formatted messages misses part of the point of structured logging. This ought to provide a structured `port` field and log a fixed message like "RPC listening".

chain/cosmos/chain_node.go
```diff
@@ -234,8 +234,8 @@ func (tn *ChainNode) WaitForBlocks(blocks int64) (int64, error) {
 
 		if deltaBlocks >= blocks {
 			tn.logger().
-				WithField("initialHeight", startingBlock).
-				Debugf("Time (sec) waiting for %d blocks: %d", blocks, i+1)
+				With("initialHeight", startingBlock).
+				Debugf("time waiting for %d blocks: %ds", blocks, i+1)
```
mark-rushakoff, line 238: Duration and blocks should be fields, not encoded in the message string.

chain/cosmos/chain_node.go
```diff
@@ -653,9 +653,8 @@ func (tn *ChainNode) CreateNodeContainer() error {
 	chainCfg := tn.Chain.Config()
 	cmd := []string{chainCfg.Bin, "start", "--home", tn.NodeHome(), "--x-crisis-skip-assert-invariants"}
 	tn.logger().
-		WithField("container", tn.Name()).
-		WithField("command", strings.Join(cmd, " ")).
-		Info()
+		With("container", tn.Name()).
+		Info(strings.Join(cmd, " "))
```
mark-rushakoff, line 657: The previous version was more appropriate, maybe with a fixed message of "Running command".

chain/cosmos/chain_node.go
```diff
@@ -785,7 +784,7 @@ func (nodes ChainNodes) PeerString() string {
 		}
 		hostName := n.HostName()
 		ps := fmt.Sprintf("%s@%s:26656", id, hostName)
-		nodes.logger().WithField("container", n.Name()).Infof("(%s) peering (%s)", hostName, ps)
+		nodes.logger().With("container", n.Name()).Infof("(%s) peering (%s)", hostName, ps)
```
mark-rushakoff, line 787: Move hostname and peer string(?) to fields.

mark-rushakoff (changes requested):
The main reason I am looking for fully structured logging here (as opposed to partially structured with Debugf/Infof) is that I frequently end up collecting logs across many runs and postprocessing them. Having fixed messages with all detail in structure makes that workflow simple and predictable, because I don't have to do any processing on the message portion.

Because we fully control the logging in this case -- that is, we are not producing a logger that a third party library needs to consume -- I would rather drop the \*f methods to more strongly encourage use of non-formatted messages.

Having agreed to stick with Zap for consistency on the other Strangelove projects, I still don't see why there is an abstraction interface wrapping it.

chain/cosmos/chain_node.go
DavidNix: I agree. I kept the existing pattern in this case.

chain/cosmos/chain_node.go
DavidNix: Good point. This was the previous pattern.

chain/cosmos/chain_node.go
DavidNix: I agree. Like other parts of the codebase, this was the previous pattern.

chain/cosmos/chain_node.go
DavidNix: The way zap prints console output, I felt it was a better user experience seeing the command in the mssage vs. trying to find it in a field. But that is just my preference. For filtering/searching, it's probably better in a field.

DavidNix:
@mark-rushakoff Given the silence on log abstraction vs. not, I went ahead and refactored to use zap's concrete logger. It's definitely not a hill worth dying on. (If we had a monorepo or different code organization architecture, I'd probably die on it. 😄). So this is ready for your 👀 again.

DavidNix:
How logs look after the refactor:
Screen Shot 2022-05-10 at 2 30 10 PM

chain/cosmos/chain_node.go
```diff
@@ -234,22 +233,25 @@ func (tn *ChainNode) WaitForBlocks(blocks int64) (int64, error) {
...
+				)
 			return mostRecentBlock, nil // done waiting for consecutive signed blocks
 		}
 	}
 	return mostRecentBlock, errors.New("timed out waiting for blocks")
 }
 
 func (tn *ChainNode) maybeLogBlock(height int64) {
-	if tn.logger().Level() != "debug" {
+	entry := tn.logger().Check(zap.DebugLevel, "current block")
```
mark-rushakoff, line 247: This is a bit of an unusual pattern -- I see that we are trying to avoid the extra transactions involved if the logger is not enabled at debug level, and that makes sense. It throws me off to see the message `current block` declared once and then different fields depending on what exit path this function takes.

I think the more common pattern here would be:

```go
if !tn.logger().Core().Enabled(zap.DebugLevel) {
  return
}
// ...
if err != nil {
  tn.logger().Debug("Failed to get block", zap.Int64("height", height), zap.Error(err))
}
// ...
tn.logger().Debug("Current block", zap.Int64("height", height), zap.String("block", pp))
```

chain/cosmos/chain_node.go
```diff
@@ -213,8 +213,7 @@ func (tn *ChainNode) WaitForBlocks(blocks int64) (int64, error) {
 	startingBlock := stat.SyncInfo.LatestBlockHeight
 	mostRecentBlock := startingBlock
 	tn.logger().
-		WithField("initialHeight", startingBlock).
-		Debug("wait for blocks")
+		Info("waiting for blocks", zap.Int64("initialHeight", startingBlock))
```
mark-rushakoff, line 216: We have logging standards docs at https://github.com/cosmos/relayer/blob/main/docs/logging.md#usage (and identical content in lens) that specifies:

> Log messages should begin with an uppercase letter ...
> Names of log fields should be all lowercase `snake_case` format.

test/relayer_test.go
```diff
@@ -30,6 +30,9 @@ func TestRelayer(t *testing.T) {
 		t.Skip()
 	}
 
-	logger := log.New(os.Stderr, "console", "info")
+	logger, err := zap.NewProduction()
```
mark-rushakoff, line 33: I would have expected `zaptest.NewLogger(t)` here, or otherwise perhaps a logger pointed at a file.

mark-rushakoff (commented):
Would you mind updating the messages to start with an uppercase letter and using `snake_case` on the fields? Then this should be good to go.

chain/cosmos/chain_node.go
DavidNix: Ah, thanks for that. I was looking for something similar, but the zap API did not make it obvious.

chain/cosmos/chain_node.go
DavidNix: TIL. I'll add a section in the README pointing to this style guide or perhaps copy that section into ibctest's README.

DavidNix: Fyi, the logging style guide will be another PR.

DavidNix:
Updates have been made to mostly follow https://github.com/cosmos/relayer/blob/main/docs/logging.md#usage. I ask that any nitpicks be addressed on future PRs.

mark-rushakoff (approved):
Thank you! :shipit:

PR #66, opened 2022-05-10
WIP: fix: Remove duplicate tests

DavidNix:


cmd/ibctest/ibctest_test.go
```diff
@@ -55,8 +55,8 @@ func setUpTestMatrix() error {
 		testMatrix.Relayers = []string{"rly"}
 		testMatrix.ChainSets = [][]ibctest.BuiltinChainFactoryEntry{
 			{
-				{Name: "gaia", Version: "latest", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
-				{Name: "osmosis", Version: "latest", ChainID: "osmosis-1001", NumValidators: 1, NumFullNodes: 1},
+				{Name: "gaia", Version: "v7.0.1", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
+				{Name: "osmosis", Version: "v7.2.0", ChainID: "osmosis-1001", NumValidators: 1, NumFullNodes: 1},
```
DavidNix, line 59: This was the quickest path to ensuring stability for CI. 

I also feel if a new user runs `ibctest` binary, they'd probably also want a stable experience. 

The tradeoff is remembering to update these versions...

An alternative (which requires more work) is to pass in a matrix file for CI.

cmd/ibctest/ibctest_test.go
```diff
@@ -55,8 +55,8 @@ func setUpTestMatrix() error {
 		testMatrix.Relayers = []string{"rly"}
 		testMatrix.ChainSets = [][]ibctest.BuiltinChainFactoryEntry{
 			{
-				{Name: "gaia", Version: "latest", ChainID: "cosmoshub-1004", NumValidators: 1, NumFullNodes: 1},
-				{Name: "osmosis", Version: "latest", ChainID: "osmosis-1001", NumValidators: 1, NumFullNodes: 1},
+				{Name: "gaia", Version: "v7.0.1", ChainID: "cosmoshub-1004", NumValidators: 2, NumFullNodes: 1},
+				{Name: "osmosis", Version: "v7.2.0", ChainID: "osmosis-1001", NumValidators: 2, NumFullNodes: 1},
```
DavidNix, line 59: This was the quickest path to ensure CI stability.

I also feel if a new user runs ibctest binary, they'd probably also want a stable experience.

The tradeoff is remembering to update these versions... Could be ldflags or similar. 

An alternative (which requires more work) is to pass in a matrix file for CI.

.github/workflows/tests.yml
```diff
@@ -33,5 +33,5 @@ jobs:
       #     docker network prune -f || true
 
       # run tests
-      - name: run tests
-        run: go test -timeout 30m -v -p 2 ./...
+      - name: run all tests
+        run: (go test -timeout 30m -v -p 2 ./...) || (echo "\n\n*****CHAIN and RELAYER LOGS*****" && tail -n 20 "$HOME/.ibctest/logs/ibctest.log" && exit 1)
```
agouin, line 37: Looks like you are purposefully throwing an error in the log tail so that it won't consider that for pass/fail, only the go test exit code will be considered? Fine by me, just took a minute to understand the intention.

cmd/ibctest/ibctest_test.go
agouin: Fine by me for CI/new user stability. I'm sure we will have our full blown matrix file(s) for the full suite output soon enough.

agouin (approved):

.github/workflows/tests.yml
DavidNix: I'll add a comment. Basically, show the chain and relayer logs on a test failure.

PR #68, opened 2022-05-11
Add Name() method to relayer and chain factories

mark-rushakoff:
This moves the name generation logic into the factory, where it belongs,
instead of in the test generation logic.

This will also assist in producing additional reports about test
results.


DavidNix (approved):
🚢

PR #69, opened 2022-05-11
docs: Add logging style guide

DavidNix:

Vendor the style guide present in other Strangelove (SL) repos such as relayer and lens. The style guide has good guidelines so that we have consistent logging across different SL services.

I propose adding this style guide here unaltered.

*The following only presents alternatives. It is not a proposal (yet).*

I think style should be enforced by static analysis tools like linters, auto-formatters, or by other automated means. And those tools run on CI.

Otherwise, you run the risk of bikeshedding as developers critique code style back-and-forth in PRs. In my experience, such behavior has low business value.

Alternatively to linters, an interesting thought experiment would be creating a `strangelove/log` repo which wraps `zap`. All other Strangelove services use this internally maintained package.

You can then guarantee style such as.

```go
// pseudocode 
func (log logger) Info(msg string, fields ...Field) {
  for _, field := range fields {
     field.Key = snakeCase(field.Key)
  }
  log.zapLogger.Info(capitalizeFirstWord(msg), fields...)
}
```

As always, there are tradeoffs such as yet-another-repo and dependency to maintain. I am not suggesting we go down this path. I only mean to show alternative approaches.

- NA

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,73 @@
+# Logging
+
+This code uses the [zap](https://pkg.go.dev/go.uber.org/zap) logging framework.
+
+Zap is a production quality logging library that is performant and highly configurable.
+Zap produces structured logging that can be easily machine-parsed,
+so that people can easily use tooling to filter through logs,
+or logs can easily be sent to log aggregation tooling.
+
+## Conventions
+
+### Instantiation
+
+Structs should hold a reference to their own `*zap.Logger` instance, as an unexported field named `log`.
+Callers should provide the logger instance as the first argument to a `NewFoo` method, setting any extra fields with the `With` method externally.
+For example:
+
+```go
+path := "/tmp/foo"
+foo := NewFoo(log.With(zap.String("path", path)), path)
+```
+
+### Usage
+
+Log messages should begin with an uppercase letter and should not end with any punctuation.
+Log messages should be constant string literals; dynamic content is to be set as log fields.
+Names of log fields should be all lowercase `snake_case` format.
+
+```go
+// Do this:
+x.log.Debug("Updated height", zap.Int64("height", height))
+
+// Don't do this:
+x.log.Debug(fmt.Sprintf("Updated height to %d.", height))
+```
+
+Errors are intended to be logged with `zap.Error(err)`.
+Error values should only be logged when they are not returned.
+
+```go
+// Do this:
+for _, x := range xs {
+  if err := x.Validate(); err != nil {
+    y.log.Info("Skipping x that failed validation", zap.String("id", x.ID), zap.Error(err))
+    continue
+  }
+}
+
+// Don't do this:
+if err := x.Validate(); err != nil {
+  y.log.Info("X failed validation", zap.String("id", x.ID), zap.Error(err))
+  return err // Bad!
+  // Returning an error that has already been logged may cause the same error
+  // to be logged many times.
+}
+```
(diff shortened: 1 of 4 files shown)


docs/logging.md
```diff
@@ -0,0 +1,73 @@
...
+### Instantiation
+
+Structs should hold a reference to their own `*zap.Logger` instance, as an unexported field named `log`.
+Callers should provide the logger instance as the first argument to a `NewFoo` method, setting any extra fields with the `With` method externally.
+For example:
+
+```go
+path := "/tmp/foo"
+foo := NewFoo(log.With(zap.String("path", path)), path)
+```
```
DavidNix, line 21: I only present this as an alternative perspective. I am not suggesting we make any changes.

This is a polarizing opinion. I like using a global (singelton) logger.

Typically the default singleton is a no-op. 

In `main()` you initialize the singleton once. 

E.g. 

```go
func main() {
    log.Init(out, "json", "debug")
}
```

A factory method lets you grab a logger at any time.

```go
log.Global().Info("Hi!")

aService := SomeService{
    log: log.Global().WithField("service", "SomeService")
}
```

The benefit is simplifying constructors and other callers. You no longer have to toil with passing a logger down a call stack. 

It has the tradeoffs of any singleton and global state, of course. Additionally, you could iadvertently hide logging in a performance critical code path. (But profiling will show it immediately.)

But, I've successfully used this singleton pattern in large-scale, microservice  systems with low toil. The diagnositc quality of the logs was very high.

docs/logging.md
```diff
@@ -0,0 +1,73 @@
...
+Info level messages and higher are on by default.
+They may contain any general information useful to an operator.
+It should be safe to discard all info level messages if an operator so desires.
+
+Warn level messages should be used to indicate a problem that may worsen without operator intervention.
+
+Error level messages should only be used to indicate a serious problem that cannot automatically recover.
+Error level messages should be reserved for events that are worthy of paging an engineer.
+
+Do not use Fatal or Panic level messages.
```
DavidNix, line 73: A wrapped logger could eliminate these rules by simply not exporting `Fatal, Panic` as public methods.

mark-rushakoff (approved):
I do like the idea of adding a linter, and my old coworkers, where these conventions were initially developed, would probably love to use that if we were to publish one. There are plenty of guides out on the internet on how to write Go linters, but I can't make time to spend on that for the foreseeable future.

This guide suited us well on a large monorepo and some separate repositories as we scaled contributors from 15-20 up to around a hundred.

There were certainly style violations that got merged on occasion, but they were easily addressed later upon noticing. Nonetheless, a linter would be better.

With structured logging, my primary objection about a singleton logger is that you lose detail about log hierarchy. In the case of Relayer, we might have a `*zap.Logger` instance associated with a chain, with a string field `chain_name="cosmos"`. That chain object may have child light client object, and we would derive a new logger and add the client name, so the client's logger would have fields `chain_name="cosmos", client_name="client-01"`. Then the client object may have a connection object with another derived logger, and once again with a channel object, and so on.

DavidNix:
Thanks for the thoughtful response. 

> With structured logging, my primary objection about a singleton logger is that you lose detail about log hierarchy. 

You would think so, but using the singleton pattern in previous production systems, it was never a problem. To clarify the singleton is never modified. It's always a base logger with no fields. Most of the time, you build a derived logger using the singleton as the base.

PR #72, opened 2022-05-12
Add initial test reporter

mark-rushakoff:
This PR is the first step towards producing machine-readable test reports.

The `ibctest` executable now emits a report, defaulting to `~/.ibctest/reports/$UNIX_TIMESTAMP_NOW.json`, that is a file containing a stream of JSON objects representing test events that would be relevant to produce a human-readable report.

Internally, this introduces a `testreporter` package with a `*Reporter` type, which is used additively during plain Go tests to collect details at an individual test level in a machine-readable JSON format. See `testreporter/doc.go` for a thorough explanation of how to use the reporter.

As of this PR, the reporter tracks suite start and end times, and individual start/stop times along with skip and fail status -- which is roughly parity with the plain output from Go test. 

However, there is one additional relevant feature of the reporter, which is its integration with testify `require` and `assert`. An instance of `require` can be created in a test with `req := require.New(reporter.TestifyT(t))`, and then any failed calls such as `req.NoError(err, "failure to foo")` will be logged to the plain `go test` output as usual, but it will also record an entry in the report log. This way, we can not only present the fact that a test failed, but we can also note the exact error message that was logged as the failure.

The two significant remaining pieces of work for the test reporter are:
1. Include output from relayer executions, to allow drilling in to failures or investigating timing outliers. Not included in this PR because it will be a decent size change.
2. Integrate skip messages with the reporter. While the reporter can see that a test _was_ skipped, it does not have insight into the specific message passed to `t.Skip`. While I could have just added a wrapper for that, we have multiple reasons that a test may be skipped, including a missing relayer capability or filtering by including or excluding labels for tests.

I expect that the information we include in the report will expand over time. We may eventually want to include the chain logs, for instance. We may have more application-specific details to log, such as block-level information.

### DIFF
--- a/relayertest/test.go
+++ b/relayertest/test.go
@@ -177,12 +177,13 @@ func sendIBCTransfersFromBothChainsWithTimeout(
 // This is intended to be used by Go unit tests.
 func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactory, rep *testreporter.Reporter) {
 	rep.TrackTest(t)
+	req := require.New(rep.TestifyT(t))
 
 	ctx, home, pool, network, err := ibctest.SetupTestRun(t)
-	require.NoError(t, err, "failed to set up test run")
+	req.NoError(err, "failed to set up test run")
 
 	srcChain, dstChain, err := cf.Pair(t.Name())
-	require.NoError(t, err, "failed to get chain pair")
+	req.NoError(err, "failed to get chain pair")
 
 	var (
 		preRelayerStartFuncs []func([]ibc.ChannelOutput)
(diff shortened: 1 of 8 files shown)


relayertest/test.go
```diff
@@ -174,12 +175,15 @@ func sendIBCTransfersFromBothChainsWithTimeout(
 
 // TestRelayer is the stable API exposed by the relayertest package.
 // This is intended to be used by Go unit tests.
-func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactory) {
+func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactory, rep *testreporter.Reporter) {
+	rep.TrackTest(t)
+	req := require.New(rep.TestifyT(t))
```
DavidNix, line 180: I like this pattern!

DavidNix (approved):
No changes from me. I like where this is headed!

agouin (approved):
Tried it out, output looks great as a start! Parsing will be a bit interesting since it is json per line instead of a json parseable file, but I do like that it streams the events to the file as they occur so that you can still see the report progress if you kill a running test.

PR #74, opened 2022-05-12
fix: Ensure skipped tests are not omitted

DavidNix:
In attempting to test acknowledgements for `height timeout` and `timestamp timeout` scenarios, I discovered `timestamp` always passed and `height` always failed.

But what was really happening was `timestamp timeout` is not a capability of the go relayer `rly`. 

We were inadvertently omitting the test entirely. Therefore, unless you looked closely, it appears as if the test passes, when in fact, it wasn't run at all. 

This PR fixes the above issue, so you will explicity see `SKIP` in the test output like so:

Screen Shot 2022-05-12 at 4 32 38 PM

### DIFF
--- a/relayertest/test.go
+++ b/relayertest/test.go
@@ -193,6 +193,7 @@ func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactor
 		testCases = append(testCases, &testCase)
 
 		if len(missingCapabilities(rf, testCaseConfig.RequiredRelayerCapabilities...)) > 0 {
+			// do not execute relayer before hook if capability missing
 			continue
 		}
 		preRelayerStartFunc := func(channels []ibc.ChannelOutput) {
--- a/relayertest/test.go
+++ b/relayertest/test.go
@@ -104,7 +104,7 @@ func requireCapabilities(t *testing.T, rf ibctest.RelayerFactory, reqCaps ...rel
 
 func missingCapabilities(rf ibctest.RelayerFactory, reqCaps ...relayer.Capability) []relayer.Capability {
 	caps := rf.Capabilities()
-	missing := []relayer.Capability{}
+	var missing []relayer.Capability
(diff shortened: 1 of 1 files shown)


relayertest/test.go
```diff
@@ -187,15 +187,16 @@ func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactor
 	)
 
 	for _, testCaseConfig := range relayerTestCaseConfigs {
-		if len(missingCapabilities(rf, testCaseConfig.RequiredRelayerCapabilities...)) > 0 {
-			continue
-		}
```
DavidNix, line 192: This was the culprit. By bailing out here, we do not add the test caste to the testCases slice on line 196. Thus the test is omitted entirely. 

We have code later that `t.Skip`s tests.

relayertest/test.go
```diff
@@ -187,15 +187,16 @@ func TestRelayer(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFactor
...
 		testCase := RelayerTestCase{
 			Config: testCaseConfig,
 		}
 		testCases = append(testCases, &testCase)
-		preRelayerStartFunc := func(channels []ibc.ChannelOutput) {
 
+		if len(missingCapabilities(rf, testCaseConfig.RequiredRelayerCapabilities...)) > 0 {
+			// do not execute relayer before hook if capability missing
+			continue
+		}
```
DavidNix, line 198: By letting all preRelayerStartFunc's go through, it seemed to cause test pollution. So it was necessary to prevent these "before hooks" from executing. 

FWIW, it's on my radar to think more about this section of code in the near future. 

I understand it's trying to do setup once. However, it has more indirection than I'd like which makes it more difficult to determine the order of operations.

It's also setting a struct field in a closure as well as calling a nested func.

mark-rushakoff (approved):
Looks better for now. I agree this section could be cleaned up more.

PR #75, opened 2022-05-12
feat(logging): Improved logging of cosmos sdk blocks and transactions

DavidNix:

If debug logging, pretty print Cosmos SDK transactions to trace the lifecycle of the chains. 

Previously, we were printing the Tendermint transactions which do not contain as much context or info.

## Example

Screen Shot 2022-05-12 at 5 02 56 PM

* It's quite a bit of text to parse through but it's easier to follow than the previous implementation which would print raw bytes. 
* It's a lot of information so not practical to add as logging fields, imo. Events alone can be lots of data.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -21,7 +22,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/cosmos/cosmos-sdk/simapp"
 	"github.com/cosmos/cosmos-sdk/types"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	authTx "github.com/cosmos/cosmos-sdk/x/auth/tx"
 	bankTypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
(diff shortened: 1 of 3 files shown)


chain/cosmos/chain_node.go
```diff
@@ -257,12 +262,29 @@ func (tn *ChainNode) maybeLogBlock(height int64) {
...
 		return
 	}
-	pp, err := tendermint.PrettyPrintTxs(ctx, txs, tn.Client.Tx)
-	if err != nil {
-		tn.logger().Info("Failed to pretty print block", zap.Error(err))
-		return
+	sb := new(strings.Builder)
+	sb.WriteString("BLOCK INFO\n")
+	sb.WriteString(fmt.Sprintf("BLOCK HEIGHT: %d\n", height))
+	sb.WriteString(fmt.Sprintf("TOTAL TXs: %d\n", len(blockRes.Block.Txs)))
```
mark-rushakoff, line 268: Just as a totally minor nit, this would likely be more appropriate as a `bytes.Buffer` -- we are dynamically building these strings, so they may as well be `fmt.Fprintf(b, "...")` to keep everything as byte slices internally.

chain/cosmos/chain_node.go
```diff
@@ -257,12 +262,29 @@ func (tn *ChainNode) maybeLogBlock(height int64) {
...
+			continue
+		}
+		sb.WriteString(fmt.Sprintf("TX TYPE: %s\n", txResp.Tx.TypeUrl))
+
+		// Purposefully zero out fields to make spew's output less verbose
+		txResp.Data = "[redacted]"
+		txResp.RawLog = "[redacted]"
+		txResp.Events = nil // already present in TxResponse.Logs
+
+		sb.WriteString(spew.Sprint(txResp))
```
mark-rushakoff, line 284: This could be `spew.Fprint(b, txResp)` too.

mark-rushakoff (approved):
The string builder could be a bytes.Buffer, but it's not going to be a material difference one way or the other.

chain/cosmos/chain_node.go
DavidNix: Ha! I had it as a `bytes.Buffer` at first. I'll make the change.

PR #76, opened 2022-05-13
Add relayer executions to test report

mark-rushakoff:
This was uniform to apply to all the one-off commands, but StartRelayer
and StopRelayer required a slightly different approach due to the
asynchronous style of starting the relayer in the background.

Now, container information, start and stop time, standard output and
standard error, for each relayer execution are included in the report,
so long as the ibc.Relayer implementation has implemented it.

### DIFF
--- a/relayer/rly/cosmos_relayer.go
+++ b/relayer/rly/cosmos_relayer.go
@@ -104,18 +104,18 @@ func (relayer *CosmosRelayer) HostName(pathName string) string {
 	return dockerutil.CondenseHostName(fmt.Sprintf("%s-%s", relayer.Name(), pathName))
 }
 
-func (relayer *CosmosRelayer) LinkPath(ctx context.Context, pathName string) error {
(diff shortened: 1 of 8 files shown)


relayer/rly/cosmos_relayer.go
```diff
@@ -138,21 +138,46 @@ func (relayer *CosmosRelayer) GetChannels(ctx context.Context, chainID string) (
...
+	})
+
+	stdout := stdoutBuf.String()
+	stderr := stderrBuf.String()
+
+	rep.TrackRelayerExec(
+		relayer.container.Name,
+		relayer.container.Args, // Is this correct for the command?
+		stdout, stderr,
+		relayer.container.State.ExitCode, // Is this accurate?
```
DavidNix, line 173: Re your comments: Good questions. 

Tough to know for sure without running the suite and observing. I personally don't know this docker pkg well enough to say one way or the other.

DavidNix (approved):
LGTM! Nice work

PR #78, opened 2022-05-13
Penumbra genesis cleanup

agouin:
Removes manually assembled genesis file. Uses `pd generate-testnet` command instead


chain/penumbra/penumbra_app_node.go
```diff
@@ -105,6 +106,61 @@ func (p *PenumbraAppNode) WalletPathContainer() string {
...
+
+func (p *PenumbraAppNode) GenerateGenesisFile(
+	ctx context.Context,
+	chainID string,
+	validators []PenumbraValidatorDefinition,
+	allocations []PenumbraGenesisAppStateAllocation,
+) error {
+	validatorsJson, err := json.Marshal(validators)
+	if err != nil {
+		return fmt.Errorf("error marshalling validators to json: %v", err)
```
mark-rushakoff, line 141: `%w` is preferred and subtly different from `%v` for errors.

`%w` allows callers to use [`errors.Is`](https://pkg.go.dev/errors#Is) to introspect the wrapped error, whereas `%v` includes the string-formatted error in a new error, and the caller has no way to get at the wrapped error any more.

chain/penumbra/penumbra_app_node.go
```diff
@@ -105,6 +106,61 @@ func (p *PenumbraAppNode) WalletPathContainer() string {
...
+	}
+	if err := os.WriteFile(p.ValidatorsInputFile(), validatorsJson, 0644); err != nil {
+		return fmt.Errorf("error writing validators to file: %v", err)
+	}
+	allocationsCsv := []byte("\"amount\",\"denom\",\"address\"\n")
+	for _, allocation := range allocations {
+		allocationsCsv = append(allocationsCsv, []byte(fmt.Sprintf("\"%d\",\"%s\",\"%s\"\n", allocation.Amount, allocation.Denom, allocation.Address))...)
```
mark-rushakoff, line 148: These might be more readable using raw strings, like `` `"%d","%s","%s"` + "\n"``. After more than two or so backslashes, I find it becomes quickly difficult to read.

mark-rushakoff (approved):
I am not very familiar with this area of the code, but the changes look good as far as I can tell.

chain/penumbra/penumbra_app_node.go
agouin: Done

chain/penumbra/penumbra_app_node.go
agouin: Done

PR #85, opened 2022-05-17
Add test labels

mark-rushakoff:
As of this change, labels are associated with individual tests and are
only used for reporting purposes. Filtering tests to run from the
command line is more complex than it might first appear, so that will
happen in a separate change. Label presence in test reports today could
be used for viewing a filtered subset of the report, however.

There are Relayer and Chain labels that are included with top-level
tests, and then individual subtests may have an optional set of labels.

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -20,6 +20,11 @@ var knownTestLabels = map[Test]struct{}{
 	TimestampTimeout: {},
 }
 
+func (l Test) IsKnown() bool {
+	_, exists := knownTestLabels[l]
+	return exists
+}
+
 // Relayer is a label associated with a relayer during tests.
 // Relayer values must be registered through RegisterRelayerLabel, typically inside init functions.
 type Relayer string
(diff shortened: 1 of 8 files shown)


label/label.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Test is a label associated with an individual test.
+// All test labels are known at compile time.
+type Test string
+
+const (
+	Timeout          Test = "timeout"
+	HeightTimeout    Test = "height_timeout"
+	TimestampTimeout Test = "timestamp_timeout"
+)
```
DavidNix, line 15: Do we need a label for `relay packet`? (A better name may be `fungible transfer` or similar.)

Also, I now see there's a small hierarchy or set membership going on. I was at first confused by `Timeout` but now I see it encompasses all timeout tests. The `HeightTimeout` and other variants let you focus more.

DavidNix (approved):
LGTM!

Just a small general question. Nothing blocking.

label/label.go
mark-rushakoff: It crossed my mind to add an explicit description associated with each test label. If it wasn't immediately obvious now, it will be even less obvious as the label count grows. 

 I'll add that in the next iteration on this part of the code. Probably only for test labels because those are all determined at compile time.

PR #86, opened 2022-05-17
refactor: Generalize WaitForBlocks; reduce ibc.Chain interface

DavidNix:

It started as an exploration for polling for state. My original hypothesis is we could remove `WaitForBlocks` and poll for some state. However, `WaitForBlocks` used extensively and we can't remove it entirely. Additionally, the state we poll can vary wildely.

Therefore, this PR is an effort to reduce the large `ibc.Chain` interface. 

This removes the `WaitForBlocks` from the `ibc.Chain` interface in favor of using the `Height()` method with a new, tested helper function. 

Chains implementing `Height()` should be much easier than `WaitForBlocks`.

* Removes "wait for blocks" log statement. I can add it back. However, I personally don't miss it.
* For cosmos chain, on debug log level, it may log more blocks. 
* Removes the `sleep(1 second)` between waiting for blocks. The tests seem to be passing without further flakiness. I don't know the context of why that sleep existed, but can add it back if necessary.

### DIFF
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -172,10 +172,6 @@ func (c *CosmosChain) WaitForBlocks(number int64) (int64, error) {
 	return c.getRelayerNode().WaitForBlocks(number)
 }
 
-func (c *CosmosChain) Height() (int64, error) {
-	return c.getRelayerNode().Height()
-}
-
 // Implements Chain interface
 func (c *CosmosChain) GetBalance(ctx context.Context, address string, denom string) (int64, error) {
 	params := &bankTypes.QueryBalanceRequest{Address: address, Denom: denom}
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -22,6 +22,7 @@ import (
 	"github.com/strangelove-ventures/ibctest/chain/tendermint"
 	"github.com/strangelove-ventures/ibctest/dockerutil"
 	"github.com/strangelove-ventures/ibctest/ibc"
+	"github.com/strangelove-ventures/ibctest/test"
 	"go.uber.org/zap"
 
 	"github.com/ory/dockertest/v3"
@@ -86,7 +87,7 @@ func (c *CosmosChain) Initialize(testName string, homeDirectory string, dockerPo
 	return nil
 }
 
-func (c *CosmosChain) getRelayerNode() *ChainNode {
+func (c *CosmosChain) getFullNode() *ChainNode {
 	if len(c.ChainNodes) > c.numValidators {
 		// use first full node
 		return c.ChainNodes[c.numValidators]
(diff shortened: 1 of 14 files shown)


chain/cosmos/cosmos_chain.go
```diff
@@ -134,52 +135,43 @@ func (c *CosmosChain) GetAddress(ctx context.Context, keyName string) ([]byte, e
...
-func (c *CosmosChain) Height() (int64, error) {
-	return c.getRelayerNode().Height()
+	return c.getFullNode().CreatePool(ctx, keyName, contractAddress, swapFee, exitFee, assets)
 }
 
 // Implements Chain interface
 func (c *CosmosChain) GetBalance(ctx context.Context, address string, denom string) (int64, error) {
 	params := &bankTypes.QueryBalanceRequest{Address: address, Denom: denom}
-	grpcAddress := dockerutil.GetHostPort(c.getRelayerNode().Container, grpcPort)
+	grpcAddress := dockerutil.GetHostPort(c.getFullNode().Container, grpcPort)
```
DavidNix, line 174: The name "relayer node" didn't really make sense to me. It's not a term you find in cosmos or tendermint. So I changed it. 

FWIW, I don't like `full node` either because I think "full" means it does everything, including validate. But "full" is used a lot in this space.

test/wait_for_blocks.go
```diff
@@ -0,0 +1,74 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+// ChainHeighter fetches the current chain block height
+type ChainHeighter interface {
+	Height(ctx context.Context) (uint64, error)
+}
+
+// WaitForBlocks blocks until all chains reach a block height delta equal to or greater than the delta argument
+func WaitForBlocks(ctx context.Context, delta int, chains ...ChainHeighter) error {
```
DavidNix, line 15: I had to use a different package `test` because of import cycles. May be worth changing in the future. 

This demonstrates how to use a tiny interface `ChainHeighter` in this case. 

For now, I had to keep `Height()` in the `ibc.Chain` interface because the factory returns the iface, not a concrete type.

mark-rushakoff: I would have expected this to fit nicely in the root `ibctest` package, but it is used in the chain implementations, and `ibctest` imports those chains. 

Package `test` is fine for now, but I would like to see it renamed once we figure out what else fits in here.

test/wait_for_blocks.go
```diff
@@ -0,0 +1,74 @@
...
+	var (
+		done  = make(chan struct{})
+		errCh = make(chan error)
+	)
+
+	var eg errgroup.Group
```
mark-rushakoff, line 21: I think this one would a proper use of https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext -- if we are waiting for multiple chains, as soon as the first one fails, the shared context should be canceled so that we don't wait any longer for the other chains.

chain/tendermint/tendermint_node.go
```diff
@@ -52,7 +50,9 @@ type ContainerPort struct {
 type Hosts []ContainerPort
 
 const (
-	blockTime   = 2 // seconds
+	// BlockTime is approx time to create a block
```
mark-rushakoff, line 53: Probably keep mentioning `in seconds` in this comment.

chain/cosmos/cosmos_chain.go
mark-rushakoff: I think the relayer node here really meant, the full node that the relayer is configured to connect to. `getFullNode` doesn't really capture that it will always return the same full node, but `getRelayerNode` wasn't intuitive either. I don't have a better name suggestion offhand yet.

mark-rushakoff (commented):
I left a handful of style comments and suggestions. I think this is ready to ship, but leaving in comment state until the next round.

test/wait_for_blocks.go
```diff
@@ -0,0 +1,74 @@
...
+		})
+	}
+
+	go func() {
+		if err := eg.Wait(); err != nil {
+			errCh <- err
+			return
+		}
+		close(done)
+	}()
```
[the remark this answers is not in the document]
DavidNix: Good point. The `ctx` being part of the `Height(...)` interface method was a late addition.

test/wait_for_blocks.go
```diff
@@ -0,0 +1,74 @@
...
+	for i := range chains {
+		chain := chains[i]
+		eg.Go(func() error {
+			h := &height{}
+			for h.Delta() < delta {
+				cur, err := chain.Height(ctx)
+				if err != nil {
+					return err
+				}
+				h.Update(cur)
```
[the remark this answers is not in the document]
DavidNix: It's a good question and hard to be sure. We used to sleep 1 second. Thus far, no rate limiting seems to be fine for the current tests.  I think for now leave it as-is and modify later if it becomes a problem.

DavidNix:
I am not sure why GH is not letting me reply to some comments but:

> I would have expected this to fit nicely in the root ibctest package

I agree. But import cycles... 😢 I'm sure a new pattern will emerge that makes sense. I didn't want to spend time fighting the import cycle for this PR.

chain/cosmos/cosmos_chain.go
DavidNix: Tough one to name here for sure. For me `relayerNode` on the surface meant the process in which the relayer itself was running. In this case, it means the chain process from which the relayer can query.

test/wait_for_blocks.go
```diff
@@ -0,0 +1,60 @@
...
+type ChainHeighter interface {
+	Height(ctx context.Context) (uint64, error)
+}
+
+// WaitForBlocks blocks until all chains reach a block height delta equal to or greater than the delta argument.
+func WaitForBlocks(parent context.Context, delta int, chains ...ChainHeighter) error {
+	if len(chains) == 0 {
+		panic("missing chains")
+	}
+	eg, ctx := errgroup.WithContext(parent)
```
mark-rushakoff, line 19: The naming we are using elsewhere is keeping `ctx` as the surrounding method's parameter, so this line looks like `eg, egCtx := errgroup.WithContext(ctx)`.

mark-rushakoff (approved):
Thanks!

PR #87, opened 2022-05-17
Rename textual occurrences of ibc-test-framework

mark-rushakoff:
All comment and doc changes, except for the string ibc-test-framework
used in dynamic docker network names; changing that should have no
functional effect.


DavidNix (approved):
Nice catch and cleanup!

PR #91, opened 2022-05-18
Replace ibctest.SetupTestRun with ibctest.DockerSetup

mark-rushakoff:
SetupTestRun was slowing me down every time I needed to read through to
understand what it did.

It returned five values, of which three are not very special: the
context was a plain context.Background, the home directory was the
result of t.TempDir(), and a returned error was intended to always
result in a test failure.

The new ibctest.DockerSetup method returns a *dockertest.Pool and a
networkID value. Because it already accepted a *testing.T, any errors
during Docker setup will cause a call to t.Fatal to immediately fail the
test, as the test cannot continue without those Docker values.

The previously returned context and home directory are instead created
in the calling test where needed. I find this clearer.

This change also inlines the previously exported CreateNetwork function
which is only called internally from DockerSetup.

### DIFF
--- a/chain/penumbra/penumbra_chain_test.go
+++ b/chain/penumbra/penumbra_chain_test.go
@@ -18,6 +18,7 @@ func TestPenumbraChainStart(t *testing.T) {
 	t.Parallel()
 
 	pool, network := ibctest.DockerSetup(t)
+	home := t.TempDir() // Must be before chain cleanup to avoid test error during cleanup.
 
 	log := zap.NewNop()
 	chain, err := ibctest.GetChain(t.Name(), "penumbra", "015-ersa-v2,v0.35.4", "penumbra-1", 4, 1, log)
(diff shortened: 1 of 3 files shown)


agouin (approved):
Nice cleanup!

chain/penumbra/penumbra_chain_test.go
```diff
@@ -15,19 +16,21 @@ func TestPenumbraChainStart(t *testing.T) {
 	}
 
 	t.Parallel()
-	ctx, home, pool, network, err := ibctest.SetupTestRun(t)
-	require.NoErrorf(t, err, "failed to set up test run")
+
+	pool, network := ibctest.DockerSetup(t)
```
agouin, line 20: maybe we keep home as part of the `DockerSetup` just so it will always be the last cleanup item to run?

DavidNix (approved):
+1 nice cleanup!

PR #92, opened 2022-05-18
Fix data race in penumbra chain setup

mark-rushakoff:
Concurrent goroutines cannot append to the same slice, but they can
write to predetermined indices in an existing slice.


agouin (approved):
Great catch

PR #95, opened 2022-05-18
[WIP] feat: Poll for packets

DavidNix:

TODO

## Exploration
* I did not reuse the existing types (created from protos) in ibc-go. Why? 1. We'll have more control over our domain in this repo. 2. ibc-go types may be more complex than we need. 

TODO

### DIFF
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -145,7 +145,7 @@ func (c *CosmosChain) SendIBCTransfer(ctx context.Context, channelID, keyName st
 	if err != nil {
 		return tx, fmt.Errorf("send ibc transfer: %w", err)
 	}
-	txResp, err := c.GetTransaction(ctx, txHash)
+	txResp, err := c.getTransaction(ctx, txHash)
 	if err != nil {
 		return tx, fmt.Errorf("failed to get transaction %s: %w", txHash, err)
 	}
@@ -154,32 +154,30 @@ func (c *CosmosChain) SendIBCTransfer(ctx context.Context, channelID, keyName st
 	// In cosmos, user is charged for entire gas requested, not the actual gas used.
 	tx.GasSpent = txResp.GasWanted
 
-	events := txResp.Events
-
 	const evType = "send_packet"
+	events := txResp.Events
 
 	tx.Packet.SourcePort = tendermint.AttributeValue(events, evType, "packet_src_port")
 	tx.Packet.SourceChannel = tendermint.AttributeValue(events, evType, "packet_src_channel")
 	tx.Packet.DestPort = tendermint.AttributeValue(events, evType, "packet_dst_port")
 	tx.Packet.DestChannel = tendermint.AttributeValue(events, evType, "packet_dst_channel")
 	tx.Packet.TimeoutHeight = tendermint.AttributeValue(events, evType, "packet_timeout_height")
+	tx.Packet.Data = []byte(tendermint.AttributeValue(events, evType, "packet_data"))
 
 	var (
 		seq       = tendermint.AttributeValue(events, evType, "packet_sequence")
-		data      = tendermint.AttributeValue(events, evType, "packet_data")
 		timeoutTs = tendermint.AttributeValue(events, evType, "packet_timeout_timestamp")
 	)
-	tx.Packet.Data = []byte(data)
 
 	seqNum, err := strconv.Atoi(seq)
 	if err != nil {
-		return tx, fmt.Errorf("invalid sequence from events %s: %w", seq, err)
+		return tx, fmt.Errorf("invalid packet sequence from events %s: %w", seq, err)
 	}
 	tx.Packet.Sequence = uint64(seqNum)
 
 	timeoutNano, err := strconv.ParseInt(timeoutTs, 10, 64)
 	if err != nil {
-		return tx, fmt.Errorf("invalid timestamp timeout %s: %w", timeoutTs, err)
+		return tx, fmt.Errorf("invalid packet timestamp timeout %s: %w", timeoutTs, err)
 	}
 	tx.Packet.TimeoutTimestamp = time.Duration(timeoutNano)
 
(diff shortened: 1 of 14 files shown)


chain/cosmos/cosmos_chain.go
```diff
@@ -141,8 +140,48 @@ func (c *CosmosChain) SendFunds(ctx context.Context, keyName string, amount ibc.
...
+	}
+	tx.Packet.Sequence = uint64(seqNum)
+
+	timeoutNano, err := strconv.ParseInt(timeoutTs, 10, 64)
+	if err != nil {
+		return tx, fmt.Errorf("invalid packet timestamp timeout %s: %w", timeoutTs, err)
+	}
+	tx.Packet.TimeoutTimestamp = time.Duration(timeoutNano)
+
+	return tx, nil
```
DavidNix, line 184: This forces the implementation to be more specific. Is it too laborous for the implementer? 

I think it's fair to say if you send an IBC transfer, a packet must be created. 

Later in the test we call `Validate()` to ensure the data is well-formed. We can then make more granular assertions on the `ibc.Packet` for example.

ibc/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package ibc provides chain and relayer agnostic types and interfaces for the IBC spec.
+// It attempts to expose the domain model of IBC.
+//
+// The official spec documentation can be found at https://github.com/cosmos/ibc/tree/master/spec.
+//
+// Currently, the interfaces may be biased towards chains built with the cosmos sdk and the cosmos/relayer.
+package ibc
```
DavidNix, line 7: Most of my reasoning is here. I like how this package is our opportunity to exose domain entities and their relationships.

DavidNix:
@mark-rushakoff @agouin This PR is still a WIP. I'm looking for general feedback on creating richer types vs. using more primitive, scalar types. 

* Is it asking too much when implementing the chain interface?
* Will we shoot ourselves in the foot?

I do like how it's an opportunity to demonstrate the IBC domain.

ibc/doc.go
agouin: +1

chain/cosmos/cosmos_chain.go
agouin: seems reasonable to me!

ibc/chain.go
```diff
@@ -71,11 +69,8 @@ type Chain interface {
 	// get the fees in native denom for an amount of spent gas
 	GetGasFeesInNativeDenom(gasPaid int64) int64
 
-	// fetch transaction
-	GetTransaction(ctx context.Context, txHash string) (*types.TxResponse, error)
```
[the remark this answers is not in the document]
agouin: woot!

ibc/packet.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
+)
+
+// Packet is a packet sent over an IBC channel as defined in ICS-4.
+// See: https://github.com/cosmos/ibc/blob/master/spec/core/ics-004-channel-and-packet-semantics/README.md
```
mark-rushakoff, line 12: Consider using a link to a tag or a specific commit hash, in case the file is ever moved in the future.

ibc/packet.go
```diff
@@ -0,0 +1,82 @@
...
+// Proto defined at: github.com/cosmos/ibc-go/v3@v3.0.0/proto/ibc/core/channel/v1/tx.proto
+type Packet struct {
+	Sequence         uint64        // the order of sends and receives, where a packet with an earlier sequence number must be sent and received before a packet with a later sequence number
+	SourcePort       string        // the port on the sending chain
+	SourceChannel    string        // the channel end on the sending chain
+	DestPort         string        // the port on the receiving chain
+	DestChannel      string        // the channel end on the receiving chain
+	Data             []byte        // an opaque value which can be defined by the application logic of the associated modules
+	TimeoutHeight    string        // a consensus height on the destination chain after which the packet will no longer be processed, and will instead count as having timed-out
+	TimeoutTimestamp time.Duration // indicates a timestamp on the destination chain after which the packet will no longer be processed, and will instead count as having timed-out
```
mark-rushakoff, line 22: Good comments on these struct fields.

Is there a more specific definition of TimeoutTimestamp? Since it is a time.Duration, is the actual timeout calculated from when the packet is Received by the Chain?

chain/tendermint/events.go
```diff
@@ -1,24 +1,22 @@
...
 
 import (
-	"bytes"
-
 	abcitypes "github.com/tendermint/tendermint/abci/types"
 )
 
 // AttributeValue returns an event attribute value given the eventType and attribute key tuple.
 // In the event of duplicate types and keys, returns the first attribute value found.
 // If not found, ok is false.
```
mark-rushakoff, line 9: If it is correct to drop the `ok` return value, then this line can be removed.

But without specific domain knowledge about the events, I don't know whether that is accurate. If it is illegal to have an empty value, then it seems fine to drop `ok` -- maybe the comments should reflect that.

But if it is possible that an attribute value that is an empty string, then we should  keep `ok`.

mark-rushakoff (commented):
The overall approach seems good to me. It seems reasonable to have domain types that are useful to the test framework that we build.

I am not familiar enough with the broad set of chains we will eventually support, to say whether we are making anything too strict that will result in difficulty with upcoming implementations. Given what I know today, it does not look like this would cause any obvious future difficulty.

One side note, since we now have a few validation functions, it might be nice to use pkg.go.dev/go.uber.org/multierr to collect all validation errors at once. It is annoying as a developer to have to repeatedly run a test and fix one validation error at a time. I suggest multierr in particular because it integrates properly with zap and we are using it in the relayer.

And thank you for standardizing on acknowledgement.

ibc/packet.go
DavidNix: Ha, I plagurized the IBC spec documents for these comments. I have a future vision where `ibctest` can become a great learning source for IBC itself.

I used `time.Duration` because the timeout is nanoseconds which is also `time.Duration`'s unit. I'll update the comment.

In Go, I typically don't like using anything other than `time.Duration` to represent durations. All too often you see `int` or similar scalars which is error prone.

chain/tendermint/events.go
DavidNix: Went back and for on this. With strings, the zero value is safe from null dereferences, so I figured callers could check for an empty string if they so cared. I will probably add back the `ok` as you suggested. The `tx.Events` is a wildly loose data structure, so perfectly reasonable to not find something.

DavidNix:
>  it might be nice to use pkg.go.dev/go.uber.org/multierr 

I love that package and will make use of it. With these validation functions, I was copying the existing patterns in cosmos's `ValidateBasic()` implementations. But in this case, it would be super annoying to re-run the long, black-box test suite.

ibc/tx.go
```diff
@@ -0,0 +1,29 @@
+package ibc
+
+import "errors"
+
+// Tx is a generalized IBC transaction.
+type Tx struct {
```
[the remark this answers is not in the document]
DavidNix: We could rename it to `Fee` or similar. I do like that. I thought `Gas` was a common enough term in most blockchains. But that's good to know this generalization may need work in the future!

ibc/packet.go
DavidNix: Oh wait, I need more coffee apparently. This is a *timestamp* not a duration. Yes, I'll re-eval this field. It should be nanoseconds (presumably since epoch). I'll dig more into it.

PR #96, opened 2022-05-19
Extract ibctest.chainSet type to simplify future tests

mark-rushakoff:
The StartChainsAndRelayerFromFactory function, which has now been
renamed to StartChainPairAndRelayer, had some lengthy sections to set up
chains in preparation for tests. The logic in those sections would apply
just the same to three or more chains. We are working towards supporting
tests beyond a single pair of chains, so extract that common behavior
into the new chainSet type.

The chainSet type is unexported for now, as it contains behavior that
was internal to the StartChainsAndRelayerFromFactory function.

### DIFF
--- a/chainset.go
+++ b/chainset.go
@@ -1,0 +1,136 @@
+package ibctest
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/crypto/hd"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/ory/dockertest/v3"
+	"github.com/strangelove-ventures/ibctest/ibc"
+	"golang.org/x/sync/errgroup"
+)
+
+// chainSet is an ordered collection of ibc.Chain,
+// to group methods that apply actions against all chains in the set.
+//
+// The main purpose of the chainSet is to unify test setup when working with any number of chains.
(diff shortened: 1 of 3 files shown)


chainset.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Initialize concurrently calls Initialize against each chain in the set.
+// Each chain may run a docker pull command,
+// so with a cold image cache, running concurrently may save some time.
+func (cs chainSet) Initialize(testName string, homeDir string, pool *dockertest.Pool, networkID string) error {
+	var eg errgroup.Group
+
+	for _, c := range cs {
+		c := c
+		eg.Go(func() error {
```
DavidNix, line 29: This pattern is interesting to me. I almost always avoid shadowing variables and package names in Go. I tend to use: 

```go
for i := range cs {
   c := cs[i]
```

Both have the same result. 

Just an observation, no need to change.

DavidNix (approved):
Thank you for breaking up that large function some.

chainset.go
mark-rushakoff: [Effective Go](https://go.dev/doc/effective_go) states:

> It may seem odd to write `req := req` but it's legal and idiomatic in Go to do this. You get a fresh version of the variable with the same name, deliberately shadowing the loop variable locally but unique to each goroutine.

I think the _idiomatic_ detail specifically applies when it is the first statement inside the loop.

PR #97, opened 2022-05-19
feat: Introduce Packet domain model with validation

DavidNix:

See discussion at closed WIP PR: https://github.com/strangelove-ventures/ibctest/pull/95

In an effort to poll for state, this further adds foundation to ibctest's IBC domain model. This creates a `Packet` and `Transaction` domain entity along with validation for those types. My hope is the validation can help guide others when implementing interfaces.

Additionally removes 1 more method from `ibc.Chain` interface.

* Gas fees may not be the right generalizaton for non-cosmos chains. Refactor is likely.

### DIFF
--- a/relayertest/test.go
+++ b/relayertest/test.go
@@ -177,8 +177,8 @@ func sendIBCTransfersFromBothChainsWithTimeout(
 	})
 
 	require.NoError(t, eg.Wait())
-	require.NoError(t, srcTx.Validate())
-	require.NoError(t, dstTx.Validate())
+	require.NoError(t, srcTx.Validate(), "source ibc transfer tx is invalid")
+	require.NoError(t, dstTx.Validate(), "destination ibc transfer tx is invalid")
 
 	testCase.TxCache = []ibc.Tx{srcTx, dstTx}
 }


chain/tendermint/events.go
```diff
@@ -1,24 +1,22 @@
...
-
 	abcitypes "github.com/tendermint/tendermint/abci/types"
 )
 
 // AttributeValue returns an event attribute value given the eventType and attribute key tuple.
 // In the event of duplicate types and keys, returns the first attribute value found.
-// If not found, ok is false.
-func AttributeValue(events []abcitypes.Event, eventType string, attrKey []byte) (found []byte, ok bool) {
+// If not found, returns empty string and false.
+func AttributeValue(events []abcitypes.Event, eventType, attrKey string) (string, bool) {
```
DavidNix, line 10: Even though the attribute data structure is `{Key: []byte, Value: []byte}`. The consumers I've noticed in cosmos want them as strings.

ibc/packet.go
```diff
@@ -0,0 +1,58 @@
...
+	// The IBC spec does not indicate the unit of time. However, ibc-go's protos define it as nanoseconds.
+	TimeoutTimestamp Nanoseconds
+}
+
+// Validate returns an error if the packet is not well-formed.
+func (packet Packet) Validate() error {
+	var merr error
+	addErr := func(err error) {
+		merr = multierr.Append(merr, err)
+	}
```
mark-rushakoff, line 35: This local function seems unnecessary. But it will probably get inlined during compilation.

```go
// Either of these should be equivalent.
merr = multierr.Append(merr, fmt.Errorf("..."))
multierr.AppendInto(&merr, fmt.Errorf("..."))
```

mark-rushakoff (approved):
Good cleanup.

ibc/packet.go
DavidNix: Stylistically, I felt the local function was easier to read and write, although I was not aware `AppendInto` was part of the package api.

PR #98, opened 2022-05-19
Add TrackSkip method to test reporter

mark-rushakoff:
We now track missing relayer capabilities as the reason when tests are
skipped for that. If and when we skip tests based on label filtering,
that will also use this new method.


DavidNix (approved):
Super clear!

PR #100, opened 2022-05-20
feat: Poll for packet acknowledgements

DavidNix:

Further solidifies the pattern to make `ibc.Chain` simple. Essentially, query some state from the chain; other functions poll, wait, or take other action on the state.

The (hopefully quick) upcoming PR is polling for timeouts. 

Timeouts was not included in this PR because it was already sizable.

The hope is polling for state may improve test suite run times.

* I could not reduce the `ibc.Chain` interface. I explored a few designs (even with generics) but having 2 separate methods for acks and timeouts will probably be easiest. 
* Reminder: The package `test` is a "sibling" package and probably not the right place. I'd prefer the root `ibctest`, but an import cycle currently prevents it. It's not worth fighting the import cycle yet.

### DIFF
--- a/chain/cosmos/codec.go
+++ b/chain/cosmos/codec.go
@@ -30,8 +30,9 @@ func newTestEncoding() simappparams.EncodingConfig {
 
 var (
 	defaultEncoding = newTestEncoding()
-	defaultDecoder  = func() sdk.TxDecoder {
-		cdc := codec.NewProtoCodec(defaultEncoding.InterfaceRegistry)
-		return authTx.DefaultTxDecoder(cdc)
-	}()
 )
+
+func decodeTX(txbz []byte) (sdk.Tx, error) {
+	cdc := codec.NewProtoCodec(defaultEncoding.InterfaceRegistry)
+	return authTx.DefaultTxDecoder(cdc)(txbz)
+}
--- a/chain/cosmos/codec.go
+++ b/chain/cosmos/codec.go
@@ -1,15 +1,18 @@
 package cosmos
 
 import (
+	"github.com/cosmos/cosmos-sdk/codec"
 	"github.com/cosmos/cosmos-sdk/simapp"
 	simappparams "github.com/cosmos/cosmos-sdk/simapp/params"
 	"github.com/cosmos/cosmos-sdk/std"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authTx "github.com/cosmos/cosmos-sdk/x/auth/tx"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	transfertypes "github.com/cosmos/ibc-go/v3/modules/apps/transfer/types"
 	ibctypes "github.com/cosmos/ibc-go/v3/modules/core/types"
 )
 
-func NewTestEncoding() simappparams.EncodingConfig {
+func newTestEncoding() simappparams.EncodingConfig {
 	// core modules
 	cfg := simappparams.MakeTestEncodingConfig()
 	std.RegisterLegacyAminoCodec(cfg.Amino)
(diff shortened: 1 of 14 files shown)


chain/cosmos/query.go
```diff
@@ -0,0 +1,44 @@
...
+
+func findMsgFromTx[T sdk.Msg](tx sdk.Tx) (zero T, _ bool) {
+	for _, m := range tx.GetMsgs() {
+		found, ok := m.(T)
+		if !ok {
+			continue
+		}
+		return found, true
+	}
+	return zero, false
```
DavidNix, line 43: Because we will expose ibctest as a library, using go1.18 probably isn't the best idea (sadly). I wrote this before our decision to make a full-fledged package yesterday. 

This demonstrates generics can really clean up and de-dupe code.

chain/cosmos/codec.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+var (
+	defaultEncoding = newTestEncoding()
+)
+
+func decodeTX(txbz []byte) (sdk.Tx, error) {
+	cdc := codec.NewProtoCodec(defaultEncoding.InterfaceRegistry)
+	return authTx.DefaultTxDecoder(cdc)(txbz)
+}
```
DavidNix, line 38: This weirdness (and above) is how you can take an opaque `[]byte` tx from Tendermint and get a Cosmos tx out of it.

conformance/test.go
```diff
@@ -284,9 +284,9 @@ func TestChainPair(t *testing.T, cf ibctest.ChainFactory, rf ibctest.RelayerFact
 	_, channels, err := ibctest.StartChainPairAndRelayer(t, ctx, rep, pool, network, home, srcChain, dstChain, rf, preRelayerStartFuncs)
 	req.NoError(err, "failed to StartChainPairAndRelayer")
 
-	// TODO poll for acks inside of each testCase `.Config.Test` method instead of just waiting for blocks here
 	// Wait for both chains to produce 10 blocks per test case.
 	// This is long to allow for intermittent retries inside the relayer.
+	// TODO(nix 05-23-2022): Remove once we poll for timeouts, otherwise timeout tests fail
 	req.NoError(test.WaitForBlocks(ctx, 10*len(testCases), srcChain, dstChain), "failed to wait for blocks")
```
DavidNix, line 290: Wanted to highlight my note here. Soon, I should be able to remove the big "wait" here.

ibc/chain.go
```diff
@@ -69,11 +69,8 @@ type Chain interface {
 	// get the fees in native denom for an amount of spent gas
 	GetGasFeesInNativeDenom(gasPaid int64) int64
 
-	// GetPacketAcknowledgement fetches ibc packet ack or an error if not found
-	GetPacketAcknowledgement(ctx context.Context, portID, channelID string, seq uint64) (PacketAcknowledgement, error)
-
-	// GetPacketSequence returns the packet sequence given the transaction's hash
-	GetPacketSequence(ctx context.Context, txHash string) (uint64, error)
+	// Acknowledgements returns all acknowledgements in a block at height
+	Acknowledgements(ctx context.Context, height uint64) ([]PacketAcknowledgement, error)
```
DavidNix, line 73: Originally I had it return just 1 ack in this method vs returning a slice. 

However, given blocks may have many transactions, this may not fit all cases, especially given parallel tests. Totally possible multiple acks are in a block.

test/poll_for_state.go
```diff
@@ -0,0 +1,56 @@
...
+type ChainAcker interface {
+	ChainHeighter
+	Acknowledgements(ctx context.Context, height uint64) ([]ibc.PacketAcknowledgement, error)
+}
+
+// PollForAck attempts to find an acknowledgement containing a packet equal to the packet argument.
+// Polling starts at startHeight and continues until maxHeight. It is safe to call this function even if
+// the chain has yet to produce blocks for the target min/max height range. Polling delays until heights exist
+// on the chain. If no acknowledgement found, returns a not-found error.
+func PollForAck(ctx context.Context, chain ChainAcker, startHeight, maxHeight uint64, packet ibc.Packet) (zero ibc.PacketAcknowledgement, _ error) {
```
DavidNix, line 20: This function went through a lot of iterations. I settled on the min/max variant because it's the most reliable when polling for blocks. 

Otherwise, you risk missing blocks due to timing issues.

mark-rushakoff (approved):
A couple small notes to consider at your discretion, but otherwise I think this is good to go.

test/poll_for_state.go
```diff
@@ -0,0 +1,56 @@
...
+// ChainAcker is a chain that can get its acknowledgements at a specified height
+type ChainAcker interface {
+	ChainHeighter
+	Acknowledgements(ctx context.Context, height uint64) ([]ibc.PacketAcknowledgement, error)
+}
+
+// PollForAck attempts to find an acknowledgement containing a packet equal to the packet argument.
+// Polling starts at startHeight and continues until maxHeight. It is safe to call this function even if
+// the chain has yet to produce blocks for the target min/max height range. Polling delays until heights exist
+// on the chain. If no acknowledgement found, returns a not-found error.
```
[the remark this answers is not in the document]
DavidNix: This may make the case for an explicit `ErrNotFound` or similar. My inclination is that you only need to know if you found acks or not. The reason why is less important. Adding a 3rd bool return argument is unecessarily complex for this use case imo.

DavidNix:
@mark-rushakoff Thanks for the review! I will address the PR comments (all great suggestions) in the next PR.

PR #101, opened 2022-05-20
Rename relayertest package to conformance

mark-rushakoff:
The new name is more accurate because it is testing not only the
relayer, but the interaction of other chains involved too.


DavidNix (approved):
👏 🚢

PR #102, opened 2022-05-20
Initial implementation of ibctest.Interchain

mark-rushakoff:
This PR adds the `Interchain`type, which provides a Builder-pattern interface for constructing a network of arbitrary Chains and Relayers with user-controlled relaying paths.

We only have a single use of the type in `ibctest.StartChainPairAndRelayer` so far, but we will continue to use the type as the set of tests expand; and we will maintain this as an exported API so that other projects can trivially construct interchain definitions for IBC testing.

### DIFF
--- a/world.go
+++ b/world.go
@@ -38,9 +38,11 @@ type relayerPath struct {
 	Path    string
 }
 
-// AddChain adds the given chain with the given name to the World.
-// If the given chain or name already exists, AddChain panics.
-func (w *World) AddChain(chain ibc.Chain, name string) *World {
+// AddChain adds the given chain to the world.
+// If the given chain already exists,
+// or if another chain with the same configured name exists, AddChain panics.
+func (w *World) AddChain(chain ibc.Chain) *World {
+	name := chain.Config().Name
 	for c, n := range w.chains {
 		if c == chain {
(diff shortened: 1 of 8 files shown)


jackzampolin (commented):
Love the direction. This is a very nice API.

world.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/cosmos/cosmos-sdk/types"
+	"github.com/ory/dockertest/v3"
+	"github.com/strangelove-ventures/ibctest/ibc"
+	"github.com/strangelove-ventures/ibctest/testreporter"
+)
+
+// World represents a full IBC network, encompassing a collection of
+// one or more chains, one or more relayer instances, and initial account configuration.
+type World struct {
```
DavidNix, line 17: I could be convinced `World` is the right name. 

But wanted to also suggestion `NewGraph`, `NewNetwork`, or perhaps `NewTestSuite`.

agouin: My vote as proposed by @joeabbey : `NewInterchain`

DavidNix (approved):
LGTM. I didn't see any flags but am excited to use this new API.

PR #103, opened 2022-05-23
Add types and functionality needed to start supporting interchain accounts testing

jtieri:
## Description

This PR adds the necessary types and functionality that should be required to support testing interchain accounts.
This PR *does not* add testing scenarios that simulate the usage of interchain accounts.

## Checklist

## Changes

Added methods to the `Chain` interface for registering an interchain account, querying an interchain account, and sending a bank transfer message to an interchain account.

Added methods for creating clients, creating connections, and querying connections to the `Relayer` interface.

Added a test chain that will utilize a docker image for the binary from the interchain accounts [demo](https://github.com/cosmos/interchain-accounts-demo)

Removed newline from end of error message.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -909,18 +909,22 @@ func (tn *ChainNode) QueryICA(ctx context.Context, connectionID, address string)
 // SendICABankTransfer builds a bank transfer message for a specified address and sends it to the specified
 // interchain account.
 func (tn *ChainNode) SendICABankTransfer(ctx context.Context, connectionID, fromAddr string, amount ibc.WalletAmount) error {
-	msg := fmt.Sprintf(`{
-	"@type":"/cosmos.bank.v1beta1.MsgSend",
-	"from_address":"%s",
-	"to_address":"%s",
-	"amount": [
-	{
-		"denom": "%s",
-		"amount": "%d"
+	msg, err := json.Marshal(map[string]interface{}{
+		"@type":        "/cosmos.bank.v1beta1.MsgSend",
+		"from_address": fromAddr,
+		"to_address":   amount.Address,
+		"amount": []map[string]interface{}{
+			{
+				"denom":  amount.Denom,
+				"amount": amount.Amount,
+			},
+		},
+	})
+	if err != nil {
+		return err
 	}
-	]}`, fromAddr, amount.Address, amount.Denom, amount.Amount)
 
-	command := []string{tn.Chain.Config().Bin, "tx", "intertx", "submit", msg,
+	command := []string{tn.Chain.Config().Bin, "tx", "intertx", "submit", string(msg),
 		"--connection-id", connectionID,
 		"--from", fromAddr,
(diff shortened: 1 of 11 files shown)


agouin (approved):
Looks great!

chain/cosmos/chain_node.go
```diff
@@ -861,3 +862,73 @@ func (tn *ChainNode) logger() *zap.Logger {
...
+
+	exitCode, stdout, stderr, err := tn.NodeJob(ctx, command)
+	if err != nil {
+		return "", dockerutil.HandleNodeJobError(exitCode, stdout, stderr, err)
+	}
+
+	// at this point stdout should look like this:
+	// interchain_account_address: cosmos1p76n3mnanllea4d3av0v0e42tjj03cae06xq8fwn9at587rqp23qvxsv0j
+	// we split the string at the : and then just grab the address before returning.
+	parts := strings.SplitN(stdout, ":", 2)
```
mark-rushakoff, line 905: [`strings.Fields`](https://pkg.go.dev/strings#Fields) could be appropriate here too. Is there no `--json` option for output on this command?

chain/cosmos/chain_node.go
```diff
@@ -861,3 +862,73 @@ func (tn *ChainNode) logger() *zap.Logger {
...
+	return strings.TrimSpace(parts[1]), nil
+}
+
+// SendICABankTransfer builds a bank transfer message for a specified address and sends it to the specified
+// interchain account.
+func (tn *ChainNode) SendICABankTransfer(ctx context.Context, connectionID, fromAddr string, amount ibc.WalletAmount) error {
+	msg := fmt.Sprintf(`{
```
mark-rushakoff, line 912: I think producing this JSON is safe here with valid arguments, but I would feel better about it as 
```go
json.Marshal(map[string]interface{}{
  "@type": "/cosmos.bank.v1beta1.MsgSend",
  "from_address": fromAddr,
  "to_address", amount.Address,
  "amount": []map[string]interface{}{
    map[string]interface{}{
      "denom": amount.Denom,
      "amount": amount.Amount.String(), // Or fmt.Sprintf, whatever is appropriate.
    },
  },
})
```

Just to possibility of producing malformed JSON. If this were in wider use I might define local struct for it, but for a one-off, `map[string]interface{}` seems fine.

mark-rushakoff (approved):

chain/cosmos/chain_node.go
jtieri: Yeah I debated sticking this into a struct but I couldn't think of anywhere else we would use this unless we wanted to support generic bank messages. I'll make the changes so it's at least marshaling and we know its valid JSON.

chain/cosmos/chain_node.go
jtieri: Hmm looking back there is a flag for producing JSON output, must have glossed over it initially.

I'm going to open another PR today that actually adds a test case built on top of this PR so I may actually make this change in that PR.

PR #106, opened 2022-05-24
fix: Increase number of blocks to poll

DavidNix:
Give ourselves more leeway to detect txs created by the relayer.

### DIFF
--- a/conformance/test.go
+++ b/conformance/test.go
@@ -49,6 +49,7 @@ import (
 const (
 	userFaucetFund = int64(10_000_000_000)
 	testCoinAmount = int64(1_000_000)
+	pollHeightMax  = uint64(100)
 )
 
 type RelayerTestCase struct {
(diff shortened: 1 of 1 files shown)


jtieri (approved):
I assume we are doubling this value because by the time we get to the assertions it's possible that the txs we want to make assertions about are too far behind in the history?

DavidNix:
> I assume we are doubling this value because by the time we get to the assertions it's possible that the txs we want to make assertions about are too far behind in the history?

On the contrary, my hunch is the txs have yet to happen. i.e. we're not looking far enough into the future. 

The way the order of operations works:
1. Setup initial test state which includes sending IBC Transfers
2. Relayer starts
3. Now start making assertions and polling. 

During parts of steps 1 & 2, the chains are creating blocks in the background. My hypothesis is if we start polling from a block height from when the relayer starts vs. the ibc send tx, that will give us a better starting point.

PR #107, opened 2022-05-25
Extract relayer.DockerRelayer from rly.CosmosRelayer

mark-rushakoff:
This wrapper type will allow a minimal definition of the hermes relayer,
or potentially any other relayer implementation, running inside Docker.

This will also be helpful if we end up having to support different
versions of rly or hermes that have a different CLI API.

For #3.


DavidNix (approved):
🚢

agouin (approved):
Looks great!

mark-rushakoff:
@joeabbey has been studying the differences between the CLI for rly and hermes, so I am going to wait for his review before merging this one, so that we can be more sure the DockerRelayer type can handle both relayers.

joeabbey (approved):

PR #108, opened 2022-05-25
fix: More descriptive polling errors

DavidNix:

"Not found" was an okay first pass. But for debugging, it's not super useful. The error message (although verbose) now inclues the error, the target packet, and all the resources searched for the target packet.

Screen Shot 2022-05-25 at 5 15 19 PM

* I think verbose error messages are ok (even preferred) for ibctest.

### DIFF
--- a/conformance/test.go
+++ b/conformance/test.go
@@ -49,7 +49,7 @@ import (
 const (
 	userFaucetFund = int64(10_000_000_000)
 	testCoinAmount = int64(1_000_000)
-	pollHeightMax  = uint64(100)
+	pollHeightMax  = uint64(50)
 )
 
 type RelayerTestCase struct {
--- a/test/poll_for_state.go
+++ b/test/poll_for_state.go
@@ -143,6 +143,7 @@ func (pe *pollError) Unwrap() error {
 	return pe.error
 }
 
+// Format is expected to be used by testify/require which prints errors via %+v
 func (pe *pollError) Format(s fmt.State, verb rune) {
 	if verb != 'v' && !s.Flag('+') {
 		fmt.Fprint(s, pe.error.Error())
--- a/test/poll_for_state.go
+++ b/test/poll_for_state.go
@@ -137,7 +137,7 @@ func (pe *pollError) Error() string {
 		searched = "(none)"
 	}
 	target := spew.Sdump(pe.targetPacket)
-	return fmt.Sprintf("error: %s\n- target packet: %s\n- searched:\n%+v", pe.error, target, searched)
+	return fmt.Sprintf("%s\n- target packet:\n%s\n- searched:\n%+v", pe.error, target, searched)
 }
 func (pe *pollError) SetErr(err error) {
 	pe.error = err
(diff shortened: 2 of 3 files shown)


conformance/test.go
```diff
@@ -49,7 +49,7 @@ import (
 const (
 	userFaucetFund = int64(10_000_000_000)
 	testCoinAmount = int64(1_000_000)
-	pollHeightMax  = uint64(100)
+	pollHeightMax  = uint64(50)
```
DavidNix, line 52: This was too high slowing down the test suite unecessarily.

test/poll_for_state.go
```diff
@@ -109,9 +117,43 @@ func (p blockPoller) findTimeout(ctx context.Context, height uint64, packet ibc.
...
+		p.pollErr.PushSearched(t)
 		if t.Packet.Equal(packet) {
 			return t, nil
 		}
 	}
 	return zero, ErrNotFound
 }
+
+type pollError struct {
+	error
```
mark-rushakoff, line 129: If I'm following this correctly, we embed `error` here so that the `.Error()` method returns `p.error.Error()`; the only special case of presentation is the `+v` verb, which emits the expanded form of target packet and searched; and it looks like `+v` is used internally by `require` when formatting errors.

test/poll_for_state_test.go
```diff
@@ -142,7 +150,6 @@ func TestPollForTimeout(t *testing.T) {
 		_, err := PollForTimeout(ctx, &chain, 1, 3, ibc.Packet{})
 
 		require.Error(t, err)
-		require.EqualError(t, err, "not found")
 		require.ErrorIs(t, err, ErrNotFound)
```
mark-rushakoff, line 153: Glad to see this line still works.

test/poll_for_state.go
```diff
@@ -109,9 +117,43 @@ func (p blockPoller) findTimeout(ctx context.Context, height uint64, packet ibc.
...
+	return pe.error
+}
+
+func (pe *pollError) Format(s fmt.State, verb rune) {
+	if verb != 'v' && !s.Flag('+') {
+		fmt.Fprint(s, pe.error.Error())
+		return
+	}
+
+	searched := strings.Join(pe.searchedPackets, "\n")
```
mark-rushakoff, line 152: Technically this does not match the documented purpose of `%+v`:

```
%v	the value in a default format
	when printing structs, the plus flag (%+v) adds field names
```

but given the limited scope we are using it, this is fine. A comment mentioning the intended use with `require` wouldn't hurt here.

mark-rushakoff (approved):
I agree with increased verbosity for test failures.

test/poll_for_state.go
DavidNix: I stole this pattern from `pkg/errors` (now archived which makes me sad). In that package, it would print the stack trace.

Good point on `require`, will add.

test/poll_for_state.go
DavidNix: > it looks like +v is used internally by require when formatting errors.

Yeah, that could be brittle if testify changes its behavior. But I thought it was nice to let the caller decide if it wants a short or very long error. 

Fwiw, I often embed with private types for convenience, especially if it's an adapter. But I almost never embed with public/exported types. But happy to never embed if we decide that's our code style.

PR #109, opened 2022-05-26
Add godoc badge and more package documentation

mark-rushakoff:
Add the godoc badge to the README so that it's one click to view the online docs. Add more package documentation so the directory list at the bottom of the godoc is more informative. Add move dockerutil and chain/tendermint to internal packages so that we are more explicitly excluding them from any public API.


DavidNix (approved):
Good cleanup. I bet there's more that can move to `internal` eventually.

PR #110, opened 2022-05-26
Add relayer setup conformance test

mark-rushakoff:
The test is not yet complete because ibc.Relayer does not yet have a
CreateChannel method -- this was likely an oversight as all other tests
so far simply use CreateLink to create clients, connections, and
channels in one shot.

But this is another example of the Interchain type greatly simplifying
setup for a conformance test.


DavidNix (approved):
Bonus: This test is also a nice example how to use the lib.

PR #111, opened 2022-05-26
Add CreateChannels method to ibc.Relayer

mark-rushakoff:
This was one missing piece in the relayer setup conformance test.


jtieri (approved):
🚀

PR #117, opened 2022-06-01
WIP: Interchain Accounts test case

jtieri:
## Description

This PR adds a test case for registering an interchain account utilizing the testing binary provided [here](https://github.com/cosmos/interchain-accounts-demo)

## Checklist

- [ ] I have made sure this PR is ready to merge
- [ ] I have made sure that I have assigned reviewers related to this project

## Changes

- Add a test case for registering a new interchain account

## Testing

- Checkout the branch in the relayer [here](https://github.com/cosmos/relayer/pull/767)
- From the root directory of the relayer `cd ibctest`
- `go test -run TestInterchainAccounts`

## Notes

The relayer does not currently have support for interchain accounts in the main branch because we are waiting for upstream changes in ibc-go so that we can parse all relevant information for an ordered channel from the tx events. This makes it necessary to use a specific branch on the relayer that has support for finishing channel handshakes. Ideally we should not merge this till we can go back to using the main branch on the relayer.


boojamya:
Closes: https://github.com/strangelove-ventures/ibctest/issues/13

jtieri:
Closing this as it will be replace with a new PR shortly.

PR #118, opened 2022-06-01
feat: Track block history for cosmos chains in sqlite database

DavidNix:

Saves cosmos block data to a sqlite database.

Screen Shot 2022-06-01 at 4 49 48 PM

This feature can be optionally configured when using the `Interchain` API as a library. 

* Using raw sqlite is going to be rough. I'll have a quick followup PR for a helper script until we can put a UI on top of it.
* Removes JSON pretty printing for now while I think on how to best achieve that. 
* I recommend we move the `logger` to the `Interchain` library options as well.

### DIFF
--- a/chainset.go
+++ b/chainset.go
@@ -3,9 +3,7 @@ package ibctest
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
-	"strings"
 	"sync"
 	"time"
 
@@ -108,37 +106,58 @@ func (cs chainSet) Start(ctx context.Context, testName string, additionalGenesis
 }
 
 // TrackBlocks initializes database tables and polls for transactions to be saved in the database.
+// This method is a nop if dbFile is blank.
 // Expected to be called after Start.
-func (cs chainSet) TrackBlocks(ctx context.Context, testName, dbFile, gitSha string) (io.Closer, error) {
+func (cs chainSet) TrackBlocks(ctx context.Context, testName, dbFile, gitSha string) error {
 	if len(dbFile) == 0 {
-		return io.NopCloser(strings.NewReader("nop")), nil
+		// nop
+		return nil
 	}
 
 	db, err := blockdb.ConnectDB(ctx, dbFile, 10)
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect to sqlite database at %s: %w", dbFile, err)
+		return fmt.Errorf("connect to sqlite database %s: %w", dbFile, err)
+	}
+
+	if err := blockdb.Migrate(db); err != nil {
+		return fmt.Errorf("migrate sqlite database %s; deleting file recommended: %w", dbFile, err)
+	}
+
+	if len(gitSha) == 0 {
+		gitSha = "unknown"
 	}
 
 	testCase, err := blockdb.CreateTestCase(ctx, db, testName, gitSha)
 	if err != nil {
 		_ = db.Close()
-		return nil, fmt.Errorf("failed to create test case in sqlite database: %w", err)
+		return fmt.Errorf("create test case in sqlite database: %w", err)
 	}
 
+	var eg errgroup.Group
 	for c := range cs {
+		c := c
 		name := c.Config().Name
 		finder, ok := c.(blockdb.TxFinder)
 		if !ok {
-			// TODO: Use a logger here.
-			fmt.Fprintf(os.Stderr, `Chain %s is not configured to save blocks; must implement "FindTxs(ctx context.Context, height uint64) ([][]byte, error)"`, name)
-			continue
+			// TODO (nix - 6/1/22) Use logger here?
+			fmt.Fprintf(os.Stderr, `Chain %s is not configured to save blocks; must implement "FindTxs(ctx context.Context, height uint64) ([][]byte, error)"`+"\n", name)
+			return nil
 		}
-		chaindb, err := testCase.AddChain(ctx, name)
-		if err != nil {
-			return nil, fmt.Errorf("failed track chain %s: %w", name, err)
-		}
-		go blockdb.NewPoller(finder, chaindb, 100*time.Millisecond, zap.NewNop()).Poll(ctx)
+		eg.Go(func() error {
+			chaindb, err := testCase.AddChain(ctx, name)
+			if err != nil {
+				return fmt.Errorf("add chain %s: %w", name, err)
+			}
+			blockdb.NewPoller(finder, chaindb, 100*time.Millisecond, zap.NewNop()).Poll(ctx)
+			return nil
+		})
 	}
 
-	return db, nil
+	go func() {
+		// TODO (nix - 6/1/22) May leak file descriptor. Interchain may need a Close() method.
+		_ = eg.Wait()
+		_ = db.Close()
+	}()
+
+	return nil
 }
--- a/chainset.go
+++ b/chainset.go
@@ -3,11 +3,17 @@ package ibctest
 import (
 	"context"
 	"fmt"
+	"io"
+	"os"
+	"strings"
 	"sync"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/types"
 	"github.com/ory/dockertest/v3"
 	"github.com/strangelove-ventures/ibctest/ibc"
+	"github.com/strangelove-ventures/ibctest/internal/blockdb"
+	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -99,4 +105,40 @@ func (cs chainSet) Start(ctx context.Context, testName string, additionalGenesis
 	}
 
 	return eg.Wait()
+}
+
+// TrackBlocks initializes database tables and polls for transactions to be saved in the database.
+// Expected to be called after Start.
+func (cs chainSet) TrackBlocks(ctx context.Context, testName, dbFile, gitSha string) (io.Closer, error) {
+	if len(dbFile) == 0 {
+		return io.NopCloser(strings.NewReader("nop")), nil
+	}
+
+	db, err := blockdb.ConnectDB(ctx, dbFile, 10)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to sqlite database at %s: %w", dbFile, err)
+	}
+
+	testCase, err := blockdb.CreateTestCase(ctx, db, testName, gitSha)
+	if err != nil {
+		_ = db.Close()
+		return nil, fmt.Errorf("failed to create test case in sqlite database: %w", err)
+	}
+
+	for c := range cs {
+		name := c.Config().Name
+		finder, ok := c.(blockdb.TxFinder)
+		if !ok {
+			// TODO: Use a logger here.
+			fmt.Fprintf(os.Stderr, `Chain %s is not configured to save blocks; must implement "FindTxs(ctx context.Context, height uint64) ([][]byte, error)"`, name)
+			continue
+		}
+		chaindb, err := testCase.AddChain(ctx, name)
+		if err != nil {
+			return nil, fmt.Errorf("failed track chain %s: %w", name, err)
+		}
+		go blockdb.NewPoller(finder, chaindb, 100*time.Millisecond, zap.NewNop()).Poll(ctx)
+	}
+
+	return db, nil
 }
--- a/chainset.go
+++ b/chainset.go
@@ -114,7 +114,7 @@ func (cs chainSet) TrackBlocks(ctx context.Context, testName, dbFile, gitSha str
 		return nil
 	}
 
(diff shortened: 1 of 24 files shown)


internal/blockdb/chain.go
```diff
@@ -1,39 +1,43 @@
...
 	}
+	return h.Sum(nil)
+}
+
+// SaveBlock tracks a block at height with its transactions.
+// This method is idempotent and can be safely called multiple times with the same arguments.
+// The txs should be human-readable.
+func (chain *Chain) SaveBlock(ctx context.Context, height int, txs [][]byte) error {
+	k := fmt.Sprintf("%d-%x", height, transactions(txs).hash())
+	_, err, _ := chain.single.Do(k, func() (interface{}, error) {
```
DavidNix, line 34: Single flight maybe overkill now. My original thought was this method may be called a lot. But I kept it because it should work fine.

chainset.go
```diff
@@ -100,3 +104,61 @@ func (cs chainSet) Start(ctx context.Context, testName string, additionalGenesis
...
+	for c := range cs {
+		c := c
+		name := c.Config().Name
+		finder, ok := c.(blockdb.TxFinder)
+		if !ok {
+			fmt.Fprintf(os.Stderr, `Chain %s is not configured to save blocks; must implement "FindTxs(ctx context.Context, height uint64) ([][]byte, error)"`+"\n", name)
+			return nil
+		}
+		eg.Go(func() error {
+			chaindb, err := testCase.AddChain(ctx, name)
```
DavidNix, line 147: My assumption here is the chain name can be arbitrary in the event of same chains like gaia <-> gaia. I confirmed in our #engineering channel that testing the same chains will be common. 

I know in Cosmos chain name is canonical in the Chain Registry. However, in this context, I'm assuming it can be arbitray. E.g. `gaia1`, `gaia2`. 

May I should instead use the chain label?

internal/blockdb/collect.go
```diff
@@ -0,0 +1,70 @@
...
+	tick := time.NewTicker(p.rate)
+	defer tick.Stop()
+	var height uint64 = 1
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-tick.C:
+			if err := p.saveTxsForHeight(ctx, height); err != nil {
+				p.log.Info("Failed to save transactions", zap.Error(err), zap.Uint64("height", height))
```
mark-rushakoff, line 52: Might be nice to avoid logging if we can detect that we are trying to access a block that isn't available yet. It might be `ErrHeightExceedsChainHead` or `ErrHeightNotAvailable ` from https://pkg.go.dev/github.com/tendermint/tendermint@v0.35.5/rpc/coretypes#pkg-variables

internal/blockdb/collect_test.go
```diff
@@ -0,0 +1,124 @@
...
+	return f(ctx, height)
+}
+
+type mockBlockSaver func(ctx context.Context, height int, txs [][]byte) error
+
+func (f mockBlockSaver) SaveBlock(ctx context.Context, height int, txs [][]byte) error {
+	return f(ctx, height, txs)
+}
+
+func TestCollector_Collect(t *testing.T) {
```
mark-rushakoff, line 28: Slightly preferable to have the outer test and/or its subtests can use `t.Parallel` since these are exercising asychronous communication across goroutines, even with the nanosecond duration.

internal/blockdb/collect_test.go
```diff
@@ -0,0 +1,124 @@
...
+				panic("zero height")
+			}
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+				if height > 3 {
+					return nil, nil
+				}
+				return [][]byte{[]byte(strconv.FormatUint(height, 10))}, nil
```
mark-rushakoff, line 41: `[]byte(strconv.FormatUint(h, 10))` can be `strconv.AppendUint(nil, h, 10)`.

test_setup.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/strangelove-ventures/ibctest/testreporter"
 )
 
+var gitSha = "unknown"
```
mark-rushakoff, line 19: A comment indicating this is set in the make target would be helpful.

internal/blockdb/sql_test.go
```diff
@@ -26,12 +29,9 @@ func migratedDB() *sql.DB {
...
 
 func TestConnectDB(t *testing.T) {
-	f, err := os.CreateTemp("", t.Name())
-	require.NoError(t, err)
-	defer f.Close()
-	defer os.RemoveAll(f.Name())
-
-	db, err := ConnectDB(context.Background(), f.Name(), 10)
+	file := filepath.Join(os.TempDir(), strconv.FormatInt(time.Now().UnixMilli(), 10), "test", t.Name()+".db")
+	defer os.RemoveAll(file)
```
mark-rushakoff, line 33: Not using t.TempDir() here, combined with only the call to `os.RemoveAll(file)`, will leave the `$TMPDIR/$(now)/test` directory sitting around after test. `t.TempDir()` makes more sense than `os.TempDir()` here.

internal/blockdb/sql.go
```diff
@@ -4,23 +4,33 @@ import (
...
 
-// ConnectDB connects to the sqlite database at filePath with max connections set to maxConns.
+// ConnectDB connects to the sqlite database at databaseFile.
 // Pings database once to ensure connection.
-// Pass :memory: as filePath for in-memory database.
-func ConnectDB(ctx context.Context, filePath string, maxConns int) (*sql.DB, error) {
-	db, err := sql.Open("sqlite", filePath)
+// Creates directory path via MkdirAll.
+// Pass :memory: as databaseFile for in-memory database.
+func ConnectDB(ctx context.Context, databaseFile string) (*sql.DB, error) {
```
mark-rushakoff, line 17: Can the parameter `databaseFile` be renamed to `databasePath` since the string is a path, not a file object?

interchain.go
```diff
@@ -137,6 +137,13 @@ type InterchainBuildOptions struct {
 	// but it does still configure keys and wallets for declared relayer-chain links.
 	// This is useful for tests that need lower-level access to configuring relayers.
 	SkipPathCreation bool
+
+	// Optional. Git sha for test invocation. Once Go 1.18 supported,
+	// may be deprecated in favor of runtime/debug.ReadBuildInfo.
+	GitSha string
+
+	// If set, saves block history to a sqlite3 database to aid debugging.
+	BlocksDatabaseFile string
```
mark-rushakoff, line 146: Can this field be named `BlocksDatabasePath` since it is a path, not a file object?

chainset.go
```diff
@@ -100,3 +104,61 @@ func (cs chainSet) Start(ctx context.Context, testName string, additionalGenesis
 
 	return eg.Wait()
 }
+
+// TrackBlocks initializes database tables and polls for transactions to be saved in the database.
+// This method is a nop if dbFile is blank.
+// Expected to be called after Start.
+func (cs chainSet) TrackBlocks(ctx context.Context, testName, dbFile, gitSha string) error {
+	if len(dbFile) == 0 {
```
mark-rushakoff, line 112: Comparing string length to zero is valid and correct of course, but I find it less direct to read than `dbFile == ""`. Seeing the zero length comparison tricks me into thinking `dbFile` is a slice.

interchain.go
```diff
@@ -145,7 +152,7 @@ type InterchainBuildOptions struct {
 // Calling Build more than once will cause a panic.
 func (ic *Interchain) Build(ctx context.Context, rep *testreporter.RelayerExecReporter, opts InterchainBuildOptions) error {
 	if ic.built {
-		panic(fmt.Errorf("Build called more than once"))
+		panic(fmt.Errorf("build called more than once"))
```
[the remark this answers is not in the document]
DavidNix: Linter didn't catch it. Happy to change it to the more descriptive `Interchain.Build`. I always lowercase errors even if it's referring to an exported method.

internal/blockdb/chain.go
```diff
@@ -3,23 +3,32 @@ package blockdb
...
 }
 
-// Txs are transactions expected to be marshalled JSON.
-type Txs [][]byte
+type transactions [][]byte
 
-// TraceBlock tracks a block at height with its transactions.
-// This method is idempotent and can be safely called multiple times with the same arguments.
-func (chain *Chain) TraceBlock(ctx context.Context, height int, txs Txs) error {
+func (txs transactions) Hash() []byte {
```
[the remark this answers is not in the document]
DavidNix: TIL about FNV!

With unexported types (as in this case), I typically use exported (uppercase) methods to indicate other callers in the same package are allowed to use these methods. With unexported (lowercase) methods, I typically use that to signal that only the type itself is allowed to use those methods.

internal/blockdb/chain.go
```diff
@@ -28,12 +37,27 @@ func (chain *Chain) TraceBlock(ctx context.Context, height int, txs Txs) error {
 	buf := new(bytes.Buffer)
 	for i, tx := range txs {
 		if err := json.Indent(buf, tx, "", "  "); err != nil {
-			return fmt.Errorf("block %d: tx %d: malformed json: %w", height, i, err)
+			jsonTxs[i] = string(tx)
```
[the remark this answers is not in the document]
DavidNix: I'm changing strategy here and allowing any arbitrary data. I think another layer will deal with whether or not it's JSON.

With Cosmos, especially, it's super convoluted how to deserialize tx (then reserliaze as JSON). Very error prone with its strange codec api. So I'm letting this be optimisic and even if not json, let's capture it.

internal/blockdb/chain_test.go
```diff
@@ -32,7 +32,7 @@ func TestChain_TraceBlock(t *testing.T) {
 
 		chain := validChain(t, db)
 
-		err := chain.TraceBlock(ctx, 5, Txs{tx1, tx2})
+		err := chain.SaveBlock(ctx, 5, transactions{tx1, tx2})
```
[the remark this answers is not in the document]
DavidNix: Whoops, good catch.

internal/blockdb/collect.go
```diff
@@ -0,0 +1,70 @@
...
+
+// TxFinder finds transactions given block at height.
+type TxFinder interface {
+	// FindTxs transactions should be in a human-readable format, preferably json.
+	FindTxs(ctx context.Context, height uint64) ([][]byte, error)
+}
+
+// BlockSaver saves transactions for block at height.
+type BlockSaver interface {
+	SaveBlock(ctx context.Context, height int, txs [][]byte) error
```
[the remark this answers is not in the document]
DavidNix: I don't think sqlite has a uint64 type. Tendermint also doesn't treat it consistently; some rpc methods want int64. 

Fwiw, I think uint64 is the correct data type. Height can never be 0 and in theory can increase forever.  

But at least in this interface I wll make uint64.

internal/blockdb/collect.go
DavidNix: True, but if not a tendermint chain...

internal/blockdb/collect_test.go
DavidNix: Yup, I need to get in the habit of that. Go tests run so fast (with caching) that I often forget about t.Parallel.

internal/blockdb/sql_test.go
DavidNix: Oh I see: `t.TempDir()`. TIL; that's a great feature! I was getting confused with `os.TempDir`. A couple characters make all the difference.

chainset.go
DavidNix: Ha, a matter of preference as I prefer `len(dbFile) == 0` for strings and bytes.

internal/blockdb/chain.go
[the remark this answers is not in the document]
mark-rushakoff: I overlooked `transactions` being an unexported type here. Exported makes sense.

agouin (approved):
Looks great! Excited to have this for debugging.

DavidNix:
@mark-rushakoff Ready for 👀 again. I also added some suggestions from Andrew from our Discord conversation.

mark-rushakoff (approved):
Looks good -- would be great to get some documentation, like a quick example of how to use this to dig into a test failure, in another PR soon.

DavidNix:
Fyi, I changed the foreign keys to have `fk_` prefix. It's a common pattern. I don't use it because I guess I'm still prisoner to the way ActiveRecord does things. 

In any case, adding a `chain_id` column on `chain` table which means the ID as in the Chain Registry forced that refactor. Otherwise the schema looks confusing and that column looks like a foreign key even though it is not.

PR #119, opened 2022-06-02
Support multiple chains with same name on Interchain

mark-rushakoff:
Key by the chain ID instead of the name, so that multiple instances of
the same chain name can coexist. This is for supporting future tests
that need to support multiple instances of the same chain, typically
while testing against a new feature unique to that chain.

Also move the general ibctest.GetChain functionality into the
BuiltinChainFactoryEntry type, which may warrant a rename soon.

And add a nop test reporter to simplify tests that use APIs needing a
test reporter, but that are safe to discard test reporter output.

Lastly, fix a small bug where concurrent tests would try to create the
same-named Docker network due to both tests using the default random
seed.


agouin (approved):
Looks good!

jtieri (approved):
🚢

PR #120, opened 2022-06-02
DockerRelayer options

agouin:
Renames `ibc.ChainDockerImage` to `ibc.DockerImage` since it can also apply to the relayer.

Adds options to customize `DockerRelayer`:

- `RelayerOptionDockerImage` to use a custom docker image
- `RelayerOptionImagePull` to optionally disable image pull (helpful for local relayer images)

### DIFF
--- a/relayer/docker.go
+++ b/relayer/docker.go
@@ -68,7 +68,10 @@ func NewDockerRelayer(log *zap.Logger, testName, home string, pool *dockertest.P
 		}
 	}
 
-	relayer.pullContainerImageIfNecessary(relayer.containerImage())
+	containerImage := relayer.containerImage()
+	if err := relayer.pullContainerImageIfNecessary(containerImage); err != nil {
+		log.Error("Error pulling container image", zap.String("repository", containerImage.Repository), zap.String("version", containerImage.Version), zap.Error(err))
+	}
 
 	return &relayer
 }
--- a/relayer/docker.go
+++ b/relayer/docker.go
@@ -31,6 +31,9 @@ type DockerRelayer struct {
 
 	testName string
 
+	customImage *ibc.DockerImage
+	pullImage   bool
+
 	// The container created by StartRelayer.
 	container *docker.Container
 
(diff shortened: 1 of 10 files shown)


relayer/docker.go
```diff
@@ -204,14 +226,28 @@ func (r *DockerRelayer) StopRelayer(ctx context.Context, rep ibc.RelayerExecRepo
...
 }
 
-func (r *DockerRelayer) createNodeContainer(pathName string) error {
-	err := r.pool.Client.PullImage(docker.PullImageOptions{
-		Repository: r.c.ContainerImage(),
-		Tag:        r.c.ContainerVersion(),
-	}, docker.AuthConfiguration{})
-	if err != nil {
-		return err
+func (r *DockerRelayer) containerImage() (res ibc.DockerImage) {
```
mark-rushakoff, line 229: Named return value is unnecessary here.

relayer/options.go
```diff
@@ -0,0 +1,31 @@
+package relayer
+
+import (
+	"github.com/strangelove-ventures/ibctest/ibc"
+)
+
+type RelayerOption interface{}
```
mark-rushakoff, line 7: For this pattern, people often add a sentinel, unexported method to the interface, such as

```go
type RelayerOption interface {
  relayerOption()
}
```

And then you would add the `relayerOption()` method, as a no-op,  to `RelayerOptionDockerImage` and `RelayerOptionImagePull`.

The advantage of this approach is that the compiler can be more restrictive about what values to allow than a type that just matches `interface{}`.

mark-rushakoff (commented):
The approach looks reasonable here.

relayer/options.go
mark-rushakoff, line 7: Godoc on the type wouldn't hurt either.

mark-rushakoff (approved):
Looks like this went out of draft while I was leaving the previous review.

relayer/docker.go
agouin: good catch. done

DavidNix (approved):

PR #122, opened 2022-06-03
Enforce chain name uniqueness on Interchain

### LINKED ISSUE #116: Support duplicate chains on Interchain
There are some test cases where it is helpful to run two independent but otherwise identical chains, e.g. two `gaia` instances. Right now, if you attempt to do this, the Interchain will reject it due to a duplicate name. Then if you allow the Interchain to accept it, the Docker setup will reference the `chain.Config().Name` from a few places, so the fix is more subtle than just adjusting the Interchain.

mark-rushakoff:
Also validate that AddChain and AddRelayer are not called with nil
values, and add tests to ensure the Add functions panic as intended upon
programmer error.

Closes #116.

### DIFF
--- a/chainfactory.go
+++ b/chainfactory.go
@@ -44,7 +44,18 @@ type BuiltinChainFactory struct {
 
 // BuiltinChainFactoryEntry describes a chain to be returned from an instance of BuiltinChainFactory.
 type BuiltinChainFactoryEntry struct {
-	Name          string
+	// Name refers to the keys in builtinChainConfigs.
+	// If the Name does not reference an existing chain config,
+	// retrieving the chain will result in an error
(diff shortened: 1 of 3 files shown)


agouin (approved):
Looks good!

DavidNix (approved):
LGTM. 

I do wonder if `Name`, `NameOverride`, and `ChainID` will get confusing. 

My initial thought is why not require only 1 canonical, unique identifier which to me is the ChainID. And then we don't need Name or NameOverride.

I know we use `Name` to construct Heighliner docker images, but I don't think it's too much to ask the user to configure with the full docker repo/tag. 

All of the above thoughts for the future, not blocking this PR at all!

PR #125, opened 2022-06-03
Add ChainSpec type for single way of configuring chains

mark-rushakoff:
The new ChainSpec type can easily represent a simple builtin chain, but
can also handle overriding any other field. This keeps test definitions
minimized, and also allows the test matrix to have just one ChainSet
field rather than a separate ChainSet and CustomChainSet.

I kept the BuiltinChainFactory type. The existing ChainFactory type
remains as an interface for now despite ibctest providing only one
implementation today.


chainfactory.go
```diff
@@ -108,107 +53,84 @@ var builtinChainConfigs = map[string]ibc.ChainConfig{
...
+
 	return chains, nil
 }
 
-func (f *BuiltinChainFactory) Name() string {
-	parts := make([]string, len(f.entries))
-	for i, e := range f.entries {
-		parts[i] = e.Name + "@" + e.Version
+func buildChain(log *zap.Logger, testName string, cfg ibc.ChainConfig, numValidators, numFullNodes *int) (ibc.Chain, error) {
+	nv := 2
```
agouin, line 88: maybe this and `nf` could be defined slightly further up in the file as something like `defaultNumValidators`?

agouin, line 88: maybe this and `nf` could be defined slightly further up in the file as something like `defaultNumValidators`?

interchain_test.go
```diff
@@ -24,11 +24,11 @@ func TestInterchain_DuplicateChain(t *testing.T) {
 	home := t.TempDir()
 	pool, network := ibctest.DockerSetup(t)
 
-	cf := ibctest.NewBuiltinChainFactory([]ibctest.BuiltinChainFactoryEntry{
+	cf := ibctest.NewBuiltinChainFactory(zaptest.NewLogger(t), []*ibctest.ChainSpec{
```
agouin, line 27: Might be cool to also offer a builder pattern for the `ibctest.ChainSpec`, but also might be overkill since we are just assembling a single struct.

agouin, line 27: Might be cool to also offer a builder pattern for the `ibctest.ChainSpec`, but also might be overkill since we are just assembling a single struct.

agouin (approved):

agouin:
Great consolidation of the chain factory allowing minimal to full customization.

chainspec_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/strangelove-ventures/ibctest"
+	"github.com/strangelove-ventures/ibctest/ibc"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainSpec_Config(t *testing.T) {
```
DavidNix, line 13: Thanks for the tests!

chainfactory.go
```diff
@@ -108,107 +53,89 @@ var builtinChainConfigs = map[string]ibc.ChainConfig{
...
-func NewCustomChainFactory(entries []CustomChainFactoryEntry, logger *zap.Logger) *CustomChainFactory {
-	return &CustomChainFactory{entries: entries, log: logger}
-}
 
-func (e CustomChainFactoryEntry) GetChain(testName string, log *zap.Logger) (ibc.Chain, error) {
-	switch e.Config.Type {
+	switch cfg.Type {
 	case "cosmos":
-		return cosmos.NewCosmosChain(testName, e.Config, e.NumValidators, e.NumFullNodes, log), nil
+		return cosmos.NewCosmosChain(testName, cfg, nv, nf, log), nil
```
DavidNix, line 104: I keep thinking how to remove the import cycle caused by this (and potentially `penumbra` package) line. I'd love the helpers in the `test` package to be at root `ibctest`. But importing `cosmos` in `ibctest` causes the cycle because cosmos uses the helpers like `WaitForBlocks`

chainspec_test.go
```diff
@@ -0,0 +1,166 @@
...
+		s := ibctest.ChainSpec{
+			ChainName: "mychain",
+
+			ChainConfig: ibc.ChainConfig{
+				Type: "cosmos",
+				// Skip Name, as that is intended to be inherited from ChainName.
+				ChainID: "mychain-123",
+				Images: []ibc.DockerImage{
+					{Repository: "docker.example.com", Version: "latest"},
+				},
```
DavidNix, line 35: I know this was part of the original design, but I'm surprised `Images` is an array vs. one image. However, it's used for penumbra which also needs a tendermint image. 

No need for changes, just a general comment. 

My first instinct is you can't configure highly custom, one-offs like penumbra. It must all be implemented in code, sidestepping this config, which may help simplify the config here. Granted, there may be a common pattern with tendermint as a "sidecar" as seen with penumbra. 

Also (from the previous pattern) comma delimited docker image version is not clear.

chainspec.go
```diff
@@ -0,0 +1,141 @@
...
+	"sort"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"github.com/strangelove-ventures/ibctest/ibc"
+)
+
+// ChainSpec is a wrapper around an ibc.ChainConfig
+// that allows callers to easily reference one of the built-in chain configs
```
DavidNix, line 15: nit:

Curious if `predefined` vs. `built-in` may be clearer?

`predefined chain config` vs. `built-in`. The word "built-in" wasn't as clear to me. But I'm perhaps overthinking this.

chainspec.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// How many validators and how many full nodes to use
+	// when instantiating the chain.
+	// If unspecified, NumValidators defaults to 2 and NumFullNodes defaults to 1.
+	NumValidators, NumFullNodes *int
+
+	// Generate the automatic suffix on demand when needed.
+	autoSuffixOnce sync.Once
+	autoSuffix     string
+}
```
DavidNix, line 45: Which one of these fields (if any) is guaranteed to be unique per test invocation?

cmd/ibctest/example_matrix_custom.json
```diff
@@ -1,49 +1,46 @@
 {
   "Relayers": ["rly"],
 
-  "CustomChainSets": [
+  "ChainSets": [
     [
       {
         "NumValidators": 2,
         "NumFullNodes": 1,
-        "Config": {
```
DavidNix, line 9: 👏

interchain_test.go
DavidNix: +1 Builder pattern *may* be more friendly for those using it as a lib. For instance, it may omit the need for the `MergeChainSpecConfig`. I would say out of scope for this PR though.

PR #131, opened 2022-06-07
Add test ensuring git sha can be omitted from interchain build options

mark-rushakoff:
When importing the current version of ibctest into relayer, this error occurs if the BlockDatabaseFile field is provided but the GitSha is empty:

```
failed to track blocks: migrate sqlite database :memory:; deleting file recommended: upsert schema_version with git sha : constraint failed: CHECK constraint failed: length(git_sha) > 0 (275)
```

Default the git sha to `unknown` before attempting to add it to `schema_version` table, so that the GitSha is optional as documented.


DavidNix (approved):
Great catch and test coverage.

PR #132, opened 2022-06-08
Add tx_flattened view to blockdb

mark-rushakoff:
This view materializes the transactions, joined with blocks, chains, and
test cases.

This is a helpful view for users who are directly interacting with the
sqlite database.

I intend to follow this up with another view that relies on this view,
and parses relevant client/connection/channel details from the tx
objects using sqlite's json functions.

### DIFF
--- a/internal/blockdb/migrate.go
+++ b/internal/blockdb/migrate.go
@@ -80,5 +80,27 @@ ON CONFLICT(git_sha) DO UPDATE SET git_sha=git_sha`, nowRFC3339(), gitSha)
 		return fmt.Errorf("create table tx: %w", err)
 	}
 
+	_, err = db.Exec(`CREATE VIEW IF NOT EXISTS tx_flattened AS
+SELECT
+  test_case.id as test_case_id
+  , test_case.created_at as test_case_created_at
+  , test_case.name as test_case_name
+  , chain.id as chain_kid
+  , chain.chain_id as chain_id
+	, block.id as block_id
+  , block.created_at as block_created_at
+  , block.height as block_height
+  , tx.id as tx_id
+  , tx.data as tx
+ FROM tx
+ LEFT JOIN block ON tx.fk_block_id = block.id
+ LEFT JOIN chain ON block.fk_chain_id = chain.id
+ LEFT JOIN test_case ON chain.fk_test_id = test_case.id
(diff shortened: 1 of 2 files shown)


internal/blockdb/migrate.go
```diff
@@ -80,5 +80,27 @@ ON CONFLICT(git_sha) DO UPDATE SET git_sha=git_sha`, nowRFC3339(), gitSha)
 		return fmt.Errorf("create table tx: %w", err)
 	}
 
+	_, err = db.Exec(`CREATE VIEW IF NOT EXISTS tx_flattened AS
```
DavidNix, line 83: A few comments. To be clear, I love this idea!

## Idempotent

To keep the migrations idempotent, I recommend [dropping the view if exists](https://www.sqlitetutorial.net/sqlite-drop-view/) first. A view is essentialy a stored query so should be fast to drop/re-create. It will allow us to change the views more easily when we change table columns.

Also, we probably want to keep view creation as the very last migration steps (as you've done here). 

## Naming Convention

I like prefixing views with `v_` to signal it's a view a not a table. E.g. `v_tx_flattened`. 

## Delcare Columns Separately

This one is a nit for style.

It may read nicer to declare the columns separately as shown [here](https://www.sqlitetutorial.net/sqlite-create-view/). Look at the section `2) Creating a view with custom column names`. This is instead of using `SELECT col as another_name`. 

E.g. (per the article)

```
CREATE VIEW v_albums (
    AlbumTitle,
    Minutes
)
AS
    SELECT albums.title,
           SUM(milliseconds) / 60000
      FROM tracks
```

mark-rushakoff: > It may read nicer to declare the columns separately

That is likely true in cases where there are a small number of columns in the view, but IMO with ten columns in `v_tx_flattened`, it is more explicit to use `AS` to name each column; less chance of someone reordering the columns and creating a mismatch. Although, at the same time, we do have tests against this view.

@agouin would you care to be a tiebreaker on column separation in the built view? https://github.com/strangelove-ventures/ibctest/blob/29d64faf404be43f6e2512f6c690f38c2a6f8044/internal/blockdb/migrate.go#L88-L103

DavidNix (approved):

PR #134, opened 2022-06-09
feat: Capture chain_type in sqlite chain table

DavidNix:
Given ibctest supports many types of chains, we need to be explicit about the chain type (ecosystem). Any debugging tools can leverage this extra dimension to present more advanced debugging options for cosmos like the messages view from https://github.com/strangelove-ventures/ibctest/pull/133.

Supports: https://github.com/strangelove-ventures/ibctest/issues/123

### DIFF
--- a/internal/blockdb/migrate.go
+++ b/internal/blockdb/migrate.go
@@ -110,7 +110,8 @@ SELECT
   , test_case.name as test_case_name
   , chain.id as chain_kid
   , chain.chain_id as chain_id
-	, block.id as block_id
+  , chain.chain_type as chain_type
+  , block.id as block_id
   , block.created_at as block_created_at
   , block.height as block_height
   , tx.id as tx_id
--- a/internal/blockdb/migrate.go
+++ b/internal/blockdb/migrate.go
@@ -20,6 +20,8 @@ import (
 //  │                    │          │                    │         │                    │          │                    │
 //  └────────────────────┘          └────────────────────┘         └────────────────────┘          └────────────────────┘
 // The gitSha ensures we can trace back to the version of the codebase that produced the schema.
+// Warning: Typical best practice wraps each migration step into its own transaction. For simplicity given
+// this is an embedded database, we omit transactions.
 func Migrate(db *sql.DB, gitSha string) error {
 	// TODO(nix 05-27-2022): Appropriate indexes?
 	_, err := db.Exec(`PRAGMA foreign_keys = ON`)
--- a/internal/blockdb/migrate.go
+++ b/internal/blockdb/migrate.go
@@ -84,7 +84,7 @@ ON CONFLICT(git_sha) DO UPDATE SET git_sha=git_sha`, nowRFC3339(), gitSha)
 		return fmt.Errorf("create table tx: %w", err)
 	}
 
-	_, err = db.Exec(`ALTER TABLE chain ADD COLUMN chain_type TEXT NOT NULL check(length(chain_type) > 0) DEFAULT "cosmos"`)
+	_, err = db.Exec(`ALTER TABLE chain ADD COLUMN chain_type TEXT NOT NULL check(length(chain_type) > 0) DEFAULT "unknown"`)
 	if errIgnoreDuplicateColumn(err, "chain_type") != nil {
 		return fmt.Errorf("alter table chain add chain_type: %w", err)
 	}
--- a/internal/blockdb/migrate.go
+++ b/internal/blockdb/migrate.go
@@ -2,7 +2,11 @@ package blockdb
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
+	"strings"
+
+	"modernc.org/sqlite"
 )
 
 // Migrate migrates db in an idempotent manner.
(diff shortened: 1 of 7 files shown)


internal/blockdb/migrate.go
```diff
@@ -80,7 +84,21 @@ ON CONFLICT(git_sha) DO UPDATE SET git_sha=git_sha`, nowRFC3339(), gitSha)
 		return fmt.Errorf("create table tx: %w", err)
 	}
 
-	_, err = db.Exec(`DROP VIEW IF EXISTS v_tx_flattened`)
+	_, err = db.Exec(`ALTER TABLE chain ADD COLUMN chain_type TEXT NOT NULL check(length(chain_type) > 0) DEFAULT "cosmos"`)
+	if errIgnoreDuplicateColumn(err, "chain_type") != nil {
```
DavidNix, line 88: Sqlite does NOT have an `alter column`. If postgres, I would `alter column chain_type drop default` after adding the column.

I think the default of `cosmos` will be fine for ibctest but if anyone has issue, happy to hear alternatives. Otherwise, an idempotent migration WITH constraints like `not null` becomes tricky. The contraints take priority imo.

internal/blockdb/migrate.go
```diff
@@ -80,7 +84,21 @@ ON CONFLICT(git_sha) DO UPDATE SET git_sha=git_sha`, nowRFC3339(), gitSha)
...
+		return fmt.Errorf("alter table chain add chain_type: %w", err)
+	}
+
+	// Creating views should be last migration step.
+	// Error already wrapped.
+	return upsertViews(db)
```
DavidNix, line 94: `Migrate(...)` was getting long, so abstracted views into another function. May help ensure views are last in the migration process.

internal/blockdb/test_case.go
```diff
@@ -29,8 +29,9 @@ func CreateTestCase(ctx context.Context, db *sql.DB, testName, gitSha string) (*
 
 // AddChain tracks and attaches a chain to the test case.
 // The chainID must be globally unique. E.g. osmosis-1001, cosmos-1004
```
mark-rushakoff, line 31: I know this line was already here, but would this be more accurate as `must be unique within the test case`?

internal/blockdb/migrate.go
mark-rushakoff: I assume that at this point in time, we are safe to expect any users to delete and recreate their database. Therefore my loosely held preference is that we just  change the `CREATE TABLE` call, and if anyone gets a weird error today, the recommendation is to delete and start over.

Moreover, we don't yet have proper migration tests in place. That is, we don't have a previous version of a database, populated with data, that we migrate to the current version. 

We don't need to decide just yet, but I prefer a firm commitment one way or the other -- we are never migrating so you always have to start fresh when we change the schema; or we are guaranteeing (within reason) that **all** your old data will continue to be available as ibctest evolves. Right now, I don't think the collected data is valuable enough to justify the latter.

internal/blockdb/migrate.go
```diff
@@ -159,3 +180,12 @@ FROM v_tx_flattened, json_each(v_tx_flattened.tx, "$.body.messages")
 
 	return nil
 }
+
+func errIgnoreDuplicateColumn(err error, col string) error {
+	var serr *sqlite.Error
+	if errors.As(err, &serr) &&
```
mark-rushakoff, line 186: `&serr` doesn't need to be dereferenced here. It looks like everything is working as intended, but this is one corner of Go that I don't understand -- why is the address to a pointer accepted?

mark-rushakoff (approved):
I think we need a more defined commitment about how we will manage migrations, but that is outside the scope of this change.

internal/blockdb/migrate.go
DavidNix: Github isn't letting me comment on your previous comment. ¯\_(ツ)_/¯

> Moreover, we don't yet have proper migration tests in place. 

This is hard to pull off in practice. I've never worked at a place that did this well if at all. There's always bad and surprising data that you never accounted for. At previous companies, the strategy was if migrations failed, you did a quick rollback of the application. But that's in the classic web server context; it's different for embedded databases.

In this case, I think committing to the strategy of we delete (or backup) and start over is the way to go. 

I'll attempt a PR at ☝️ strategy first.

DavidNix: > &serr doesn't need to be dereferenced here.

It does need to be dereferenced, otherwise my IDE prints the warning: `Second argument to 'errors.As' must be a pointer to an interface or to a type implementing the error interface`. 

Only the pointer type to `sqlite.Error` implements the error interface (not the value).

internal/blockdb/migrate.go
DavidNix: Github isn't letting me comment on your previous comment. ¯_(ツ)_/¯

Moreover, we don't yet have proper migration tests in place.
This is hard to pull off in practice. I've never worked at a place that did this well if at all. There's always bad and surprising data that you never accounted for. At previous companies, the strategy was if migrations failed, you did a quick rollback of the application. But that's in the classic web server context; it's different for embedded databases.

In this case, I think committing to the strategy of we delete (or backup) and start over is the way to go.

I'll attempt a PR at ☝️ strategy first.

DavidNix:
I changed my opinion. I'm going to merge this one. 

> Therefore my loosely held preference is that we just change the CREATE TABLE call

The problem with this strategy is we `CREATE TABLE IF NOT EXISTS`. Therefore, if we add columns to a table on an existing schema, it will skip the create table call. And then downstream callers may expect new columns, but the columns won't be there resulting in potentially obscure errors.

But if we do `CREATE TABLE` without the existence check, the migrations will fail every time if the schema exists. I think that's terrible user experience - a database that holds only the first test case. The same problems applies to `ADD COLUMN` without ignoring the "duplicate column" error.

My hope was this database would not get too complex. If we feel the complexity grows more, it may be time to consider something like https://github.com/golang-migrate/migrate. With SQlite, my hope was to avoid it. (May need CGO for this migrate pkg.)

I think a simple, detailed error message to the user to rm or mv the old database is sufficient. I was going to attempt that but given we may move the database off the `Interchain` object, I will wait until that refactor is complete.

PR #136, opened 2022-06-10
Add FlushPackets and FlushAcknowledgements to ibc.Relayer

mark-rushakoff:
The previous ClearQueue method only flushed packets.

Both FlushPackets and FlushAcknowledgements are intended to be run as
one-off commands that do one chunk of work and stop. This is useful for
tests that need to make one unit of progress before making a set of
assertions.

In the future we may add a function like ClearQueue to the test package,
but for now the choices are StartRelayer, or manual flushing.

### DIFF
--- a/relayer/rly/cosmos_relayer.go
+++ b/relayer/rly/cosmos_relayer.go
@@ -65,7 +65,12 @@ const (
 // to distinguish between multiple rly versions.
 func Capabilities() map[relayer.Capability]bool {
 	m := relayer.FullCapabilities()
+
+	// These two capabilities are not supported in the 2.0.0-beta4 tag,
+	// but they do work on the current main branch of relayer.
+	// The first 2.0 release candidate should be able to have full capabilities.
 	m[relayer.TimestampTimeout] = false
+	m[relayer.FlushAcknowledgements] = false
 	return m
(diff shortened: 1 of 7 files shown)


relayer/rly/cosmos_relayer.go
```diff
@@ -65,7 +65,12 @@ const (
 // to distinguish between multiple rly versions.
 func Capabilities() map[relayer.Capability]bool {
 	m := relayer.FullCapabilities()
+
+	// These two capabilities are not supported in the 2.0.0-beta4 tag,
+	// but they do work on the current main branch of relayer.
+	// The first 2.0 release candidate should be able to have full capabilities.
 	m[relayer.TimestampTimeout] = false
+	m[relayer.FlushAcknowledgements] = false
```
DavidNix, line 73: Do we need a GH issue to remind us to adjust these lines when 2.0 released?

DavidNix (approved):
LGTM!

relayer/rly/cosmos_relayer.go
mark-rushakoff: It's front of mind for me, so I wasn't on adding an issue to track it.

PR #141, opened 2022-06-15
chore: Run code generation

DavidNix:

Easy to forget to run `go generate`. 

Code generation is run as part of creating the ibctest binary, `make ibctest`.

* May slightly slow down creating the binary.
* I've had CI steps that fail if code generation causes a dirty git state. But I don't think we need to go that far now.


mark-rushakoff (approved):
Should have been part of CI from the start -- thanks for adding it.

PR #143, opened 2022-06-15
Use timeouts for some docker operations

mark-rushakoff:
I continue to see regular hangs in Docker. I've added context timeouts
in a few places where commands should complete quickly, and I've updated
the docker wrapper for the relayer to respect context cancellation in
more docker API calls.

Now, instead of waiting for go test's default 10m timeout when a
container gets stuck in the Created state, you may instead see a context
deadline exceeded error like this:

    failed to restore key to relayer r for chain g0: context deadline exceeded


DavidNix (approved):
Good idea here. I would assume and hope non-cosmos chains can perform these operations fairly quickly as well.

PR #144, opened 2022-06-16
Default to rly v2.0.0-rc1

mark-rushakoff:
And update capabilities to match. Now, the full test suite from ibctest
should not skip any relayer tests due to missing capabilities.


DavidNix (approved):
Assuming we know all conformance tests pass (i.e. no bugs in relayer).

PR #145, opened 2022-06-16
feat: Add search feature to debug TUI

DavidNix:

Allows users to search text which gets highlighted in the associated text view. 

The last TUI feature I'll tackle before calling this MVP done is the ability to download the txs as a JSON array in case the user wants to use alternate tools to slice and dice the txs.

## Demo

tx-detail-search

* The UX may be a little confusing. It searches across all txs, highlighting all words so you can flip between txs without having to redo the search on every page.

### DIFF
--- a/internal/blockdb/tui/presenter/highlight.go
+++ b/internal/blockdb/tui/presenter/highlight.go
@@ -4,38 +4,39 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 )
 
 type Highlight struct {
-	regionText string
-	regions    []string
+	rx *regexp.Regexp
 }
 
 // NewHighlight returns a presenter that adds regions around text that matches searchTerm.
-func NewHighlight(fullText string, searchTerm string) *Highlight {
-	r, err := regexp.Compile(fmt.Sprintf(`(?is)(%s)`, regexp.QuoteMeta(searchTerm)))
-	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/blockdb/tui/presenter/highlight.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// NewHighlight returns a presenter that adds regions around text that matches searchTerm.
+func NewHighlight(searchTerm string) *Highlight {
+	searchTerm = strings.TrimSpace(searchTerm)
+	if searchTerm == "" {
+		return &Highlight{}
+	}
+	// Should always be valid with regexp.QuoteMeta.
+	return &Highlight{rx: regexp.MustCompile(fmt.Sprintf(`(?is)(%s)`, regexp.QuoteMeta(searchTerm)))}
```
mark-rushakoff, line 21: I suppose the `s` flag, which is intended to "let `.` match `\n`", will not do anything here because we will never have a `.` after the QuoteMeta.

mark-rushakoff (approved):

internal/blockdb/tui/presenter/highlight.go
DavidNix: I'll try it without that. My intention was to match multi-line.

DavidNix: Works fine without the flag. Nice catch.

PR #157, opened 2022-06-21
Add (*Interchain).Close method

mark-rushakoff:
This cleans up any created resources from Build(). The actual docker
containers are torn down via t.Cleanup, but the block collector
goroutine associated with the chainSet is now destroyed upon call to
Close, which reduces the number of goroutines present in a stack trace
due to a panic, test timeout, or deliberate SIGQUIT.

This PR is split into two commits for easier reading if so desired.

### DIFF
--- a/chainset.go
+++ b/chainset.go
@@ -2,6 +2,7 @@ package ibctest
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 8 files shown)


DavidNix (approved):
Long overdue, thanks!

One question is should we include an `os.Notify` in `cmd/ibctest_test.go`?

mark-rushakoff:
Good question on an os.Notify at the top level. I think it mainly makes sense if we associate it with a package level context so that the tests can use it. That can happen in a separate change if we go that route.

PR #160, opened 2022-06-21
feat: Copy TUI transactions to clipboard as JSON

DavidNix:

Finishes https://github.com/strangelove-ventures/ibctest/issues/123. There is some work in progess that allows us to link originating txs with the entire ibc lifecycle. I recommend creating a new issue; however, for that useful feature.

When viewing a list of txs, the new `c` command copies all txs to the user's clipboard. It *should* be cross-OS compatible. When on their clipboard, the user is in total control of where to save or process the text. I felt this was a simpler and more flexible approach than creating a JSON file on their filesystem.

This feature allows the user to use other tools to analyze the data, if they wish.

## Demo

txs-clipboard-demo

* To keep things simple, there is no UX feedback like a flash message or modal. For copying text to clipboard, the user should be able to figure out quickly if it worked.


DavidNix:
I did not swap out the clipboard pkg. The one I'm using seems actively maintained vs. other alternatives. 

@agouin I installed x11 on test-runner-2 via `sudo apt install libx11-dev`. 

The reason was to solve for the error:

```
# golang.design/x/clipboard
851
clipboard_linux.c:15:10: fatal error: X11/Xlib.h: No such file or directory
852
   15 | #include <X11/Xlib.h>
853
      |          ^~~~~~~~~~~~
854
compilation terminated.
855
FAIL	github.com/strangelove-ventures/ibctest/cmd/ibctest [build failed]
```

If we need to update something (like a launch template) to ensure X11 headers are always present on our test runner image, let me know.

cmd/ibctest/ibctest_test.go
```diff
@@ -258,6 +259,11 @@ func subcommand() string {
 }
 
 func runDebugTerminalUI(ctx context.Context) error {
+	// TUI has features that copy text to your clipboard.
+	if err := clipboard.Init(); err != nil {
```
mark-rushakoff, line 263: I don't like the hard dependency on CGo for clipboard support -- this tends to show up once you're shipping a binary to an Alpine system, using musl instead of glibc, and you need to start jumping through hoops to get the binary to run there.

mark-rushakoff (approved):
It would be nice to allow the TUI to run without clipboard support, for if and when we run into CGo build or cross-compile issues.

We have some options for this:

1. Continue if `clipboard.Init` fails, and just don't add the "copy to clipboard" keybinding
2. Add a wrapper around the clipboard library that can handle a `!cgo` build constraint, and either disable the clipboard feature or fall back to e.g. https://github.com/atotto/clipboard.

Whether we do one of those or something different, that can happen in a separate PR.

agouin (approved):
> I installed x11 on test-runner-2 via sudo apt install libx11-dev.

Sounds good. Maybe there's a way to workaround the error on linux systems that don't use X11? Not sure if the clipboard functionality would work with Wayland, for example.

I like @mark-rushakoff suggestion of `Continue if clipboard.Init fails, and just don't add the "copy to clipboard" keybinding`

cmd/ibctest/ibctest_test.go
DavidNix: I'm with you and great pushback. CGO 👎.  Like I originally planned, I'll swap out the clipboard library and handle when clipboard isn't available.

mark-rushakoff (approved):

mark-rushakoff (approved):

Issue #155, opened 2022-06-21
Create channel with specific version


DavidNix:
From cian, ibc-go team:

a channel needs to be created with a specific version for it to be enabled, I don't know is that configurable with ibctest? This command is using hermes but I think it should probably work in a similar way with go relayer

```
hermes -c ./network/hermes/config.toml create channel test-1 connection-0 \
--port-a transfer \
--port-b transfer \
-v "{\"fee_version\":\"ics29-1\",\"app_version\":\"ics20-1\"}"
```

in our tests it would be nice to test behaviour when one side does not have it enabled etc.
looking at the code, I'm guessing this would be done through CreateChannelOptions.Version

---

Note: This may have relayer dependencies.

jtieri:
The relayer commands for creating a new path should provide a `--version` flag so we *should* just need to add a field to the `CreateChannelOptions` struct like you mentioned.

jtieri:
Closed by #167

PR #167, opened 2022-06-22
Add support for declaring the channel configuration options

jtieri:
## Description

This PR addresses #155 by adding adding `CreateChannelOptions` as a field to the `InterchainBuildOptions` struct as well as adding it as a parameter to `LinkPath`.

If `CreateChannelOptions` are not specified on the `InterchainBuildOptions` struct then when the paths are linked in `interchain.Build` we default to the channel configuration options for an ics20 token transfer channel.

## Checklist

### DIFF
--- a/ibc/relayer.go
+++ b/ibc/relayer.go
@@ -2,6 +2,7 @@ package ibc
 
 import (
 	"context"
+	"fmt"
 	"time"
 )
 
@@ -90,9 +91,32 @@ type CreateChannelOptions struct {
 type Order int
 
 const (
-	Ordered Order = iota
+	Invalid Order = iota
+	Ordered
 	Unordered
 )
+
+// String returns the lowercase string representation of the Order.
+func (o Order) String() string {
+	switch o {
+	case Unordered:
+		return "unordered"
+	case Ordered:
+		return "ordered"
+	default:
+		return "invalid"
+	}
+}
+
+var ErrInvalidOrderType = fmt.Errorf("the specified channel order is invalid")
+
+// Validate checks that the Order type is a valid value.
+func (o Order) Validate() error {
+	if o == Ordered || o == Unordered {
+		return nil
+	}
+	return ErrInvalidOrderType
+}
 
 // DefaultChannelOpts returns the default settings for creating an ics20 fungible token transfer channel.
 func DefaultChannelOpts() CreateChannelOptions {
@@ -109,7 +133,7 @@ func (opts CreateChannelOptions) IsFullyConfigured() bool {
 	return opts.SourcePortName != "" &&
 		opts.DestPortName != "" &&
 		opts.Version != "" &&
-		opts.Order == Ordered || opts.Order == Unordered
+		opts.Order != Invalid
 }
 
 // ExecReporter is the interface of a narrow type returned by testreporter.RelayerExecReporter.
--- a/ibc/relayer.go
+++ b/ibc/relayer.go
@@ -81,19 +81,35 @@ type CreateChannelOptions struct {
 	SourcePortName string
 	DestPortName   string
 
-	Order string
+	Order Order
 
 	Version string
 }
 
-// DefaultChannelOpts returns the default settings for creating a ics20 fungible token transfer channel.
+// Order represents an IBC channel's order.
+type Order int
+
+const (
+	Ordered Order = iota
+	Unordered
+)
+
+// DefaultChannelOpts returns the default settings for creating an ics20 fungible token transfer channel.
 func DefaultChannelOpts() CreateChannelOptions {
 	return CreateChannelOptions{
 		SourcePortName: "transfer",
 		DestPortName:   "transfer",
-		Order:          "unordered",
+		Order:          Unordered,
 		Version:        "ics20-1",
 	}
+}
+
+// IsFullyConfigured returns true if all the CreateChannelOption's fields have been specified.
+func (opts CreateChannelOptions) IsFullyConfigured() bool {
+	return opts.SourcePortName != "" &&
+		opts.DestPortName != "" &&
+		opts.Version != "" &&
+		opts.Order == Ordered || opts.Order == Unordered
 }
 
 // ExecReporter is the interface of a narrow type returned by testreporter.RelayerExecReporter.
--- a/ibc/relayer.go
+++ b/ibc/relayer.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"fmt"
 	"time"
+
+	chanTypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+	ptypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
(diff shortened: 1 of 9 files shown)


interchain.go
```diff
@@ -226,7 +230,11 @@ func (ic *Interchain) Build(ctx context.Context, rep *testreporter.RelayerExecRe
 			)
 		}
 
-		if err := rp.Relayer.LinkPath(ctx, rep, rp.Path); err != nil {
+		// If channel creation options are not specified, default to the ics20 fungible token transfer channel options.
+		if (opts.CreateChannelOpts == ibc.CreateChannelOptions{}) {
+			opts.CreateChannelOpts = ibc.DefaultChannelOpts()
+		}
```
DavidNix, line 236: What happens if the user only sets partial info? 

E.g. `CreateChannelOptions{Version: "ics123"}` and they don't set other fields like `Order`.

DavidNix: What I'm hinting at is perhaps we want to default all fields to a reasonable value even if the user only sets some of the fields.

jtieri: Good point! I'll make those changes 👍

ibc/relayer.go
```diff
@@ -86,6 +86,16 @@ type CreateChannelOptions struct {
 	Version string
 }
 
+// DefaultChannelOpts returns the default settings for creating a ics20 fungible token transfer channel.
+func DefaultChannelOpts() CreateChannelOptions {
+	return CreateChannelOptions{
+		SourcePortName: "transfer",
+		DestPortName:   "transfer",
+		Order:          "unordered",
```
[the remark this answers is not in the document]
jtieri: I considered using an enum here but noticed there was another channel order type in the pipeline [[1]](https://github.com/cosmos/ibc/pull/636) and kind of second guessed it. I suppose realistically the number of values that could be added for this in the IBC spec shouldn't exceed more than a few so I'll make that change as well!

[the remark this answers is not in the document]
DavidNix: Cool. I'm glad I held off on suggesting a bool. In this case, `Order` should be a finite set of which enums typically are a good fit.

ibc/relayer_test.go
```diff
@@ -0,0 +1,34 @@
...
+	require.False(t, opts.IsFullyConfigured())
+
+	// Test invalid Order type in channel opts
+	opts = CreateChannelOptions{
+		SourcePortName: "transfer",
+		DestPortName:   "transfer",
+		Order:          3,
+		Version:        "123",
+	}
+	require.False(t, opts.IsFullyConfigured())
```
mark-rushakoff, line 25: Seems like this should be a different validation. The order is configured, but it has an invalid value.

jtieri: I've changed it up a bit locally, just trying to get tests passing and i'll have another commit here that addresses this.

ibc/relayer_test.go
```diff
@@ -0,0 +1,35 @@
+package ibc
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestChannelOptsConfigured(t *testing.T) {
```
DavidNix, line 9: 🔥

ibc/relayer_test.go
```diff
@@ -0,0 +1,35 @@
...
+	// Test the default channel opts
+	opts := DefaultChannelOpts()
+	require.True(t, opts.IsFullyConfigured())
```
DavidNix, line 12: nit:

`IsFullyConfigured` is a mouthful. I would opt for `IsValid`. Programmers are lazy. 😄

interchain.go
```diff
@@ -226,7 +230,17 @@ func (ic *Interchain) Build(ctx context.Context, rep *testreporter.RelayerExecRe
 			)
 		}
 
-		if err := rp.Relayer.LinkPath(ctx, rep, rp.Path); err != nil {
+		// If channel creation options are not fully specified, default to the ics20 fungible token transfer channel options.
+		if !opts.CreateChannelOpts.IsFullyConfigured() {
+			opts.CreateChannelOpts = ibc.DefaultChannelOpts()
```
DavidNix, line 235: Should we return an err here? Ideally telling the user what data they are missing?

ibc/relayer_test.go
DavidNix: Additionally, I often see `Validate() error` functions. That way, the error message can tell the user what they are missing.

jtieri: Yeah for sure, `Validate() error` is what I went with for determining the `Order` value is valid but went with `IsFullyConfigured()` to check that all fields were initialized because I noticed another method with the same name and functionality and wanted to keep naming consistent. https://github.com/strangelove-ventures/ibctest/blob/0a2f39ca8fdeac8e36ea2a3a8072aa550e49bc3e/ibc/types.go#L68-L78

I'm open to changing this to `Validate() error` though if that seems better!

interchain.go
jtieri: If we want to change `IsFullyConfigured() bool` to `Validate() error` then this would be ideal. 
Then we can either receive a fully configured `CreateChannelOpts` object with valid values or we return an error instead of trying to default to some acceptable behavior.

DavidNix: I'll approve for the first pass. But it's probably a better user experience to know what data they're missing.

DavidNix: To be clear, if a detailed error isn't in first pass that's fine - however, I think we need to return an error here or else the user may get unexpected behavior that's tricky to debug. If I expect a certain channel version and see it's defaulting to another, I'd be quite confused.

jtieri: I agree. I'm just going to include those changes real quick as well, thanks!

interchain.go
```diff
@@ -160,6 +160,10 @@ type InterchainBuildOptions struct {
 	// This is useful for tests that need lower-level access to configuring relayers.
 	SkipPathCreation bool
 
+	// If set, these options will be used when creating the channel in the path link step.
+	// If not set, the application will default to the options for creating an ics20 fungible token transfer.
```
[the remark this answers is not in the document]
jtieri: This is no longer true, I just missed this comment when I made changes last. 
Now we just check that either all of the `ChannelCreateOptions` are initialized and valid and if not we return an error on `(*Interchain).Build`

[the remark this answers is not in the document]
mark-rushakoff: Could we allow a zero ChannelCreateOptions? The struct only contains strings and ints, so a simple comparison would be valid:

```go
ccOpts := opts.ChannelCreateOpts
if ccOpts == (ChannelCreateOptions{}) {
  ccOpts = DefaultChannelOpts()
}
```

This makes the API a bit simpler to work with, as we don't have to explicitly pass `DefaultChannelOpts()` everywhere that we aren't concerned with special options.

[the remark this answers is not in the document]
jtieri: Yeah I think this is solid.

Then we are still returning errors if users provide a `CreateChannelOptions` where some of the fields are not initialized properly but we allow a zero value struct where we fall back to the default options.

PR #171, opened 2022-06-23
Add ibctest.TempDir function

### LINKED ISSUE #168: Consider using ioutil.TempDir instead of t.TempDir to create home directories
`t.TempDir` removes the home directory upon test passing or failing. This means that if a test fails, it is difficult to exec into the containers to debug as the home dir has been removed. [ioutil.TempDir](https://pkg.go.dev/io/ioutil#TempDir) leaves it to the caller to remove the directory. 

I propose something along the lines of

```golang
home, err := ioutil.TempDir("", t.Name())
// handle err

defer func(){
   if !t.Failed(){
     // remove home dir
  }
}
```

This way we can still exec into the containers and debug upon test failure.

mark-rushakoff:
This behaves similarly to `(*testing.T).TempDir()`, except that it
supports runtime configuration to determine whether to remove created
directories for failed tests.

By default, the created directories are removed as part of t.Cleanup. If
the environment variable `IBCTEST_KEEP_FAILURE_DIR` is set to any
non-empty value, and if the test failed, then the test will log that the
directory is not being cleaned up.

Importers of the ibctest package also have the option to set the package
level variable `KeepTempDirOnFailure` to opt in to this behavior.

Closes #168.

### DIFF
--- a/tempdir.go
+++ b/tempdir.go
@@ -25,9 +25,9 @@ type TempDirTestingT interface {
 // is retained or deleted following a test failure.
 //
 // It defaults to false, but can be initialized to true by setting the
-// environment variable IBCTEST_KEEP_FAILURE_DIR to a non-empty value.
+// environment variable IBCTEST_SKIP_FAILURE_CLEANUP to a non-empty value.
 // Alternatively, importers of the ibctest package may set the variable to true.
-var KeepTempDirOnFailure = os.Getenv("IBCTEST_KEEP_FAILURE_DIR") != ""
+var KeepTempDirOnFailure = os.Getenv("IBCTEST_SKIP_FAILURE_CLEANUP") != ""
 
 // TempDir resembles (*testing.T).TempDir, except that it conditionally
 // keeps the temporary directory on disk, and it uses a new temporary directory
--- a/tempdir.go
+++ b/tempdir.go
@@ -1,0 +1,91 @@
+package ibctest
+
+import (
+	"fmt"
+	"os"
+	"strings"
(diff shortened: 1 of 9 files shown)


tempdir.go
```diff
@@ -0,0 +1,91 @@
...
+	Errorf(format string, args ...interface{})
+}
+
+// KeepTempDirOnFailure determines whether a directory created by TempDir
+// is retained or deleted following a test failure.
+//
+// It defaults to false, but can be initialized to true by setting the
+// environment variable IBCTEST_KEEP_FAILURE_DIR to a non-empty value.
+// Alternatively, importers of the ibctest package may set the variable to true.
+var KeepTempDirOnFailure = os.Getenv("IBCTEST_KEEP_FAILURE_DIR") != ""
```
mark-rushakoff, line 30: I'm open to suggestions on a better name for this environment variable.

The `IBCTEST_` prefix would be standard. `KEEP_FAILURE_DIR` reads a little like "the directory in which failures should be kept", but matching the whole name `IBCTEST_KEEP_TEMP_DIR_ON_FAILURE` is a bit too long IMO.

DavidNix: `KEEP_DATA_ON_FAILURE`, `KEEP_DIR_ON_FAILURE`?

I also can't think of a shorter name that's as clear. 

---

Another note: Should we add this env var to the README or assume godoc is good enough?

DavidNix (approved):

jtieri (approved):

PR #173, opened 2022-06-24
feat: Exec arbitrary container commands for chain

DavidNix:

90% finishes: https://github.com/strangelove-ventures/ibctest/issues/158. The remaining 10% is exposing the HomeDir and container hostnames which are often required for commands.

A job container is a short-lived container that runs a command and exits. I still think it's possible to exec into an already running container vs. running one-off containers. The only edge case is needing to run a chain command BEFORE you start the chain. Locally and on CI creating/destroying these job containers is quite performant.

This is an attempt to DRY up some of the docker code and tame it by abstracting it into our own package `dockerutil`. The docker config and order of operations is delicate. 

Also, fixes the problem of a non-zero status code omitting useful error info.

* Chain needs to expose more docker information like host and home dir. See [this example](https://github.com/cosmos/ibc-go/pull/1553/files#diff-9a2a689096b5a2d214655a4fa9790246fd58d90d2a442c6b20c7389cbe026d3d).
* Bloats the ibc.Chain interface a little more.
* Provides abstraction but does not use that abstraction for long lived containers yet.
* Cleaning up via DockerSetup is still more brittle than I'd like because of the test name as the label value. You have to ensure you are using that name consistently and it's not super obvious. Also the test name could change. 
* Relayer probably needs exec as well.


chain/cosmos/cosmos_chain.go
```diff
@@ -293,153 +299,6 @@ type ValidatorWithIntPower struct {
 	PubKeyBase64 string
 }
 
-// Bootstraps the chain and starts it from genesis
-func (c *CosmosChain) StartWithGenesisFile(testName string, ctx context.Context, home string, pool *dockertest.Pool, networkID string, genesisFilePath string) error {
```
DavidNix, line 297: This was never used.

ibc/chain.go
```diff
@@ -16,8 +16,9 @@ type Chain interface {
 	// sets up everything needed (validators, gentx, fullnodes, peering, additional accounts) for chain to start from genesis
 	Start(testName string, ctx context.Context, additionalGenesisWallets ...WalletAmount) error
 
-	// start a chain with a provided genesis file. Will override validators for first 2/3 of voting power
-	StartWithGenesisFile(testName string, ctx context.Context, home string, pool *dockertest.Pool, networkID string, genesisFilePath string) error
+	// Exec runs an arbitrary command using Chain's docker environment.
+	// "env" are environment variables in the format "MY_ENV_VAR=value"
+	Exec(ctx context.Context, cmd []string, env []string) (stdout, stderr []byte, err error)
```
DavidNix, line 21: FWIW, I think we need to break this interface up into separate categories of chains.

Such as:
```
BaseChain // probably includes IBCTransfer
InterchainAccountChain
QueryChain // IBC queries
SmartContractChain
ExportableChain // export and start from state
```

chain/cosmos/chain_node.go
```diff
@@ -678,11 +682,6 @@ func (tn *ChainNode) CreateNodeContainer(ctx context.Context) error {
 	return nil
 }
 
-func (tn *ChainNode) StopContainer(ctx context.Context) error {
```
DavidNix, line 681: Never called.

internal/dockerutil/image.go
```diff
@@ -0,0 +1,272 @@
...
+	"strings"
+	"time"
+
+	"github.com/ory/dockertest/v3"
+	"github.com/ory/dockertest/v3/docker"
+	"go.uber.org/zap"
+)
+
+// Image is a docker image.
+type Image struct {
```
DavidNix, line 17: The type and `Container` are the meat of this PR.

internal/dockerutil/image.go
```diff
@@ -0,0 +1,275 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	var (
+		client     = c.image.pool.Client
+		notFound   *docker.NoSuchContainer
+		notRunning *docker.ContainerNotRunning
+	)
+
+	err := client.StopContainerWithContext(c.container.ID, 0, ctx)
```
mark-rushakoff, line 255: > StopContainerWithContext stops a container, killing it after the given timeout (in seconds). The context can be used to cancel the stop container request.


This sounds like we would still want a non-zero timeout.

mark-rushakoff (approved):
I've left a handful of comments, but nothing blocking merge -- any of the more involved details can be a separate change later.

This updated structure should make it easier to isolate the cause of the containers being stuck in created state, if it does continue to happen.

ibc/chain.go
```diff
@@ -16,8 +16,9 @@ type Chain interface {
 	// sets up everything needed (validators, gentx, fullnodes, peering, additional accounts) for chain to start from genesis
 	Start(testName string, ctx context.Context, additionalGenesisWallets ...WalletAmount) error
 
-	// start a chain with a provided genesis file. Will override validators for first 2/3 of voting power
-	StartWithGenesisFile(testName string, ctx context.Context, home string, pool *dockertest.Pool, networkID string, genesisFilePath string) error
+	// Exec runs an arbitrary command using Chain's docker environment.
+	// "env" are environment variables in the format "MY_ENV_VAR=value"
```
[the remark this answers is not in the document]
DavidNix: It could be either. Right now, it's always a one-off command. I'll clarify.

internal/dockerutil/setup.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		nets, _ := pool.Client.ListNetworks()
+		for _, n := range nets {
+			for k, v := range n.Labels {
+				if k == CleanupLabel && v == testName {
+					_ = pool.Client.RemoveNetwork(n.ID)
+					break
+				}
+			}
+		}
```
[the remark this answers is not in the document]
DavidNix: Yup, this was copy/pasted from the original implementation in the ibctest root level package. All that nesting didn't sit well with me either.

internal/dockerutil/setup.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+	if err != nil {
+		t.Fatalf("failed to create docker network: %v", err)
+	}
+
+	return pool, network.Network.ID
+}
+
+// dockerCleanup will clean up Docker containers, networks, and the other various config files generated in testing
+func dockerCleanup(testName string, pool *dockertest.Pool) func() {
```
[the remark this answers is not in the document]
DavidNix: I agree. 

What also doesn't sit well is the fact the `testName` can easily change if the developer decides to refactor the name of a test. E.g. `TestChainFoo(t *testing.T)` -> `TestChainBar(t *testing.T)`. 

You also have to be sure to pass the testName correctly down the call stack. 

The cleanup strategy is delicate and may be missing some key parts. It's largely copy/pasted from the original ibctest location.

chain/cosmos/cosmos_chain.go
agouin: This was used during the juno halt to test the new genesis file. That has been removed and a link to it is in the trophies in the README. Something like this could be used for the test caching, e.g. export state after creating clients, connections, and channels, to have a quick restore point, then a function similar to this could do a quick startup. Would also be nice to have something like this available for testing existing genesis files in one off test cases.

agouin (commented):
Awesome cleanup 👏

chain/cosmos/chain_node.go
agouin: Was also used for the juno halt recovery to test taking down nodes, writing a new genesis file, then bringing them back up. Will be nice for one off tests to still have the ability to do this.

PR #176, opened 2022-06-27
Adding support for `RecoverKey` in the chain interface.

### LINKED ISSUE #169: Enable a RestoreKey function for the chains
Since the relayer address has been exposed, I ran into a new issue where I needed to run an sdk command and sign a transaction using the relayer address. The relayer address is not in the keychain (as the relayer address/mnemonic is generated before the chain nodes exist)

I propose a function that looks something like this
```golang
func (c *CosmosChain) RecoverKey(ctx context.Context, name, mnemonic string) error {
	tn := c.ChainNodes[0]

	cmd := []string{
		"bash",
		"-c",
		fmt.Sprintf(`echo "%s" | %s keys add %s --recover --keyring-backend %s --home %s`, mnemonic, c.Config().Bin, name, keyring.BackendTest, tn.NodeHome()),
	}

	exitCode, stdout, stderr, err := tn.NodeJob(ctx, cmd)
	if err != nil {
		return dockerutil.HandleNodeJobError(exitCode, stdout, stderr, err)
	}

	return nil
}
```

I don't know if there is a more elegant way of piping values using `NodeJob` or maybe there is a command that doesn't require it.

chatton:
## Description

This PR adds the functionality to recover a test user from a given mnemonic.

## Checklist

- [X] I have made sure the upstream branch for this PR is correct
- [X] I have made sure this PR is ready to merge

## Changes

- [x] Added `RecoverKey` to `ChainNode` and `CosmosChain` types.
- [x] Added `RecoverKey` to `Chain` interface.
- [ ] ~Added `Shell` field to `ChainConfig`. This is to allow you to specify which shell commands which run using `-c <script>` will use, as not all images will have the same shells.~
- [x] Added `GetAndFundTestUserWithMnemonic` which generates a user using a given mnemonic.
- [x] Changed `GetAndFundTestUsers` to delegate to `GetAndFundTestUserWithMnemonic`
- [x] Changed `generateUserWallet` to either `Create` or `Recover` based on the presence of a mnemonic.

## Testing

Tests have been added which verify that users can be created using `CreateKey` and `RecoverKey`

closes #169

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -321,6 +321,18 @@ func (tn *ChainNode) CreateKey(ctx context.Context, name string) error {
 	return dockerutil.HandleNodeJobError(tn.NodeJob(ctx, command))
 }
 
+// RecoverKey restores a key from a given mnemonic.
+func (tn *ChainNode) RecoverKey(ctx context.Context, keyName, mnemonic string) error {
+	command := []string{
+		tn.Chain.Config().GetShell(),
+		"-c",
+		fmt.Sprintf(`echo "%s" | %s keys add %s --recover --keyring-backend %s --home %s --output json`, mnemonic, tn.Chain.Config().Bin, keyName, keyring.BackendTest, tn.NodeHome()),
+	}
+	tn.lock.Lock()
+	defer tn.lock.Unlock()
+	return dockerutil.HandleNodeJobError(tn.NodeJob(ctx, command))
+}
+
 // AddGenesisAccount adds a genesis account for each key
 func (tn *ChainNode) AddGenesisAccount(ctx context.Context, address string, genesisAmount []types.Coin) error {
 	amount := ""
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -324,9 +324,9 @@ func (tn *ChainNode) CreateKey(ctx context.Context, name string) error {
 // RecoverKey restores a key from a given mnemonic.
 func (tn *ChainNode) RecoverKey(ctx context.Context, keyName, mnemonic string) error {
 	command := []string{
-		tn.Chain.Config().GetShell(),
+		"sh",
 		"-c",
-		fmt.Sprintf(`echo "%s" | %s keys add %s --recover --keyring-backend %s --home %s --output json`, mnemonic, tn.Chain.Config().Bin, keyName, keyring.BackendTest, tn.NodeHome()),
+		fmt.Sprintf(`echo %q | %s keys add %s --recover --keyring-backend %s --home %s --output json`, mnemonic, tn.Chain.Config().Bin, keyName, keyring.BackendTest, tn.NodeHome()),
 	}
 	tn.lock.Lock()
 	defer tn.lock.Unlock()
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -330,7 +330,8 @@ func (tn *ChainNode) RecoverKey(ctx context.Context, keyName, mnemonic string) e
 	}
 	tn.lock.Lock()
 	defer tn.lock.Unlock()
-	return dockerutil.HandleNodeJobError(tn.NodeJob(ctx, command))
+	_, _, err := tn.Exec(ctx, command, nil)
+	return err
 }
 
 // AddGenesisAccount adds a genesis account for each key
--- a/chain/penumbra/penumbra_chain.go
+++ b/chain/penumbra/penumbra_chain.go
@@ -140,6 +140,10 @@ func (c *PenumbraChain) CreateKey(ctx context.Context, keyName string) error {
 	return c.getRelayerNode().PenumbraAppNode.CreateKey(ctx, keyName)
 }
 
+func (c *PenumbraChain) RecoverKey(ctx context.Context, name, mnemonic string) error {
+	return fmt.Errorf("RecoverKey not implemented for PenumbraChain")
+}
+
 // Implements Chain interface
 func (c *PenumbraChain) GetAddress(ctx context.Context, keyName string) ([]byte, error) {
 	return c.getRelayerNode().PenumbraAppNode.GetAddress(ctx, keyName)
--- a/ibc/types.go
+++ b/ibc/types.go
@@ -14,6 +14,7 @@ type ChainConfig struct {
 	GasAdjustment  float64
 	TrustingPeriod string
 	NoHostMount    bool
+	Shell          string
 }
 
 func (c ChainConfig) MergeChainSpecConfig(other ChainConfig) ChainConfig {
@@ -60,6 +61,14 @@ func (c ChainConfig) MergeChainSpecConfig(other ChainConfig) ChainConfig {
 	// Skip NoHostMount so that false can be distinguished.
 
 	return c
+}
+
+// GetShell returns the shell that should be used when executing scripts.
+func (c ChainConfig) GetShell() string {
+	if c.Shell == "" {
+		return "sh"
+	}
+	return c.Shell
 }
 
 // IsFullyConfigured reports whether all required fields have been set on c.
--- a/ibc/types.go
+++ b/ibc/types.go
@@ -14,7 +14,6 @@ type ChainConfig struct {
 	GasAdjustment  float64
 	TrustingPeriod string
 	NoHostMount    bool
-	Shell          string
 }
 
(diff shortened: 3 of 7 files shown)


chain/penumbra/penumbra_chain.go
```diff
@@ -140,6 +140,10 @@ func (c *PenumbraChain) CreateKey(ctx context.Context, keyName string) error {
 	return c.getRelayerNode().PenumbraAppNode.CreateKey(ctx, keyName)
 }
 
+func (c *PenumbraChain) RecoverKey(ctx context.Context, name, mnemonic string) error {
+	return fmt.Errorf("RecoverKey not implemented for PenumbraChain")
+}
```
chatton, line 145: I'm not sure exactly what would need to be done here.

mark-rushakoff: I think this error is fine for now.

test_user.go
```diff
@@ -37,6 +44,31 @@ func generateUserWallet(ctx context.Context, keyName string, chain ibc.Chain) (*
...
+	user, err := generateUserWallet(ctx, keyName, mnemonic, chain)
+	require.NoError(t, err, "failed to get source user wallet")
+
+	err = chain.SendFunds(ctx, FaucetAccountKeyName, ibc.WalletAmount{
+		Address: user.Bech32Address(chainCfg.Bech32Prefix),
+		Amount:  amount,
+		Denom:   chainCfg.Denom,
+	})
+	require.NoError(t, err, "failed to get funds from faucet")
+	require.NoError(t, test.WaitForBlocks(ctx, 5, chain), "failed to wait for blocks")
```
mark-rushakoff, line 67: Did 5 come from anywhere special? Would 4 or 3 work?

Some of these tests can take a long time to complete, so if there is a way to minimize time spent waiting, we should take it.

Maybe it would make more sense to not wait inside this function, and indicate in Godoc that the caller should wait for some blocks to complete before the funds will be accessible. There is one wait at the end of `GetAndFundTestUsers` that would be redundant with this call, if it stays.

interchain_test.go
```diff
@@ -145,6 +148,83 @@ func TestInterchain_GetRelayerWallets(t *testing.T) {
...
+
+	t.Run("Test user creation without mnemonic", func(t *testing.T) {
+		keyName := "regular-user-name"
+		users := ibctest.GetAndFundTestUsers(t, ctx, keyName, 10000, gaia0)
+		require.Len(t, users, 1)
+		require.NotEmpty(t, users[0].Address)
+		require.NotEmpty(t, users[0].KeyName)
+	})
+
+	_ = ic.Close()
```
mark-rushakoff, line 225: Preferable to `defer ic.Close()` immediately after the initial assignment. 

If the linter complains about a discarded error value, `defer func() { _ = ic.Close() }` is fine.

ibc/types.go
```diff
@@ -62,6 +63,14 @@ func (c ChainConfig) MergeChainSpecConfig(other ChainConfig) ChainConfig {
...
 }
 
+// GetShell returns the shell that should be used when executing scripts.
+func (c ChainConfig) GetShell() string {
+	if c.Shell == "" {
+		return "sh"
+	}
+	return c.Shell
+}
+
```
mark-rushakoff, line 73: For now, I think we should assume all containers have `sh` available. It will likely be a long time before we run into something we can't do with plain `sh`.

interchain_test.go
```diff
@@ -145,6 +148,83 @@ func TestInterchain_GetRelayerWallets(t *testing.T) {
...
+
+	ic := ibctest.NewInterchain().
+		AddChain(gaia0).
+		AddChain(gaia1).
+		AddRelayer(r, "r").
+		AddLink(ibctest.InterchainLink{
+			Chain1:  gaia0,
+			Chain2:  gaia1,
+			Relayer: r,
+		})
```
mark-rushakoff, line 184: Because this test is only exercising funding a user, I think this can be reduced to `ic := ibctest.NewInterchain().AddChain(gaia0)`. 

This should be testable even without an `Interchain`, but since that does the heavy lifting to start a chain, it's probably more convenient than calling several methods directly on a chain.

interchain_test.go
```diff
@@ -145,6 +148,83 @@ func TestInterchain_GetRelayerWallets(t *testing.T) {
...
+
+	ctx := context.Background()
+	require.NoError(t, ic.Build(ctx, eRep, ibctest.InterchainBuildOptions{
+		TestName:  t.Name(),
+		HomeDir:   home,
+		Pool:      pool,
+		NetworkID: network,
+	}))
+
+	t.Run("Test User creation with mnemonic", func(t *testing.T) {
```
mark-rushakoff, line 197: Names of subtests should be shorter because they are displayed after the outer test's name. This line can be `t.Run("with mnemonic", ...)`.

interchain_test.go
```diff
@@ -145,6 +148,83 @@ func TestInterchain_GetRelayerWallets(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NotEmpty(t, mnemonic)
+
+		user := ibctest.GetAndFundTestUserWithMnemonic(t, ctx, keyName, mnemonic, 10000, gaia0)
+
+		require.NotEmpty(t, user.Address)
+		require.NotEmpty(t, user.KeyName)
+	})
+
+	t.Run("Test user creation without mnemonic", func(t *testing.T) {
```
mark-rushakoff, line 217: `t.Run("without mnemonic", ...)`

interchain_test.go
```diff
@@ -145,6 +148,83 @@ func TestInterchain_GetRelayerWallets(t *testing.T) {
...
+			hd.Secp256k1,
+		)
+
+		require.NoError(t, err)
+		require.NotEmpty(t, mnemonic)
+
+		user := ibctest.GetAndFundTestUserWithMnemonic(t, ctx, keyName, mnemonic, 10000, gaia0)
+
+		require.NotEmpty(t, user.Address)
+		require.NotEmpty(t, user.KeyName)
```
mark-rushakoff, line 214: I would like to see a call to `gaia0.GetBalance` to assert the funds were sent, in both tests here.

chain/cosmos/chain_node.go
```diff
@@ -321,6 +321,18 @@ func (tn *ChainNode) CreateKey(ctx context.Context, name string) error {
 	return dockerutil.HandleNodeJobError(tn.NodeJob(ctx, command))
 }
 
+// RecoverKey restores a key from a given mnemonic.
+func (tn *ChainNode) RecoverKey(ctx context.Context, keyName, mnemonic string) error {
+	command := []string{
+		tn.Chain.Config().GetShell(),
```
mark-rushakoff, line 327: This can be hardcoded to `"sh"` after removing the `GetShell` method.

chain/cosmos/chain_node.go
```diff
@@ -321,6 +321,18 @@ func (tn *ChainNode) CreateKey(ctx context.Context, name string) error {
...
+		"-c",
+		fmt.Sprintf(`echo "%s" | %s keys add %s --recover --keyring-backend %s --home %s --output json`, mnemonic, tn.Chain.Config().Bin, keyName, keyring.BackendTest, tn.NodeHome()),
```
mark-rushakoff, line 329: `echo %q | ...` should be safer, as that will surround the string with quotes and add some escaping of special characters.

mark-rushakoff (commented):
There are a few things to clean up, but I think we'll be able to get this merged soon.

DavidNix (dismissed):

DavidNix (commented):
Apologies, I did not mean to approve previously, just comment.

test_user.go
chatton: This wait was just taken from the existing `GetAndFundTestUsers` function, however I think it's no harm removing it and letting the caller know as you suggested. I can remove from both.

chain/cosmos/chain_node.go
chatton: Thanks for this, I wasn't familiar with %q here!

chatton:
@mark-rushakoff ready for another look, I've addressed your feedback, thanks for taking a look so quickly! 

Tests seem to be failing but it looks like an environmental issue on the test host?

agouin:
Hi @chatton, looking at the test failure logs:

> tn.NodeJob undefined (type *ChainNode has no field or method NodeJob)

Earlier today, `NodeJob` was replaced by `Exec` in #173, so those calls will need to be updated in `chain_node.go`.

chatton:
@agouin thanks for letting me know! I had missed that PR being merged.

PR #180, opened 2022-06-27
fix: Minor docker cleanup improvements

DavidNix:
Feedback from https://github.com/strangelove-ventures/ibctest/pull/173

### DIFF
--- a/internal/dockerutil/setup.go
+++ b/internal/dockerutil/setup.go
@@ -2,6 +2,7 @@ package dockerutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
@@ -47,8 +48,6 @@ func DockerSetup(t *testing.T) (*dockertest.Pool, string) {
 		t.Fatalf("failed to create docker network: %v", err)
 	}
 
-	t.Logf("Created docker network %s", network.Network.ID)
-
 	return pool, network.Network.ID
 }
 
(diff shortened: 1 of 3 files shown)


internal/dockerutil/setup.go
```diff
@@ -51,29 +52,57 @@ func DockerSetup(t *testing.T) (*dockertest.Pool, string) {
...
 					ctxWait, cancelWait := context.WithTimeout(context.Background(), time.Second*5)
-					_, _ = pool.Client.WaitContainerWithContext(c.ID, ctxWait)
-					_ = pool.Client.RemoveContainer(docker.RemoveContainerOptions{ID: c.ID, Force: true})
+					if _, err := pool.Client.WaitContainerWithContext(c.ID, ctxWait); err != nil {
+						t.Logf("Failed to wait for container %s during docker cleanup: %v", c.ID, err)
+					}
+					if err := pool.Client.RemoveContainer(docker.RemoveContainerOptions{
+						ID:            c.ID,
+						Force:         true,
+						RemoveVolumes: true}); err != nil {
```
DavidNix, line 71: (*dockertest.Pool).Prune sets the RemoveVolumes to true. Previously, throughout the codebase, we were omitting this field. I opted to mimic what the package uses. 

It's subtle which is why I wanted to point this out. 🤞 maybe it helps with docker hanging issues.

mark-rushakoff: I couldn't find a quick answer on what remove volumes does, if it is a volume in use by other containers. I assume it silently leaves the volume in place. Removing the volume makes sense to me.

internal/dockerutil/setup.go
```diff
@@ -51,29 +52,57 @@ func DockerSetup(t *testing.T) (*dockertest.Pool, string) {
 }
 
 // dockerCleanup will clean up Docker containers, networks, and the other various config files generated in testing
-func dockerCleanup(testName string, pool *dockertest.Pool) func() {
+func dockerCleanup(t *testing.T, pool *dockertest.Pool) func() {
 	return func() {
 		cont, _ := pool.Client.ListContainers(docker.ListContainersOptions{All: true})
 		for _, c := range cont {
 			for k, v := range c.Labels {
```
mark-rushakoff, line 59: `c.Labels` should be a `map[string]string`, so I think we can directly check `v, ok := c.Labels[CleanupLabel]`.

mark-rushakoff (approved):

PR #182, opened 2022-06-27
feat: Add GetHomeDir method to ibc.Chain

DavidNix:

Closes https://github.com/strangelove-ventures/ibctest/issues/158

Many cosmos commands require this argument.

* Bloats the ibc.Chain interface more. However, I like interface methods that expose data vs. provide some functionality. By simply exposing data, we can typically write reusable functions that leverage the data. 
* `GetHomeDir` may be too cosmos specific.
* There may be a convention that's always true such as `/tmp/{{chain binary name}}`. But it's unclear how to enforce and make obvious such a convention.


jtieri (approved):
I think the notion of a home directory in the filesystem is generic enough to work fine across Chain implementations.

 🚀

agouin (approved):
Looks good to me. Shouldn't it be named `HomeDir()` instead of `GetHomeDir()`?

DavidNix:
Fyi, it should be `HomeDir()` per Go naming conventions. However, we have `GetGRPCAddress()` and `GetRPCAddress()` so I was being consistent with the existing `ibc.Chain` interface by choosing the `Get` prefix.

PR #186, opened 2022-06-30
Add support for broadcasting arbitrary messages as test users

### LINKED ISSUE #185: Feature Request: Support broadcasting messages as test users from the test host
From the tests, I would like to be able to broadcast arbitrary messages as users created in the tests themselves.

In issue #172 I found myself needing the private key of the users in order to sign transactions, however since then I found out that it should be possible to broadcast messages using [this function](https://github.com/cosmos/cosmos-sdk/blob/c6315ec1b102f77a2f3e766093175a7535b3a52c/client/tx/tx.go#L65).

If we use this function, I would no longer need private keys exposed (although I suspect it would still be useful in certain situations)

When attempting to utilize this function, I ran into a few issues.

* `BroadCastTx` signs the transactions by reading from the specified keyring, which means the test host needs access to valid keyring for this to work.
* A `tx.Factory` needs to be configured based on the chains.
* A `client.Context` needs to be configured based on user/chain/test environment.

chatton:
## Description

This PR adds support for the ability to Broadcast transactions on CosmosChains

closes #185

## Checklist

- [X] I have made sure the upstream branch for this PR is correct
- [X] I have made sure this PR is ready to merge

## Changes

- [X] Added new interface `Broadcaster`
- [X] Add Cosmos chain implementation.
- [X] Added `NewLocalKeyringFromDockerContainer` a mechanism to use the keychain of created test users.
- [X] Added test to ensure a broadcast with a created user succeeds (this also verifies `NewLocalKeyringFromDockerContainer` works as this would fail if the keyring was not there. )

## Notes

- I tried to design the interfaces in such a way so that the backing congiuration (tx.Factory and client.Context) are fully configurable by the caller so that we can change any arbitrary options we need.


chain/cosmos/cosmos_broadcaster.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := defaultEncoding.Marshaler.UnmarshalJSON(bytes, &resp); err != nil {
+		return sdk.TxResponse{}, err
+	}
+	return resp, nil
+}
+
+// defaultClientContext returns a default client context configured with the user as the sender.
+func (b *Broadcaster) defaultClientContext(fromUser broadcast.User, sdkAdd sdk.AccAddress) client.Context {
+	// initialize a clean buffer each time
+	b.buf = &bytes.Buffer{}
```
mark-rushakoff, line 138: This should probably be `b.buf.Reset()`.

internal/dockerutil/keyring.go
```diff
@@ -0,0 +1,63 @@
...
+		if isDirectory {
+			continue
+		}
+
+		filePath := fmt.Sprintf("%s/keyring-test/%s", localDirectory, extractedFileName)
+		if err := os.WriteFile(filePath, fileBuff.Bytes(), os.ModePerm); err != nil {
+			return nil, err
+		}
+	}
+	return keyring.New("", keyring.BackendTest, localDirectory, os.Stdin)
```
mark-rushakoff, line 62: Could this be an in-memory keyring? AFAICS this doesn't need to cross process boundaries.

internal/dockerutil/keyring.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		name := hdr.Name
+		splitName := strings.Split(name, "/")
+
+		extractedFileName := splitName[len(splitName)-1]
+		isDirectory := extractedFileName == ""
+		if isDirectory {
+			continue
+		}
```
mark-rushakoff, line 55: This could use `isDirectory := strings.HasSuffix(name, "/")` and `extractedFileName := path.Base(name)`.

mark-rushakoff (commented):
A couple high-level notes:

- Move `Broadcast` interface into root `ibctest` package
- `cosmos_broadcaster.go` can be named just `broadcaster.go` since it is already in the `cosmos` directory

I think the Broadcast interface is fine to focus on Cosmos, but others on the team have given Cosmos vs non-Cosmos more thought than I have, so it would be good to hear from someone else.

broadcast/broadcast.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type ClientContextOpt func(clientContext client.Context) client.Context
+
+type FactoryOpt func(factory tx.Factory) tx.Factory
+
+type User interface {
```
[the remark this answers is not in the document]
chatton: if we are in the `ibctest` package I believe we may just be able to accept an `*ibctest.User` directly.

chatton:
@mark-rushakoff moving the code into the `ibctest` package results in a circular import between `ibctest` and `cosmos` with a call to `cosmos.NewCosmosChain` in `chainfactory.go`. What would be your preferred way of resolving this? 

Some options are:

* Create an additional package just to hold types
* Remove broadcaster interface and just use the concrete `cosmos.Broadcaster` and `cosmos.BroadCastUser`
* Maintain separate package for broadcast interfaces and types.

internal/dockerutil/keyring.go
chatton: I created [this issue](https://github.com/cosmos/cosmos-sdk/issues/12438)

chatton:
I ended up keeping the interfaces in a `broadcast` package, but moved the main function to `ibctest` as `BroadcastTx`

internal/dockerutil/keyring.go
mark-rushakoff: I would have expected to be able to iterate through the files and add them to the in-memory keychain, or maybe to read it locally as a test backend and do a bulk import into an in-memory keychain. Looks like neither of those approaches are straightforward, so this can stay as-is for now.

chain/cosmos/broadcaster.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// ConfigureClientContextOptions ensure the given configuration functions are run when calling GetClientContext
+// after all default options have been applied.
+func (b *Broadcaster) ConfigureClientContextOptions(opts ...broadcast.ClientContextOpt) {
+	b.clientContextOptions = append(b.clientContextOptions, opts...)
+}
+
+// GetFactory returns an instance of tx.Factory that is configured with this Broadcaster's CosmosChain
+// and the provided user.
```
mark-rushakoff, line 64: Wouldn't hurt to mention `ConfigureFactoryOptions` in the godoc for this method.

chain/cosmos/broadcaster.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, opt := range b.clientContextOptions {
+		clientContext = opt(clientContext)
+	}
+	return clientContext, nil
+}
+
+// GetTxResponseBytes returns the sdk.TxResponse bytes which returned from broadcast.Tx.
+func (b *Broadcaster) GetTxResponseBytes(ctx context.Context, user broadcast.User) ([]byte, error) {
+	if b.buf == nil || b.buf.Len() == 0 {
+		return nil, fmt.Errorf("empty buffer, transaction has not be executed yet")
```
mark-rushakoff, line 120: s/be/been/

mark-rushakoff (approved):
Thanks!

PR #187, opened 2022-06-30
Switch to plain Docker SDK

mark-rushakoff:
This switches from using ory/dockertest to using the official Docker SDK.

The initial purpose of switching was to rule out dockertest as a cause for containers being stuck in the created state, but unfortunately the symptom persists after this change. However, it is at least @DavidNix's and my opinion that the official Docker SDK is the preferred choice moving forward. The SDK offers full control over any available Docker features, whereas dockertest abstracts and hides many implementation details.

This was a mostly mechanical change. There are a few `context.TODO()` instances in places where a function was not yet accepting a Context; those can get cleaned up in separate PRs.


chain/internal/tendermint/tendermint_node.go
```diff
@@ -192,51 +187,56 @@ func (tn *TendermintNode) CreateNodeContainer(ctx context.Context, additionalFla
...
-	cont, err := tn.Pool.Client.CreateContainer(docker.CreateContainerOptions{
-		Name: tn.Name(),
-		Config: &docker.Config{
-			User:         dockerutil.GetDockerUserString(),
-			Cmd:          cmd,
-			Hostname:     tn.HostName(),
+	cc, err := tn.DockerClient.ContainerCreate(
+		ctx,
+		&container.Config{
+			Image: tn.Image.Repository + ":" + tn.Image.Version,
```
jtieri, line 193: This is the third time I've seen something along the lines of, `tn.Image.Repository + ":" + tn.Image.Version`

Should this be it's own function? Something like `DockerImageName() or ImageName()`

jtieri (approved):
I believe I counted 6 locations where `containerImage.Repository + ":" + containerImage.Version` is being used to compose the image name which to me indicates that a function should be in use.

Besides for that and the few TODO's you have this looks good.

chain/internal/tendermint/tendermint_node.go
mark-rushakoff: I've changed these from `fmt.Sprintf("%s:%s", repo, version)` to `repo + ":" + version` because I was touching those lines anyway, and`Sprintf` just isn't necessary for a simple string concatenation.

It may as well be a method off the `Image` type. I'll open a followup PR.

PR #188, opened 2022-06-30
RelayerOption for additional startup flags

agouin:
`StartupFlags` `RelayerOption` to append to the `StartRelayer` `commander` command. Will need to be implemented if needed separately for future `RelayerCommander` implementations.

### DIFF
--- a/relayer/options.go
+++ b/relayer/options.go
@@ -43,7 +43,7 @@ type RelayerOptionStartupFlags struct {
 	Flags []string
 }
 
-func StartupFlags(flags []string) RelayerOption {
+func StartupFlags(flags ...string) RelayerOption {
 	return RelayerOptionStartupFlags{
 		Flags: flags,
 	}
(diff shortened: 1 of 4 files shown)


relayer/options.go
```diff
@@ -38,3 +38,15 @@ func ImagePull(pull bool) RelayerOption {
 }
 
 func (opt RelayerOptionImagePull) relayerOption() {}
+
+type RelayerOptionStartupFlags struct {
```
mark-rushakoff, line 42: Maybe `RelayerOptionExtraStartFlags`? Since the method is `Start`, and these are additional flags, not overriding the existing ones.

relayer/rly/cosmos_relayer.go
```diff
@@ -23,6 +23,12 @@ type CosmosRelayer struct {
 
 func NewCosmosRelayer(log *zap.Logger, testName, home string, pool *dockertest.Pool, networkID string, options ...relayer.RelayerOption) *CosmosRelayer {
 	c := commander{log: log}
+	for _, opt := range options {
+		switch typedOpt := opt.(type) {
```
mark-rushakoff, line 27: `switch o := opt.(type)` would probably be a more typical name given the short scope of the variable here.

mark-rushakoff (approved):

relayer/options.go
agouin: done

PR #192, opened 2022-07-05
Add example test for packet forward middleware

jtieri:
## Description

This PR adds a standalone test, for the IBC [packet-forward-middleware](https://github.com/strangelove-ventures/packet-forward-middleware), to an `examples` directory.

The test asserts that token transfers work in both directions using the middleware and also has the funds returned to the appropriate address in the case of one hop failing.

## Checklist

### DIFF
--- a/examples/packet_forward_test.go
+++ b/examples/packet_forward_test.go
@@ -15,7 +15,9 @@ import (
 )
 
 func TestPacketForwardMiddleware(t *testing.T) {
-	t.Skip()
+	if testing.Short() {
+		t.Skip()
+	}
 
 	home := ibctest.TempDir(t)
 	client, network := ibctest.DockerSetup(t)
--- a/examples/packet_forward_test.go
+++ b/examples/packet_forward_test.go
@@ -76,6 +76,7 @@ func TestPacketForwardMiddleware(t *testing.T) {
 	channels, err := r.GetChannels(ctx, eRep, gaia.Config().ChainID)
 	require.NoError(t, err)
 
+	// Start the relayer on both paths
 	err = r.StartRelayer(ctx, eRep, pathOsmoHub)
 	require.NoError(t, err)
 
@@ -86,11 +87,11 @@ func TestPacketForwardMiddleware(t *testing.T) {
 		func() {
 			err := r.StopRelayer(ctx, eRep)
 			if err != nil {
-				// do stuff
+				t.Logf("an error occured while stopping the relayer: %s", err)
 			}
 			for _, c := range chains {
 				if err = c.Cleanup(ctx); err != nil {
-					// do stuff
+					t.Logf("an error occured while stopping chain %s: %s", c.Config().ChainID, err)
 				}
 			}
 		},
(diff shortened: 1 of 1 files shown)


jackzampolin:
This is fire justin!

examples/packet_forward_test.go
```diff
@@ -0,0 +1,191 @@
...
+	gaiaUser := users[1]
+	junoUser := users[2]
+
+	osmosisBalOG, err := osmosis.GetBalance(ctx, osmosisUser.Bech32Address(osmosis.Config().Bech32Prefix), osmosis.Config().Denom)
+	require.NoError(t, err)
+
+	// Send packet from Osmosis->Hub->Juno
+	// receiver format: {intermediate_refund_address}|{foward_port}/{forward_channel}:{final_destination_address}
+	const transferAmount int64 = 100000
+	receiver := fmt.Sprintf("%s|%s/%s:%s", gaiaUser.Bech32Address(gaia.Config().Bech32Prefix), channels[1].PortID, channels[1].ChannelID, junoUser.Bech32Address(juno.Config().Bech32Prefix))
```
jackzampolin, line 111: might be a nice-to-have to expose a function in the module to format these

examples/packet_forward_test.go
```diff
@@ -0,0 +1,191 @@
...
+	transfertypes "github.com/cosmos/ibc-go/v3/modules/apps/transfer/types"
+	"github.com/strangelove-ventures/ibctest"
+	"github.com/strangelove-ventures/ibctest/ibc"
+	"github.com/strangelove-ventures/ibctest/test"
+	"github.com/strangelove-ventures/ibctest/testreporter"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestPacketForwardMiddleware(t *testing.T) {
```
mark-rushakoff, line 17: It would be great to get this as a "larger example" as detailed in https://go.dev/blog/examples#larger-examples and https://pkg.go.dev/testing#hdr-Examples.

But examples are expected to be niladic functions, so to get this to run as an example, I think we'd have to provide a mocked `testing.T`, which will end up being a whole can of worms.

Until it's realistic to put that together, why does this call `t.Skip`?

examples/packet_forward_test.go
```diff
@@ -0,0 +1,191 @@
...
+	t.Skip()
+
+	home := t.TempDir()
```
mark-rushakoff, line 20: Use `home := ibctest.TempDir(t)`.

examples/packet_forward_test.go
```diff
@@ -0,0 +1,191 @@
...
+	// Start the relayer on both paths
+	err = r.StartRelayer(ctx, eRep, pathOsmoHub)
+	require.NoError(t, err)
+
+	err = r.StartRelayer(ctx, eRep, pathJunoHub)
+	require.NoError(t, err)
+
+	t.Cleanup(
+		func() {
+			err := r.StopRelayer(ctx, eRep)
```
mark-rushakoff, line 88: I think this will hit a relayer bug -- IIRC there is a `container` field on the relayer type that will be overwritten on the second call to `r.StartRelayer`.

This is fine for now but we should fix that separately.

examples/packet_forward_test.go
jtieri: I was kinda thinking the same thing but didn't want to make the `Chain` interface too congested and wasn't sure where a one off utility function like this would live if it was outside of the test.

examples/packet_forward_test.go
jtieri: Thanks for the links Mark!

I added the `t.Skip` call as a last minute after thought because I wasn't sure if we want all the tests that could live in an `example` directory to run on calls to `go test` since there are intermittent Docker failures that could cause false positives. Happy to remove this line though if you think otherwise.

PR #193, opened 2022-07-06
Use empty entrypoint in every ContainerCreate call

### LINKED ISSUE #183: Override entry point in `Exec` function
Currently, the entry point is not overridden when creating a container to run a job. I encountered the following error

```bash
    reporter.go:226:
        	Error Trace:	testsuite.go:191
        	            				fee_middleware_test.go:287
        	            	failed to create faucet accounts: failed to create common account with name faucet: failed to create key with name "faucet" on chain simapp-a: exit code 1:  Error: unknown command "simd" for "simd"
        	            	Run 'simd --help' for usage.
```

The entrypoint to this specific container is `["simd"]`, as a result the command the ibctest builds is appended to this entrypoint.

I believe this should be solvable by just specifying an empty array for the entrypoint.
```golang
	Config: &docker.Config{
		Entrypoint: []string{},
                 // other fields
	},
```

mark-rushakoff:
The entrypoint was unspecified in two ContainerCreate calls, resulting
in the possibility of being unable to construct custom commands.

Closes #183.

### DIFF
--- a/chain/penumbra/penumbra_app_node.go
+++ b/chain/penumbra/penumbra_app_node.go
@@ -232,7 +232,7 @@ func (p *PenumbraAppNode) SendIBCTransfer(ctx context.Context, channelID, keyNam
 }
 
 func (p *PenumbraAppNode) CreateNodeContainer(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


chatton:
Thanks @mark-rushakoff !

PR #195, opened 2022-07-07
Use Docker volumes instead of host mounts for relayer containers

mark-rushakoff:
This may help with the hanging Docker containers, or it may not. But it
is importantly a step towards running ibctest against a remote Docker
engine.

This removes the home directory from the Docker Relayer implementations,
as all filesystem operations are now persisted to a short-lived Docker
Volume instead of the host mount.


DavidNix (approved):
🚢

PR #197, opened 2022-07-08
Upgrade test example

agouin:
Adds a public method for importing into integration tests, `CosmosChainUpgradeTest`. Adds CI test for a known working chain upgrade juno v6 to v7.

Chain upgrade procedure:
- Start chain
- Submit software upgrade proposal
- Vote yes from all validators
- Wait for voting period to expire
- Wait for chain to halt
- Shutdown all nodes
- Upgrade version on all nodes
- Start all nodes back up
- Wait for chain to produce blocks

### DIFF
--- a/chainspec.go
+++ b/chainspec.go
@@ -31,6 +31,8 @@ type ChainSpec struct {
 	GasAdjustment *float64
 	NoHostMount   *bool
 
+	ModifyGenesis func([]byte) ([]byte, error)
+
 	// Embedded ChainConfig to allow for simple JSON definition of a ChainSpec.
 	ibc.ChainConfig
 
@@ -107,6 +109,9 @@ func (s *ChainSpec) applyConfigOverrides(cfg ibc.ChainConfig) (*ibc.ChainConfig,
 	}
 	if s.NoHostMount != nil {
 		cfg.NoHostMount = *s.NoHostMount
+	}
+	if s.ModifyGenesis != nil {
+		cfg.ModifyGenesis = s.ModifyGenesis
 	}
 
 	// Set the version depending on the chain type.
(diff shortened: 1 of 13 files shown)


DavidNix (approved):
🚀

chainspec.go
```diff
@@ -31,6 +31,8 @@ type ChainSpec struct {
 	GasAdjustment *float64
 	NoHostMount   *bool
 
+	ModifyGenesis func([]byte) ([]byte, error)
```
colin-axner, line 34: I was actually thinking about this same issue

Another solution we thought of was to explicitly split the calls to `chain.Init` and `chain.Start`. Currently I believe they are called from the same function, but if you split them apart then a chain could be initialized, allowed to modify its own genesis (since it should have access to get/write the file), then it could continue by calling back into the function it halted out of

This is an interesting approach. In general I'd favor limiting the amount of functions in the interfaces defined (since I am not sure every chain has concepts of configurable genesis', though it seems natural they would)

agouin: The `Initialize` and `Start` methods were purposefully split up to allow node structure customization or other pre-start tasks to occur in between. The current `Interchain` builder calls both from the same method, but that wouldn't stop a one-off test case such as this upgrade test from doing the tasks without the `Interchain`. But, we get other useful things when using the `Interchain` such as the faucet account initialization so that wallets can request funds during the test, which avoids boilerplate in the tests.

I agree that genesis modification is a common need across all chains, so I kept the function signature simple with bytes in and out. But, this is not an interface method so it does not need to be implemented. `Chain` implementations can choose to ignore this from the `ChainSpec` entirely if they don't have a need for it. The cosmos implementation checks if it is changed from the zero-value of `nil` before calling.

colin-axner: Yup, sounds good. Nice work on this pr

Issue #200, opened 2022-07-11
Use docker volumes for chain containers


mark-rushakoff:
Following up on #195, use Docker volumes -- not host mounts -- for chain containers. There is a chance that this will address the macOS Docker engine hang, but even if it doesn't, this will allow using ibctest against a remote Docker engine.

PR #201, opened 2022-07-11
Use docker to retrieve genesis hashes

mark-rushakoff:
This is one step towards #200, using ephemeral containers to read the
genesis file content from the docker volume instead of reading from the
host disk.

Note that as of this commit, the docker volume is still a host mount,
but it should be seamless to switch this usage to a volume once
everything else for #200 is in place.

### DIFF
--- a/internal/dockerutil/fileretriever_test.go
+++ b/internal/dockerutil/fileretriever_test.go
@@ -1,0 +1,63 @@
+package dockerutil_test
+
+import (
+	"context"
+	"testing"
+
+	volumetypes "github.com/docker/docker/api/types/volume"
+	"github.com/strangelove-ventures/ibctest"
+	"github.com/strangelove-ventures/ibctest/internal/dockerutil"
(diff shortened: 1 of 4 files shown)


internal/dockerutil/fileretriever_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+
+	volumetypes "github.com/docker/docker/api/types/volume"
+	"github.com/strangelove-ventures/ibctest"
+	"github.com/strangelove-ventures/ibctest/internal/dockerutil"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestFileRetriever(t *testing.T) {
```
DavidNix, line 14: These tests are fast enough for unit tests? i.e. doesn't need `testing.Short` check?

DavidNix (approved):
One nit but otherwise 🚢

Issue #158, opened 2022-06-21
Extend chain interface to run arbitrary commands and return an output. 


DavidNix:
## Problem

Per ibc-go's team:

> I was thinking maybe when iterating on tests it would be nice to have some hooks that allow for arbitrary command execution, since as we develop the features the commands might change before they're done. Ideally the interfaces get updated once the feature is released (but that's a bit of a chicken and egg scenario!)

So they're trying to test unreleased, new features. We wouldn't update the chain interface until those features are finalized. As cian mentions, those commands could change a lot before the final release.

DavidNix:
> one other misc note: I noticed that the function dockerutil.HandleNodeJobError(exitCode, stdout, stderr, err) was not returning an error when I used the wrong command and was prompted with the help output

mark-rushakoff:
There have been several times when I needed to debug a miswritten command, and I had to update `HandleNodeJobError` to include the stdout and stderr (the two ignored parameters to the function).

https://github.com/strangelove-ventures/ibctest/blob/8b23c0fa73fa956d7261b644fdbbf34e04bb585c/internal/dockerutil/errors.go#L5-L13

I am on the fence about including that detail by default, but it would probably be more helpful if it were included, so you could quickly see an error like `--foo is not a valid flag`.

DavidNix:
These may be too cosmos-flavored, but you can reach node endpoints via these methods in `ibc.Chain`.

```
	// retrieves rpc address that can be reached by other containers in the docker network
	GetRPCAddress() string

	// retrieves grpc address that can be reached by other containers in the docker network
	GetGRPCAddress() string

	// GetHostRPCAddress returns the rpc address that can be reached by processes on the host machine.
	// Note that this will not return a valid value until after Start returns.
	GetHostRPCAddress() string

	// GetHostGRPCAddress returns the grpc address that can be reached by processes on the host machine.
	// Note that this will not return a valid value until after Start returns.
	GetHostGRPCAddress() string
```

PR #202, opened 2022-07-12
Execute arbitrary relayer commands

### LINKED ISSUE #174: Execute arbitrary commands for relayer.
Similar to https://github.com/strangelove-ventures/ibctest/issues/158 but for the relayer instead of the chain.

jtieri:
## Description

This PR adds the ability to execute arbitrary relayer commands similar to #158 

Should close #174 

## Checklist


relayer/docker.go
```diff
@@ -257,27 +263,56 @@ func (r *DockerRelayer) GetChannels(ctx context.Context, rep ibc.RelayerExecRepo
...
+	opts := dockerutil.ContainerOptions{
+		Env:   env,
+		Binds: r.Bind(),
+	}
+
+	var exitCode int
+	startedAt := time.Now()
+	stdout, stderr, err = job.Run(ctx, cmd, opts)
+	if err != nil {
+		exitCode = 1 // TODO this is possibly not correct, need some feedback on this
```
jtieri, line 301: I'm not 100% sure this is achieving exactly the same behavior as we had before.

DavidNix (approved):

relayer/docker.go
jtieri: I'll get a follow up PR opened separately, tomorrow at some point, that brings in the `ContainerError` type!

mark-rushakoff: Capturing non-zero exits as an error is subtle. This is much like writing an HTTP client library that treats a completed request with a 404 status as an error. 

There is a difference between "I ran the command to execution, and it exited 1" versus "I failed to run the command, because the binary `rly` was not on the path".

It looks like the `job.Run` signature is insufficient to distinguish this detail right now.

internal/dockerutil/image.go
```diff
@@ -87,14 +87,27 @@ type ContainerOptions struct {
...
 // Run blocks until the command completes. Thus, Run is not suitable for daemons or servers. Use Start instead.
 // A non-zero status code returns an error.
-func (image *Image) Run(ctx context.Context, cmd []string, opts ContainerOptions) (stdout, stderr []byte, err error) {
+func (image *Image) Run(ctx context.Context, cmd []string, opts ContainerOptions) ContainerExecResult {
 	c, err := image.Start(ctx, cmd, opts)
 	if err != nil {
-		return nil, nil, err
+		return ContainerExecResult{
+			Err:      err,
+			ExitCode: 1,
```
mark-rushakoff, line 107: I think typically `-1` would indicate an exceptional case where the execution did not happen.

chain/cosmos/chain_node.go
```diff
@@ -831,7 +831,8 @@ func (tn *ChainNode) Exec(ctx context.Context, cmd []string, env []string) ([]by
 		Env:   env,
 		Binds: tn.Bind(),
 	}
-	return job.Run(ctx, cmd, opts)
+	err := job.Run(ctx, cmd, opts)
```
mark-rushakoff, line 834: Can you rename these value usages to `res` instead of `err`? There are a handful of them.

ibc/relayer.go
```diff
@@ -82,6 +82,13 @@ type Relayer interface {
 	// Relayer implementations provided by the ibctest module will report true,
 	// but custom implementations may report false.
 	UseDockerNetwork() bool
+
+	// Exec runs an arbitrary command using Relayer's docker environment.
+	// Whether the invoked command is run in a one-off container or execing into an already running container
+	// is up to the relayer implementation.
+	//
+	// "env" are environment variables in the format "MY_ENV_VAR=value"
+	Exec(ctx context.Context, rep RelayerExecReporter, cmd []string, env []string) (stdout, stderr []byte, err error)
```
mark-rushakoff, line 91: This API is awkward in that it hides the error code inside the error value, and then you would have to have knowledge of the `dockerutil` package to get at the `ContainerExecResult` to get it. That makes me think we may want `ExecResult` to live in the `ibc` package.

`github.com/strangelove-ventures/ibctest/relayer` already imports both `ibc` and `dockerutil`; to avoid introducing a new `ibc`<->`dockertest` dependency, we may need both of those package to have their own `ExecResult` and then the `relayer` package would translate the `dockertest`'s type into the `ibc`'s type.

mark-rushakoff (commented):
I think there's a couple more small things to clean up.

chain/cosmos/chain_node.go
jtieri: Yeah, I'll get these changed.

ibc/relayer.go
jtieri: It didn't look like we needed the error code outside of `Exec` since it was only ever being used to pass into the `RelayerExecReporter` & the call to `HandleNodeJobError`. We still pass the error code into the test reporter and we get errors back that are formatted with the error code after the call to `image.Run` which makes `HandleNodeJobError` unused code.
https://github.com/strangelove-ventures/ibctest/blob/cb2a4ec120c97af72f4af6bf6b0c1ad49839def5/relayer/docker.go#L205-L208

If it seems like there could be additional use cases for accessing the error code that are outside of how we previously used it then I can definitely make the changes to return the entire struct by using some adapter like you suggested.

mark-rushakoff: I expect that we will eventually have tests that want to assert against relayer exit codes. It is possible those tests won't live in ibctest, but will rather be written by consumers of ibctest. I think we should expose the exit code, via something like `ExecResult`, to support that use case.

jtieri: I'll get those changes made as well!

jtieri: I came back to this PR after two weeks and wasn't sure if the suggestion to have a `ContainerExecResult` type in both `dockerutil` and `relayer` was to avoid a circular dependency, or forward thinking to avoid some issue later down the line. 
 
 I adjusted the API to return the struct vs. it's fields, but if it's still desirable to have a unique type in both packages then I can definitely get that taken care of.

DavidNix (approved):
LGTM.

I think it could be surprising to have to check the `result.Err` field. In Go, it's idiomatic to have the last return value be an err. However, I like how we hide this internal type from the public interfaces. So I'm 👍 with this API change. 

An alt design is an error type but keeping the return values `stderr, stdout []byte, err error`. 

Such as:

```go
type ExecError struct {
  error // embedding error auto-implements error interface or you implement Error() string manually.
  ExitCode int
  Stderr, Stdout []byte
}
```

Then the caller has to unwrap using `errors.As` is the tradeoff. 

This is not a change request, just an illustration of an alt design.

PR #208, opened 2022-07-14
Use docker API to copy cosmos contract into docker volume

mark-rushakoff:
I don't think anything in ibctest exercises this, but it looks like it
should work as intended.

For #200.


jtieri (approved):
Correct, I think Andrew so far just used this to test the Juno network upgrade.

PR #210, opened 2022-07-14
Extract SetVolumeOwner function in dockerutil

mark-rushakoff:
This behavior already existed internally in the relayer docker
boilerplate; extract it to a standalone function in the dockerutil
package so that it can be used more generally, in an upcoming change to
use volumes for cosmos chain containers.

Also move the busybox helpers to their own file.

### DIFF
--- a/internal/dockerutil/busybox.go
+++ b/internal/dockerutil/busybox.go
@@ -1,0 +1,56 @@
+package dockerutil
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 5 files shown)


jtieri (approved):
`volumeowner.go` is the meat of this PR and I'm somewhat of a Docker noob so I don't fully understand some of the choices here, but looks valid when comparing to the existing code that this is replacing. 👍

PR #211, opened 2022-07-14
Use Docker volumes for cosmos chains

mark-rushakoff:
A major milestone towards completing #200.

The remaining work will be to respect IBCTEST_SKIP_FAILURE_CLEANUP for
volumes, and adjusting the penumbra and internal/tendermint packages to
also use Docker volumes.


mark-rushakoff:
This commit ended up as part of #219.

PR #212, opened 2022-07-15
Extract mocktesting package

mark-rushakoff:
To add tests around an upcoming feature (opt-in retention of Docker
volumes on test failure), I was going to need a third instance of a mock
*testing.T.

This extracts a more full-featured mock instance of *testing.T to be
used in existing and future tests that need to assert against
interactions with real tests.

PR #213, opened 2022-07-15
Use IBCTEST_SKIP_FAILURE_CLEANUP to retain Docker volumes

mark-rushakoff:
(Draft until #212 is merged.)

Only if that environment variable is non-empty and the test fails.

As of this commit, this only skips calling docker volume prune on the
volumes. The next step in this sequence will be to add an option to copy
the volumes' content to local disk.


DavidNix (approved):
❤️

PR #215, opened 2022-07-19
Default to Docker containers' default users

mark-rushakoff:
Setting a different user by default creates strange permission issues
that manifest differently on macOS and Linux.


mark-rushakoff:
This commit ended up as part of #219.

PR #218, opened 2022-07-21
Migrate remaining direct file writes to docker operations

mark-rushakoff:
I think this is all the outstanding os.WriteFile and os.ReadFile calls.
This should enable switching the host mounts to docker volumes in an
upcoming commit.

For #200.

### DIFF
--- a/chain/internal/tendermint/tendermint_node.go
+++ b/chain/internal/tendermint/tendermint_node.go
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


chain/internal/tendermint/tendermint_node.go
```diff
@@ -137,25 +138,18 @@ func (tn *TendermintNode) PrivValKeyFilePath() string {
...
-}
-
 // Bind returns the home folder bind point for running the node
 func (tn *TendermintNode) Bind() []string {
 	return []string{fmt.Sprintf("%s:%s", tn.Dir(), tn.HomeDir())}
 }
 
 func (tn *TendermintNode) HomeDir() string {
-	return filepath.Join("/tmp", tn.Chain.Config().Name)
+	return path.Join("/tmp", tn.Chain.Config().Name)
```
DavidNix, line 147: Just curious: Why the change from `filepath` to `path`? Docker volumes always `/`?

DavidNix (approved):

chain/internal/tendermint/tendermint_node.go
mark-rushakoff: Correct, within the Docker volumes, we can rely on forward slashes for paths.

PR #219, opened 2022-07-21
Migrate to Docker volumes

### LINKED ISSUE #200: Use docker volumes for chain containers

mark-rushakoff:
Complete the migration from host mounts to Docker volumes.

I wanted to split this into smaller steps, but it seemed that this last step needed to happen all at once.

Switch all the chains to using Docker volumes, default to the containers' default Docker user in most circumstances, and add an additional label to volumes to make post hoc debugging a little simpler.

Closes #200.

### DIFF
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -14,10 +14,12 @@ import (
 	bankTypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	chanTypes "github.com/cosmos/ibc-go/v4/modules/core/04-channel/types"
 	dockertypes "github.com/docker/docker/api/types"
+	volumetypes "github.com/docker/docker/api/types/volume"
 	"github.com/docker/docker/client"
 	"github.com/strangelove-ventures/ibctest/chain/internal/tendermint"
 	"github.com/strangelove-ventures/ibctest/ibc"
 	"github.com/strangelove-ventures/ibctest/internal/blockdb"
+	"github.com/strangelove-ventures/ibctest/internal/dockerutil"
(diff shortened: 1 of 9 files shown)


DavidNix:
👀

chain/cosmos/cosmos_chain.go
```diff
@@ -262,16 +266,16 @@ func (c *CosmosChain) GetGasFeesInNativeDenom(gasPaid int64) int64 {
...
 	networkID string,
-) {
-	var chainNodes []*ChainNode
+) error {
 	count := c.numValidators + c.numFullNodes
 	chainCfg := c.Config()
 	for _, image := range chainCfg.Images {
 		rc, err := cli.ImagePull(
-			context.TODO(),
+			ctx,
```
DavidNix, line 278: 👏

DavidNix (approved):
LGTM!

So this alleviates or potentially prevents docker hangs on Macs?

mark-rushakoff:
Since implementing and running this locally, I haven't seen any more Docker Engine hangs on Docker for Mac. (But, I also had to do an OS upgrade partway through to work around a kernel panic when trying to run [UTM](https://mac.getutm.app/).)

Even if this doesn't 100% address the Docker hangs, it should fully enable running ibctest against a remote Docker engine simply by setting the `DOCKER_HOST` environment variable.

chain/cosmos/cosmos_chain.go
jackzampolin: TIL `context.TODO()`

PR #222, opened 2022-07-26
Remove Chain Cleanup

agouin:
Cleanup no longer required since the permissions issue is gone now with the docker volumes integration


mark-rushakoff (approved):
Nice removal. If we need it again we can easily add it back later.

PR #223, opened 2022-07-26
Tail daemon logs

agouin:
Ran into the need for this with #67 . Noticed that we are using a log tail of 50 for some commands that will have longer output than that. I set the default for the `Exec` calls to not tail (return all logs), and added a way to customize with an opt `LogTail`.

This also brings back the daemon container logs in the case of test failure. The default is `50` lines, but can be customized with `CONTAINER_LOG_TAIL` env var.

Additionally, daemon container logs can be printed regardless of test failure with `SHOW_CONTAINER_LOGS` env var != ""

### DIFF
--- a/internal/dockerutil/image.go
+++ b/internal/dockerutil/image.go
@@ -75,6 +75,11 @@ func NewImage(logger *zap.Logger, cli *client.Client, networkID string, testName
 	return i
 }
 
+const (
+	LogTailAll     = 0
+	LogTailDefault = 50
+)
+
 // ContainerOptions optionally configures starting a Container.
 type ContainerOptions struct {
 	// bind mounts: https://docs.docker.com/storage/bind-mounts/
@@ -85,6 +90,9 @@ type ContainerOptions struct {
 
 	// If blank, defaults to the container's default user.
 	User string
+
+	// If non-zero, will limit the amount of log lines returned
+	Tail uint64
 }
 
 // Run creates and runs a container invoking "cmd". The container resources are removed after exit.
@@ -96,7 +104,7 @@ func (image *Image) Run(ctx context.Context, cmd []string, opts ContainerOptions
 	if err != nil {
 		return nil, nil, err
 	}
-	return c.Wait(ctx)
+	return c.Wait(ctx, opts.Tail)
 }
 
 func (image *Image) imageRef() string {
@@ -232,7 +240,7 @@ type Container struct {
 // A non-zero status code returns an error.
 //
 // Wait implicitly calls Stop.
-func (c *Container) Wait(ctx context.Context) (stdout, stderr []byte, err error) {
+func (c *Container) Wait(ctx context.Context, tail uint64) (stdout, stderr []byte, err error) {
 	waitCh, errCh := c.image.client.ContainerWait(ctx, c.containerID, container.WaitConditionNotRunning)
 	var exitCode int
 	select {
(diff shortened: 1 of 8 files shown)


internal/dockerutil/image.go
```diff
@@ -252,11 +260,15 @@ func (c *Container) Wait(ctx context.Context) (stdout, stderr []byte, err error)
...
 
-	rc, err := c.image.client.ContainerLogs(ctx, c.containerID, types.ContainerLogsOptions{
+	logOpts := types.ContainerLogsOptions{
 		ShowStdout: true,
 		ShowStderr: true,
-		Tail:       "50",
-	})
+	}
+	if tail != 0 {
+		logOpts.Tail = fmt.Sprint(tail)
```
mark-rushakoff, line 268: Could be `strconv.FormatUint(tail, 10)` if you want a slightly more focused function.

mark-rushakoff (approved):
`tailLen` or `tailLines` might be a slightly more precise name for the parameters and fields. But `Tail` is the name of the field in the docker `ContainerLogsOptions` struct, so I guess `tail` is more consistent. Change or leave it at your discretion.

internal/dockerutil/image.go
```diff
@@ -232,7 +240,7 @@ type Container struct {
 // A non-zero status code returns an error.
 //
 // Wait implicitly calls Stop.
-func (c *Container) Wait(ctx context.Context) (stdout, stderr []byte, err error) {
+func (c *Container) Wait(ctx context.Context, tail uint64) (stdout, stderr []byte, err error) {
```
DavidNix, line 243: Explaining what the arg "tail" is for in the doc comment would be useful.

internal/dockerutil/image.go
```diff
@@ -75,6 +75,11 @@ func NewImage(logger *zap.Logger, cli *client.Client, networkID string, testName
 	return i
 }
 
+const (
+	LogTailAll     = 0
```
DavidNix, line 79: Curious, can we instead default to all logs all the time?

Then we wouldn't need to worry about this extra argument and complexity.

internal/dockerutil/image.go
DavidNix: Also (negligible in this case) but strconv will be more performant than `fmt` alternatives.

internal/dockerutil/image.go
agouin: After digging a bit, for the `Exec` case, all logs are probably fine. The tail was more useful for the daemons to log out the end of their logs in case of a test failure for example. I will change this to not tail for the `Exec` jobs.

We lost the daemon logs which were printed out with tail 50 in the case of an env var in this PR, it would be nice to bring them back in some form:
https://github.com/strangelove-ventures/ibctest/pull/173/files#diff-4a9052564209349d7ee9d73566c8e6a5ef2e18983757bbdfcabc2912c38b4bb0

agouin: I brought back the daemon tailing in case of test failure and added `SHOW_CONTAINER_LOGS` and `CONTAINER_LOG_TAIL` env vars for use when debugging.

internal/dockerutil/image.go
agouin: done

internal/dockerutil/image.go
agouin: done

internal/dockerutil/setup.go
```diff
@@ -89,8 +91,9 @@ func DockerSetup(t DockerSetupTestingT) (*client.Client, string) {
 // dockerCleanup will clean up Docker containers, networks, and the other various config files generated in testing
 func dockerCleanup(t DockerSetupTestingT, cli *client.Client) func() {
 	return func() {
+		showContainerLogs := os.Getenv("SHOW_CONTAINER_LOGS")
```
mark-rushakoff, line 94: Instead of a later check that `showContainerLogs != ""`, can you set this as a bool, i.e. `showContainerLogs := os.Getenv("SHOW_CONTAINER_LOGS") != ""`?

mark-rushakoff (approved):

internal/dockerutil/setup.go
agouin: done

PR #224, opened 2022-07-27
refactor: Move BroadcastTx into cosmos package

DavidNix:
The `ibctest.BroadcastTx` and the `broadcast.Broadcaster` interface was specific to Cosmos only. The ibctest root package attempts to be chain agnostic. 

All chain-specific functionality should live in an exported chain package; in this case `chain/cosmos`. 

The PR moves `BroadcastTx` into `chain/cosmos`, deprecates the old function, and removes the `Broadcaster` interface.


DavidNix:
@chatton fyi

broadcast.go
```diff
@@ -2,39 +2,12 @@ package ibctest
...
-	if err != nil {
-		return sdk.TxResponse{}, err
-	}
-
-	return broadcaster.UnmarshalTxResponseBytes(ctx, txBytes)
+// BroadcastTx is deprecated.
+//
+// Deprecated: Use chain/cosmos.BroadcastTx for broadcasting to chains built with Cosmos SDK.
+func BroadcastTx(ctx context.Context, broadcaster any, broadcastingUser any, msgs ...any) (any, error) {
+	panic(fmt.Errorf("BroadcastTx is deprecated; use chain/cosmos.BroadcastTx"))
```
chatton, line 12: If something is deprecated I would expect it to function as-is until it is removed at a later date. In this case maybe it's fine to just remove this function altogether rather than mark as deprecated. WDYT?

mark-rushakoff (approved):

broadcast.go
DavidNix: I opted for a panic here to warn the consumer where the new functionality lives because it may not be obvious. In this case, the consumer also gets a compile time error because of the `any` refactors, specifically with the return value. But given I think ibc-go is the main consumer, and they are aware, I'll simply remove it now. 

@chatton Our current ibctest attitude is the cliche of "move fast and break things." We are currently not optimizing for backwards compatability. The library consumer is expected to pin to a specific git sha. When we move to a more traditional sematic versioning, we'll optimize more for backwards compat. At this time, I don't have a timeline for moving to semver.

PR #230, opened 2022-08-03
Polkadot key derivation

agouin:
Phase 1 of #67 
substrate has a specific way of deriving the built-in dev keys from names "alice", "bob", etc.

This ports that implementation to go.

ed25519 - grandpa (consensus) key
sr25519 - account and stash keys
secp256k1 - beefy (ecdsa) key

A test is also included for generating a p2p key from a hex string and deriving the peer id from that.


chain/polkadot/keys_test.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func Test_DeriveSr25519FromNameAccount(t *testing.T) {
+	privKeyAccount, err := polkadot.DeriveSr25519FromName([]string{"Alice"})
+	require.NoError(t, err, "error deriving account sr25519 private key")
+
+	pubKeyAccount := make([]byte, 32)
+	for i, mkByte := range privKeyAccount.Public().Encode() {
+		pubKeyAccount[i] = mkByte
+	}
```
mark-rushakoff, line 48: I assume this was written in this style due to an error like `invalid operation: C() (value of type [4]byte) (slice of unaddressable value)`.

Instead of the byte-by-byte copy, put the array in a temporary variable and then slice it:

```go
b := privKeyAccount.Public().Encode()
pubKeyAccount := b[:]
// Or if there is a chance that the array b may change, copy it instead:
pubKeyAccount = append([]byte(nil), b[:]...)
```

go.mod
```diff
@@ -153,6 +183,12 @@ require (
...
 
+replace github.com/vedhavyas/go-subkey => github.com/strangelove-ventures/go-subkey v1.0.7
+
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
 replace github.com/keybase/go-keychain => github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
+
+replace github.com/ChainSafe/go-schnorrkel => github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d
+
+replace github.com/ChainSafe/go-schnorrkel/1 => github.com/ChainSafe/go-schnorrkel v1.0.0
```
mark-rushakoff, line 194: The replace directives might be okay for now, but replace directives in general do not carry over through dependencies. That is, if `xchain` imports `ibctest`, and if they use ibctest's `chain/polkadot` package, they will fail to import `.../go-schnorrkel/1` unless they copy the replace lines here.

Forcing the importer to copy the replace directives is a clunky solution, but it works.

One alternative would be to fork go-schnorrkel and have both revisions under some `strangelove/schnorrkel` repo. We can do that in a way that retains the git history, so that we can show that the source isn't modified. OTOH this repo is just for test, so the git history of a key encoding library is less important than a production chain.

mark-rushakoff (approved):
Just a couple small comments, but approving now.

chain/polkadot/keys.go
```diff
@@ -0,0 +1,118 @@
...
+	p2pCrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"golang.org/x/crypto/blake2b"
+)
+
+const (
+	ss58Ed25519HDKDPrefix = "Ed25519HDKD"
+	ss58Secp256k1Prefix   = "Secp256k1HDKD"
+)
+
+var DEV_SEED, _ = hex.DecodeString("fac7959dbfe72f052e5a0c3c8d6530f202b02fd8f9f5ca3580ec8deb7797479e")
```
jackzampolin, line 21: should this be in a test file?

go.mod
jackzampolin: I'm fine forking them to strangelove, but how do we keep them up to date?

chain/polkadot/keys.go
```diff
@@ -0,0 +1,118 @@
...
+
+	schnorrkel "github.com/ChainSafe/go-schnorrkel/1"
+	"github.com/StirlingMarketingGroup/go-namecase"
+	p2pCrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"golang.org/x/crypto/blake2b"
+)
+
+const (
+	ss58Ed25519HDKDPrefix = "Ed25519HDKD"
+	ss58Secp256k1Prefix   = "Secp256k1HDKD"
```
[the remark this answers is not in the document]
agouin: done

chain/polkadot/keys.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if _, err := hasher.Write(toHash); err != nil {
+		return nil, fmt.Errorf("error writing data to hasher: %w", err)
+	}
+
+	newKey := hasher.Sum(nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("error deriving: %w", err)
+	}
```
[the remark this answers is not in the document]
agouin: good catch

chain/polkadot/keys.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	privKey := make([]byte, 32)
+	for i, mkByte := range miniSecret.ExpandEd25519().Encode() {
+		privKey[i] = mkByte
+	}
+	pubKey := make([]byte, 32)
+	for i, mkByte := range miniSecret.Public().Encode() {
+		pubKey[i] = mkByte
+	}
```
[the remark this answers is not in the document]
agouin: yes, dead code now. Removed.

chain/polkadot/keys_test.go
agouin: done

chain/polkadot/keys.go
agouin: This is the polkadot dev seed that is used as the base for the dev name-based key derivation

https://github.com/polkadot-js/common/blob/2d4509c32ab1bcddb85c2db16c8d40d03e85aae5/packages/keyring/src/defaults.ts#L8

go.mod
mark-rushakoff: I doubt the repos are very active, so we would have to set up some manual notification of new commits and write a script to merge in the upstream changes. It would be mildly tedious, but infrequent, so I think the "once in a while, run a script by hand" approach is a better use of time than investing into fully automating it.

PR #231, opened 2022-08-05
[WIP] Test if relayers are keeping clients alive

boojamya:
This PR adds a test to see if relayers are autoupdating light clients.
This is needed on low traffic paths.

In order to accomplish this, the option to pass in a `--client-tp` (client trusting period) flag when creating clients with the Go Relayer was needed.

### DIFF
--- a/examples/autoClientUpdate_test.go
+++ b/examples/autoClientUpdate_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/strangelove-ventures/ibctest"
 	"github.com/strangelove-ventures/ibctest/ibc"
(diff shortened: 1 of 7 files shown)


examples/autoClientUpdate_test.go
```diff
@@ -0,0 +1,102 @@
...
+		//NEED TO FIGURE OUT HOW LOGGING WORKS
+		t.Log(out)
+		t.Log(serr)
+	}
+	t.Log("OUT GAIA1: ", out)
+	t.Log("SERR GAIA2: ", serr)
+
+	// log queriered client trusing period to verify tp (chain2)
+	command = []string{
+		"gaiad", "query", "ibc", "client", "state", gaia2ClientID,
```
jackzampolin, line 82: you should use something something like `gaia1.Binary` instead of `gaiad` directly.

boojamya:
There is a specific test case written in the Go Relayer here: https://github.com/cosmos/relayer/blob/main/interchaintest/client_threshold_test.go

Closing as stale.

PR #234, opened 2022-08-08
Add polkadot relay chain node

agouin:
Phase 2 of #67 

Polkadot chains are composed of the relay chain (polkadot chain) nodes and the parachain nodes for involved parachains (e.g. Composable). This brings in the relay chain node docker orchestration methods.


chain/polkadot/relay_chain_node.go
```diff
@@ -0,0 +1,327 @@
...
+func (p *RelayChainNode) AccountAddress() (string, error) {
+	pubKey := make([]byte, 32)
+	for i, mkByte := range p.AccountKey.Public().Encode() {
+		pubKey[i] = mkByte
+	}
+	return EncodeAddressSS58(pubKey)
+}
+
+// AccountAddress returns the ss58 encoded stash address
+func (p *RelayChainNode) StashAddress() (string, error) {
```
DavidNix, line 124: nit: godoc and method name mismatch. `// AccountAddress` -> `// StashAddress`

chain/polkadot/relay_chain_node.go
```diff
@@ -0,0 +1,327 @@
...
+	pubKey := make([]byte, 32)
+	for i, mkByte := range p.StashKey.Public().Encode() {
+		pubKey[i] = mkByte
+	}
+	return EncodeAddressSS58(pubKey)
+}
+
+// AccountAddress returns the ss58 encoded secp256k1 address
+func (p *RelayChainNode) EcdsaAddress() (string, error) {
```
DavidNix, line 133: nit: Godoc and method name again.

chain/polkadot/relay_chain_node.go
```diff
@@ -0,0 +1,327 @@
...
+		fmt.Sprintf("--public-addr=%s", multiAddress),
+		"--base-path", p.NodeHome(),
+	}
+	p.logger().
+		Info("Running command",
+			zap.String("command", strings.Join(cmd, " ")),
+			zap.String("container", p.Name()),
+		)
+
+	cc, err := p.DockerClient.ContainerCreate(
```
DavidNix, line 247: Is there functionality missing from our `dockerutil.Image` helpers? 

I favor using our docker abstractions before dipping into the lower-level docker client API in other parts of the codebase. Our abstractions will ensure things like proper cleanup. In other words, isolate docker low-level to `dockerutil` pkg and try not to leak it elsewhere. 

If `dockerutil.Image` is missing functionality, we can extend or refactor by offering more config options and perhaps another method for `CreatorContainer()` that creates the container but does not run it. Then, this chain node can hold onto a `dockerutil.Image` instead of a container id.

DavidNix (commented):
Had a question about using low-level docker api. 

An additional question is the code looks good but it is not used anywhere (so effectively dead code).

Do you think it's worthwhile to include a small test (skipped in short mode) to prove that we can run this new chain?

agouin:
> Do you think it's worthwhile to include a small test (skipped in short mode) to prove that we can run this new chain?

The test is coming, includes the relay chain nodes and the parachain nodes, see #67. This is an attempt to keep the PRs minimal.

chain/polkadot/relay_chain_node.go
agouin: This follows the same strategy for starting the daemon process as the others, `chain_node`, `tendermint_node`, and `penumbra_app_node`.

chain/polkadot/relay_chain_node.go
agouin: done

chain/polkadot/relay_chain_node.go
agouin: done

chain/polkadot/relay_chain_node.go
DavidNix: Ah so it is. I feel that's not ideal design and the better one is to depend on our abstractions. In the case of proper cleanup, a new contributor (including our future selves) may not realize the subtleties of tagging, for example. 

There's also quite a bit of docker client duplication with the various chain nodes, so another signal it's time for an abstraction imo. 

But all that's to say; it's probably a future refactor.

agouin: Yes that would be a great consolidation, I created #237 as a follow-on

PR #238, opened 2022-08-09
Polkadot parachain node

agouin:
Phase 3 of #67 

Polkadot chains are composed of the relay chain (polkadot chain) nodes and the parachain nodes for involved parachains (e.g. Composable). This brings in the parachain node docker orchestration methods.

### DIFF
--- a/chain/polkadot/parachain_node.go
+++ b/chain/polkadot/parachain_node.go
@@ -1,0 +1,274 @@
+package polkadot
+
+import (
+	"context"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	gsrpc "github.com/centrifuge/go-substrate-rpc-client/v4"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/client"
+	p2pCrypto "github.com/libp2p/go-libp2p-core/crypto"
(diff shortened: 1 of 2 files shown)


chain/polkadot/parachain_node.go
```diff
@@ -0,0 +1,274 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/avast/retry-go/v4"
+	gsrpc "github.com/centrifuge/go-substrate-rpc-client/v4"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/client"
+	p2pCrypto "github.com/libp2p/go-libp2p-core/crypto"
```
mark-rushakoff, line 17: Import names should be all lowercase.

mark-rushakoff (approved):
A couple small style notes but otherwise looks ready to ship.

chain/polkadot/parachain_node.go
DavidNix: You'd think a linter somewhere should catch that.

mark-rushakoff: Yeah, you'd think so. The blog post on this is 7.5 years old, so there's been plenty of time for someone to write one: https://go.dev/blog/package-names

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns...

DavidNix: There may be a linter with golangci-lint we need to activate. There's so many of them; hard to keep track.

PR #245, opened 2022-08-12
Polkadot ChainFactory, ChainSpec, and test

agouin:
Wires up `polkadot` chain in `ChainFactory`, only handling `composable` as a supported parachain for now.

Adds config overrides in `ChainSpec` for `polkadot` to configure the docker image versions.

Adds test for polkadot chain to ensure relay chain and parachain can start and that composable chain produces 10 blocks.


DavidNix (approved):
🚀

PR #248, opened 2022-08-15
Add ability for relayers to pass in options when creating clients. 

boojamya:
## Description

This PR adds the ability to pass in `CreateClientOptions` when a relayer creates a client. This PR only adds the single option of "client trusting period", but it should be easy to add other options and flags in the future.

This flow of adding options mimics `CreateChannelOptions`.

If no options are passed in, default options are used. Default options are determined by the relayer. For the Go Relayer's case, default trusting period is 85% of the unbonding time.

A test utilizing these options will be in a PR to follow.

## Checklist


agouin (approved):
:drum: :guitar:

Issue #220, opened 2022-07-22
ibctest.User/ibctest.RelayerWallet types could maybe be combined


charleenfei:
the difference is seems to be only in use case of the type and not the type itself. 

for example, currently, we have to do things like this ```chainAWallet.Bech32Address(chainB.Config().Bech32Prefix)``` to get the bech32 address from a User type, but the RelayerWallet type offers it out of the box. just a little confusing as from the chain/SDK POV these could be the same account address types right?

jackzampolin:
Thank you for the issue @charleenfei, that should be a relatively quick change.

PR #249, opened 2022-08-16
Combined User and Relayer into Wallet

danbryan:
## Description

This PR combined the User and Relayer into Commontype Wallet


interchain_test.go
```diff
@@ -206,7 +206,7 @@ func TestInterchain_CreateUser(t *testing.T) {
 
 		user := ibctest.GetAndFundTestUserWithMnemonic(t, ctx, keyName, mnemonic, 10000, gaia0)
 		require.NoError(t, test.WaitForBlocks(ctx, 2, gaia0))
-		require.NotEmpty(t, user.Address)
+		require.NotEqual(t, "", user.Address)
```
DavidNix, line 209: nit:

A more suitable matcher here is `require.NotEmpty`

DavidNix (approved):
Nice cleanup!

interchain_test.go
DavidNix: Or `require.NotZero` which in Go, refers to [zero values](https://go.dev/tour/basics/12).

danbryan: resolved with 
https://github.com/strangelove-ventures/ibctest/pull/255

danbryan:
closes #220

PR #253, opened 2022-08-17
Make events default rly start method

jtieri:
## Description

This PR makes the `-p events` flag the default method for starting `rly`.

I kept running the tests consecutively and without adding a slightly greater initial block history the tests were failing intermittenly.

## Checklist

### DIFF
--- a/relayer/rly/cosmos_relayer.go
+++ b/relayer/rly/cosmos_relayer.go
@@ -204,6 +204,8 @@ func (commander) RestoreKey(chainID, keyName, mnemonic, homeDir string) []string
 func (c commander) StartRelayer(pathName, homeDir string) []string {
 	cmd := []string{
 		"rly", "start", pathName, "--debug",
+		"-p", "events",
+		"-b", "50",
 		"--home", homeDir,
 	}
 	cmd = append(cmd, c.extraStartFlags...)


DavidNix (approved):

mark-rushakoff (changes requested):
I would prefer to see this as a separate relayer implementation, for now. Otherwise this makes it impossible to test the legacy implementation, doesn't it?

jtieri:
Hmm looks like the `TestPacketForwardMiddleware` is failing and looks like it has to do with some contention in trying to start two relayer processes for the two paths required in this test. I wasn't noticing it on my local machine because I was skipping the test with `--short`. After fiddling with it for the last hour or so it seems to be a more involved fix.

jtieri:
> I would prefer to see this as a separate relayer implementation, for now. Otherwise this makes it impossible to test the legacy implementation, doesn't it?

That is correct but iirc we had discussed it internally and agreed that because we are still testing the legacy implementation in the `relayer` repo's CI that this would be acceptable. It does seem like, however, this may require some additional work at this point to get more than one relayer instance working or to get one relayer instance relaying on multiple paths.

mark-rushakoff:
That’s fair, you can dismiss my review. A comment near the added flag would
be helpful though.

jtieri:
This is the default behavior in the relayer now, so no need to explicitly set this after #295

PR #265, opened 2022-08-30
Fix volume ownership when heighliner images don't have chown or chmod

agouin:
Problem: new slim heighliner images don't have `chown` or `chmod`.

Adds `UidGid` parameter to `ibc.DockerImage`. If empty, root user is assumed. If non-empty, `uid:gid` format is required so that busybox can change ownership without knowing the contents of the target image's `/etc/passwd`

With this, we don't need to assume that any other custom images would have `chown` or `chmod` either.

### DIFF
--- a/relayer/options.go
+++ b/relayer/options.go
@@ -16,11 +16,12 @@ type RelayerOptionDockerImage struct {
 	DockerImage ibc.DockerImage
 }
 
-func CustomDockerImage(repository string, version string) RelayerOption {
+func CustomDockerImage(repository string, version string, uidGid string) RelayerOption {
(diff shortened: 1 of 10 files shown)


relayer/options.go
```diff
@@ -16,11 +16,12 @@ type RelayerOptionDockerImage struct {
 	DockerImage ibc.DockerImage
 }
 
-func CustomDockerImage(repository string, version string) RelayerOption {
+func CustomDockerImage(repository string, version string, uidGid string) RelayerOption {
```
DavidNix, line 19: A doc comment here may be useful

PR #271, opened 2022-08-31
EncodingConfig override

agouin:
Allow overriding the default `EncodingConfig` for custom codecs.

Make `cosmos.DefaultEncoding()` public so that default config can be fetched, enriched, then used as an override.

Example usage:
https://github.com/strangelove-ventures/ibctest/blob/3669846acdc03d09f1b9756823952e199fe13667/examples/osmosis/osmosis.go#L10-L17

https://github.com/strangelove-ventures/ibctest/blob/3669846acdc03d09f1b9756823952e199fe13667/examples/osmosis/cosmos_chain_gamm_pool_test.go#L34

### DIFF
--- a/chain/cosmos/broadcaster.go
+++ b/chain/cosmos/broadcaster.go
@@ -135,7 +135,7 @@ func (b *Broadcaster) GetTxResponseBytes(ctx context.Context, user User) ([]byte
 // instance of sdk.TxResponse.
 func (b *Broadcaster) UnmarshalTxResponseBytes(ctx context.Context, bytes []byte) (sdk.TxResponse, error) {
 	resp := sdk.TxResponse{}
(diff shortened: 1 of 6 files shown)


DavidNix (commented):
I like the idea of exposing the default so the consumer can override. 

However, in the case of Osmosis example, is there any harm in this text context to simply register as many codecs as possible?

DavidNix (approved):

agouin:
> However, in the case of Osmosis example, is there any harm in this text context to simply register as many codecs as possible?

This requires importing osmosis for the codecs, which brings along the necessity to do go mod replaces on cosmos-sdk and iavl to the osmosis ones, which is undesirable in ibctest directly. For the osmosis example tests, putting them in their own module with the go mod overrides and using this override mechanism for the codecs seems to work well and be a clean separation.

PR #273, opened 2022-09-01
[v3] Client ICQ

agouin:


examples/interchain_queries_client_test.go
```diff
@@ -0,0 +1,304 @@
...
+)
+
+func modifyGenesisStride() func([]byte) ([]byte, error) {
+	return func(genbz []byte) ([]byte, error) {
+		g := make(map[string]interface{})
+		if err := json.Unmarshal(genbz, &g); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal genesis file: %w", err)
+		}
+
+		if err := dyno.Set(g, dayEpochLen, "app_state", "epochs", "epochs", dayEpochIndex, "duration"); err != nil {
```
jackzampolin, line 270: til `dyno` this is a nice way to deal with this in go

DavidNix: I haven't used `dyno`. It seems fine for our use case. 

I've also had success with http://jsoniter.com/migrate-from-go-std.html and https://github.com/Jeffail/gabs for other options if `dyno` ever gives us trouble.

agouin:
Closing this as this test has been added to the relayer along with stride ICQ support https://github.com/cosmos/relayer/pull/1076

PR #274, opened 2022-09-02
InterchainLink create client/channel opts

agouin:
Move `CreateClientOptions` and `CreateChannelOptions` to `InterchainLink` to allow differences for different paths.


jtieri (approved):
This is a good idea 👍

PR #277, opened 2022-09-08
Fix issue with custom matrix files

boojamya:
This PR fixes an issue when passing in custom matrix files.

If the chain is fully configured but not pre-configured, we need to register a new chain label. 

Closes: https://github.com/strangelove-ventures/ibctest/issues/263

### DIFF
--- a/chainspec.go
+++ b/chainspec.go
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 
 	"github.com/strangelove-ventures/ibctest/ibc"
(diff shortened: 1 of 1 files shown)


DavidNix (approved):

boojamya:
Whoops. Let me make it pass tests.
Making sure I have the  `ChainSpect[Name, ChainName]` and the `ChainConfig[Name]` and the `ChainLabel` all straight here. 🙀

PR #279, opened 2022-09-10
blockdb - begin and end block events

agouin:
Tracks begin and end block events in blocks db.

Ensures txs are saved with `Data` in json format so that `v_cosmos_messages` will be able to query even if transactions are unsuccessfully decoded

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -320,7 +320,7 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx, 
 	}
 	if len(blockRes.BeginBlockEvents) > 0 {
 		beginBlockTx := blockdb.Tx{
-			Data: []byte(`{"data":"begin_block"}`),
+			Data: []byte(`{"data":"begin_block","note":"this is a transaction artificially created for debugging purposes"}`),
 		}
 		beginBlockTx.Events = make([]blockdb.Event, len(blockRes.BeginBlockEvents))
 		for i, e := range blockRes.BeginBlockEvents {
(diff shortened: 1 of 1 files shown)


chain/cosmos/chain_node.go
```diff
@@ -285,29 +298,65 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx,
...
+			newTx.Events[j] = blockdb.Event{
+				Type:       e.Type,
+				Attributes: attrs,
+			}
+		}
+		txs = append(txs, newTx)
+	}
+	if len(blockRes.BeginBlockEvents) > 0 {
+		beginBlockTx := blockdb.Tx{
+			Data: []byte(`{"data":"begin_block"}`),
```
DavidNix, line 323: So we are artificially creating a tx in order to query these events?

Would this confuse a newcomer to cosmos who would then expect to see these txs on the blockchain? (i.e. if they look in mintscan). 

I didn't realize there were events not associated with a tx in cosmos. 

I'm ok with this as a workaround/hack, but perhaps the long-term solution is saving it in `tendermint_event` table which may or may not have a relationship to a tx but always has a relationship to a block. IIUC, these begin and end block events are 0:many to a block.

DavidNix (approved):

chain/cosmos/chain_node.go
DavidNix: You could also extend the `{"data": <begin/end block>, "note": "this is a transaction artificially created for debugging purposes"}`

agouin: I considered schema changes, but since it's really only for debugging, the effort did not seem necessary. Yes, the events are 0:many just like a typical tx, so it seemed to plug in well to the existing queries. Good call about the `note` property to aid in understanding, added.

PR #280, opened 2022-09-11
consolidate cosmos chain node commands

agouin:
DRY up the chain node flag usage within the cosmos `ChainNode`, exporting public helpers to get or execute the commands from test cases.

`BinCommand` - returns aggregate command of `command []string` args/flags after the chain node binary, with `--home` flag appended at the end.
`ExecBin` - executes aggregate command from `BinCommand`.
`NodeCommand` - returns aggregate command of `BinCommand` with additional flags `--node` and `--chain-id` appended, necessary when commands interact with the node URL.
`QueryCommand` - returns aggregate command of `NodeCommand` with prepended arg `query` and appended flag `--output json` to return response formatted as json.
`ExecQuery` - executes aggregate command from `QueryCommand`.
`ExecTx` - assembles aggregate command of `NodeCommand` with prepended arg `tx` and appended flags necessary for transaction broadcast (`--from`, `--gas-prices`, `--gas-adjustment`, `--keyring-backend`, `-y`), then executes it.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -361,11 +361,11 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx, 
 	return txs, nil
 }
 
-// ExecTx executes a transaction, waits for blocks, waits
-// for 2 blocks if successful, then returns the tx hash.
-func (tn *ChainNode) ExecTx(ctx context.Context, keyName string, command ...string) (string, error) {
+// TxCommand is a helper to retrieve a full command for broadcasting a tx
+// with the chain node binary.
+func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
 	command = append([]string{"tx"}, command...)
-	command = tn.NodeCommand(append(command,
+	return tn.NodeCommand(append(command,
 		"--from", keyName,
 		"--gas-prices", tn.Chain.Config().GasPrices,
 		"--gas-adjustment", fmt.Sprint(tn.Chain.Config().GasAdjustment),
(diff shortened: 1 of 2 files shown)


DavidNix (approved):
Improved LOC, which is great. My only holdup is perhaps this is too much indirection. As a dev or user trying to learn cosmos, it might be beneficial to see the entire command and args all in one spot. Perhaps a helper to generate the common, repeated flags. But that is a minor nit, and I think this is fine to merge.

agouin:
> Perhaps a helper to generate the common, repeated flags.

I did this for `BinCommand`, `NodeCommand`, and `QueryCommand`, but did not see an immediate use for `TxCommand`. You're right that this would be helpful for debugging, so added `TxCommand` also.

chain/cosmos/chain_node.go
```diff
@@ -479,17 +569,9 @@ type CosmosTx struct {
 }
 
 func (tn *ChainNode) SendIBCTransfer(ctx context.Context, channelID string, keyName string, amount ibc.WalletAmount, timeout *ibc.IBCTimeout) (string, error) {
-	command := []string{tn.Chain.Config().Bin, "tx", "ibc-transfer", "transfer", "transfer", channelID,
+	command := []string{
+		"ibc-transfer", "transfer", "transfer", channelID,
```
boojamya, line 573: Out of scope for this PR. But should we concerned about hard-coding the src-port of "transfer" ?

chain/cosmos/chain_node.go
```diff
@@ -362,6 +361,94 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx,
...
+	command = tn.BinCommand(command...)
+	return append(command,
+		"--node", fmt.Sprintf("tcp://%s:26657", tn.HostName()),
+		"--chain-id", tn.Chain.Config().ChainID,
+	)
+}
+
+// BinCommand is a helper to retrieve a full command for a chain node binary.
+// For example, if chain node binary is `gaiad`, and desired command is `gaiad keys show key1`,
+// pass ("keys", "show", "key1") for command to return the full command.
```
boojamya, line 416: This example is the same as the NodeCommand. Is that on purpose?

boojamya (approved):
I definitely share @DavidNix's concerns but I see the benefits of dryer code and hopefully ease of maintainability.
I trust your judgment. 👍

chain/cosmos/chain_node.go
agouin: yes, `NodeCommand` returns the command, `ExecNode` builds the command and also executes it.

chain/cosmos/chain_node.go
agouin: Only transfer port for now. A future need for different ports might involve a refactor to this method or a new method.

PR #281, opened 2022-09-11
[v3] expedited proposals test

### LINKED ISSUE #268: Add TestExpeditedProposals
https://github.com/osmosis-labs/osmosis/blob/1ec83c30f7537719e0154b08a51cb6f2d66455b3/tests/e2e/e2e_test.go#L219

agouin:
Introduces a new module for `examples/osmosis` which includes it's go mod replaces for cosmos-sdk and iavl.

Closes #268


DavidNix (approved):
LGTM!

One minor note, it may seem odd we're building a network with just 1 chain. IMO, this indicates this test may be better suited for a feature or unit test within the osmosis codebase. But, that is an ideal. In this case, some test coverage is better than no test coverage. So 🚢 !

jackzampolin (approved):

PR #287, opened 2022-09-13
Remove cosmos specific methods from chain interface

agouin:
These methods are only applicable to cosmos chains for now, so they can be removed from the `ibc.Chain` interface. Tests that want to use these methods can cast the chain to a `*CosmosChain`.


agouin:
Tests failing because of sed command usage in tendermint node for penumbra. Fix here #288

DavidNix (approved):

PR #288, opened 2022-09-13
Remove sed command calls from tendermint node

agouin:
heighliner tendermint images no longer have `sed` and this was a dirty hack anyways


DavidNix (approved):
Good cleanup. `sed` is 👎

PR #291, opened 2022-09-14
CosmosExternalChain

agouin:
Add `CosmosExternalChain` `ibc.Chain` implementation for a mechanism to use existing public chains in tests.

Use `Type: "cosmos_external"` in chain configuration to use this.


Reecepbcups:
closing as stale, linked in local-ic for a future update when we get to it & it makes sense

PR #292, opened 2022-09-14
[v3] Proposal Polling and Pools

agouin:
Add methods to `CosmosChain`:
- `QueryProposal` to fetch proposal information e.g. title, description, status, etc.
- `TextProposal` to submit a text governance proposal

Adds `cosmos.PollForProposalStatus` to poll for a proposal to change status, e.g. wait for proposal to transition from voting period to pass/fail status.

`CreatePool` was never fully implemented. This is used in #282 
Osmosis specific methods in their own file

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/cosmos/cosmos-sdk/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
@@ -109,7 +110,7 @@ func (tn *ChainNode) CliContext() client.Context {
 	cfg := tn.Chain.Config()
 	return client.Context{
 		Client:            tn.Client,
-		ChainID:           tn.Chain.Config().ChainID,
+		ChainID:           cfg.ChainID,
 		InterfaceRegistry: cfg.EncodingConfig.InterfaceRegistry,
 		Input:             os.Stdin,
 		Output:            os.Stdout,
@@ -691,6 +692,20 @@ func (tn *ChainNode) VoteOnProposal(ctx context.Context, keyName string, proposa
 	return err
 }
 
+// QueryProposal returns the status of a proposal.
+func (tn *ChainNode) QueryProposal(ctx context.Context, proposalID string) (*govtypes.Proposal, error) {
+	stdout, _, err := tn.ExecQuery(ctx, "gov", "proposal", proposalID)
+	if err != nil {
+		return nil, err
+	}
+	var proposal govtypes.Proposal
+	err = json.Unmarshal(stdout, &proposal)
+	if err != nil {
+		return nil, err
+	}
+	return &proposal, nil
+}
+
 // UpgradeProposal submits a software-upgrade proposal to the chain.
 func (tn *ChainNode) UpgradeProposal(ctx context.Context, keyName string, prop ibc.SoftwareUpgradeProposal) (string, error) {
 	command := []string{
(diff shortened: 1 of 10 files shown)


chain/cosmos/chain_node.go
```diff
@@ -745,14 +774,49 @@ func (tn *ChainNode) UnsafeResetAll(ctx context.Context) error {
 	return err
 }
 
-func (tn *ChainNode) CreatePool(ctx context.Context, keyName string, contractAddress string, swapFee float64, exitFee float64, assets []ibc.WalletAmount) error {
-	// TODO generate --pool-file
-	poolFilePath := "TODO"
-	_, err := tn.ExecTx(ctx,
+func (tn *ChainNode) CreatePool(ctx context.Context, keyName string, params ibc.PoolParams) (string, error) {
```
DavidNix, line 777: A doc comment would be useful here.

chain/cosmos/chain_node.go
```diff
@@ -745,14 +774,49 @@ func (tn *ChainNode) UnsafeResetAll(ctx context.Context) error {
...
+	}
+
+	numPools, ok := res["num_pools"]
+	if !ok {
+		return "", fmt.Errorf("could not find number of pools in query response: %w", err)
+	}
+	return numPools, nil
+}
+
+func (tn *ChainNode) SwapExactAmountIn(ctx context.Context, keyName string, coinIn string, minAmountOut string, poolIDs []string, swapDenoms []string) (string, error) {
```
DavidNix, line 813: Doc comment would also be useful here. As an outsider, the function name is not descriptive enough. 

Additionally, are `SwapExactAmountIn` and `CreatePool` osmosis specific? Do these methods apply to all chains? 

If not, one option is a new `cosmos.OsmosisChain` type. You can embed a `cosmos.ChainNode` to get all of its functionality. Otherwise, it appears as if all cosmos chains has this Osmosis-specific functionality.

DavidNix: Or, another alternative is to separate data from function. In other words, instead of creating new methods off the type, create functions that take the chain as an argument. 

E.g. 

```go
// Ideally, *ChainNode is replaced by an interface. 
// https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b
func OsmosisSwapExactAmountIn(chain *ChainNode, args...)
```

This will also further signal this new functionality only relates to Osmosis. 

I like this option better than a new Osmosis chain type.

chain/cosmos/cosmos_chain.go
```diff
@@ -315,8 +333,12 @@ func (c *CosmosChain) ExportState(ctx context.Context, height int64) (string, er
 }
 
 // Implements Chain interface
-func (c *CosmosChain) CreatePool(ctx context.Context, keyName string, contractAddress string, swapFee float64, exitFee float64, assets []ibc.WalletAmount) error {
-	return c.getFullNode().CreatePool(ctx, keyName, contractAddress, swapFee, exitFee, assets)
+func (c *CosmosChain) CreatePool(ctx context.Context, keyName string, params ibc.PoolParams) (string, error) {
```
DavidNix, line 336: `CreatePool` is now part of the Chain interface? I think the doc comment is stale.

ibc/types.go
```diff
@@ -205,3 +213,42 @@ type SoftwareUpgradeProposal struct {
 	Height      uint64
 	Info        string // optional
 }
+
+// PoolParams defines parameters for creating a gamm liquidity pool
+type PoolParams struct {
```
DavidNix, line 218: One source of my confusion is the new types like `ibc.Params`. IBC is used for proposals and software upgrades? I thought these were all done on chain. Can you submit a proposal to another chain via IBC? Do these new types apply to all chains, even non-cosmos?

agouin: good call, these types belong in `cosmos`

chain/cosmos/cosmos_chain.go
agouin: yes removed it yesterday from the ChainInterface, will clean up

chain/cosmos/chain_node.go
agouin: done

chain/cosmos/chain_node.go
agouin: Good idea, added an `osmosis.go` file to the `cosmos` package with the osmosis specific functions

jackzampolin (approved):

boojamya (approved):

PR #294, opened 2022-09-16
move builtin chains to yaml file

boojamya:
## Description

This PR moves the builtin chains from a hardcoded struct to a yaml file.

This makes it straight forward for testers to view and edit the "preconfigured" settings.

Ibctest will embed ./confiugredChains.yaml. It will refer to this embedded version UNLESS the environment variable  IBCTEST_CONFIGURED_CHAINS is set to a path for "custom" preconfigured chains and settings.

This should satisfy: https://github.com/strangelove-ventures/ibctest/issues/175

### DIFF
--- a/chainfactory.go
+++ b/chainfactory.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/strangelove-ventures/ibctest/v5/chain/cosmos"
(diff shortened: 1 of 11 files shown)


chainfactory.go
```diff
@@ -42,16 +45,40 @@ type BuiltinChainFactory struct {
...
+// initBuiltinChainConfig returns an ibc.ChainConfig mapping all configured chains
+func initBuiltinChainConfig() (map[string]ibc.ChainConfig, error) {
+
+	// checks if configuredChains.yaml is found in cwd or parent of cwd,
+	// otherwise it will use the the embedded version of configuredChains.yaml
+	configuredChainsFile := ""
+	if _, err := os.Stat("./configuredChains.yaml"); err == nil {
+		configuredChainsFile = "./configuredChains.yaml"
+	}
+	if _, err := os.Stat("../configuredChains.yaml"); err == nil {
```
mark-rushakoff, line 60: I'm not so sure about this. I would typically be very surprised about defaulting to a file in a single parent directory.

chainfactory.go
```diff
@@ -42,16 +45,40 @@ type BuiltinChainFactory struct {
...
+	}
+
+	builtinChainConfigs := make(map[string]ibc.ChainConfig)
+
+	var dat []byte
+	var err error
+	if configuredChainsFile != "" {
+		dat, err = os.ReadFile(configuredChainsFile)
+		if err != nil {
+			return nil, fmt.Errorf("error reading configuredChains.yaml: %v", err)
```
mark-rushakoff, line 71: This should use the actual path of the file.

chainspec.go
```diff
@@ -75,6 +75,11 @@ func (s *ChainSpec) Config() (*ibc.ChainConfig, error) {
 		return s.applyConfigOverrides(s.ChainConfig)
 	}
 
+	builtinChainConfigs, err := initBuiltinChainConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get preconfigured chains")
```
mark-rushakoff, line 80: Wrap the actual returned error.

configuredChains.yaml
```diff
@@ -0,0 +1,103 @@
+## NOTICE: This file gets embedded into ibctest binary
```
mark-rushakoff, line 1: Should this file be generated from another source, or do we have to hardcode its contents?

ibc/types.go
```diff
@@ -9,27 +9,27 @@ import (
...
+	ChainID string `yaml:"chain-id"`
 	// Docker images required for running chain nodes.
-	Images []DockerImage
+	Images []DockerImage `yaml:"images"`
 	// Binary to execute for the chain node daemon.
-	Bin string
+	Bin string `yaml:"bin"`
 	// Bech32 prefix for chain addresses, e.g. cosmos.
-	Bech32Prefix string
+	Bech32Prefix string `yaml:"bech32Prefix"`
```
mark-rushakoff, line 22: All the other names are hyphenated except this one. That is confusing.

mark-rushakoff (commented):
There are a couple things I'm not sure about.

chainfactory.go
boojamya: As discussed on our call. Going to prioritize the embedded `configuredChains.yaml` unless an environment variable is present.

configuredChains.yaml
boojamya: For lets "hardcode" this. But the idea around this PR is to make this info more transparent and easily editable.

mark-rushakoff (approved):
Rename the `once` variable, but otherwise this looks good to go.

PR #295, opened 2022-09-16
Updates for rly v2.1.0

agouin:
- Link paths in parallel to speed up tests now possible with https://github.com/cosmos/relayer/pull/987. Tests will pass once we update the go relayer `DefaultContainerVersion` to a release version that has this included.
- Faucet reqs in parallel to speed up tests
- Multiple paths same process
- Use native denom instead of `stake` for delegation. Increase faucet amount. 
- Pass bech flag to KeyBech32.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -886,7 +886,7 @@ func (tn *ChainNode) InitValidatorGenTx(
 	if err := tn.CreateKey(ctx, valKey); err != nil {
 		return err
 	}
-	bech32, err := tn.KeyBech32(ctx, valKey)
+	bech32, err := tn.KeyBech32(ctx, valKey, "")
 	if err != nil {
 		return err
 	}
@@ -925,10 +925,14 @@ func (tn *ChainNode) NodeID(ctx context.Context) (string, error) {
 }
 
 // KeyBech32 retrieves the named key's address in bech32 format from the node.
-func (tn *ChainNode) KeyBech32(ctx context.Context, name string) (string, error) {
+func (tn *ChainNode) KeyBech32(ctx context.Context, name string, bech string) (string, error) {
 	command := []string{tn.Chain.Config().Bin, "keys", "show", "--address", name,
 		"--home", tn.HomeDir(),
 		"--keyring-backend", keyring.BackendTest,
+	}
+
+	if bech != "" {
+		command = append(command, "--bech", bech)
 	}
 
 	stdout, stderr, err := tn.Exec(ctx, command, nil)
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -886,7 +886,7 @@ func (tn *ChainNode) InitValidatorGenTx(
 	if err := tn.CreateKey(ctx, valKey); err != nil {
 		return err
 	}
-	bech32, err := tn.KeyBech32(ctx, valKey, "")
+	bech32, err := tn.AccountKeyBech32(ctx, valKey)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


chain/cosmos/chain_node.go
```diff
@@ -925,12 +925,16 @@ func (tn *ChainNode) NodeID(ctx context.Context) (string, error) {
 }
 
 // KeyBech32 retrieves the named key's address in bech32 format from the node.
-func (tn *ChainNode) KeyBech32(ctx context.Context, name string) (string, error) {
+func (tn *ChainNode) KeyBech32(ctx context.Context, name string, bech string) (string, error) {
```
DavidNix, line 928: nit:

It would be helpful in the doc comment to know what values `bech` can be. If they are part of a known, finite list, an enum may be more suitable. 

Also, it would be helpful to mention a blank string for `bech` is acceptable.

chain/cosmos/cosmos_chain.go
```diff
@@ -213,7 +214,7 @@ func (c *CosmosChain) RecoverKey(ctx context.Context, keyName, mnemonic string)
 
 // Implements Chain interface
 func (c *CosmosChain) GetAddress(ctx context.Context, keyName string) ([]byte, error) {
-	b32Addr, err := c.getFullNode().KeyBech32(ctx, keyName)
+	b32Addr, err := c.getFullNode().KeyBech32(ctx, keyName, "")
```
DavidNix, line 217: nit: 

Having to pass a blank string is not common in Go APIs.

I've seen this cleaned up via 2 functions/methods. One could be something like `StandardKeyBech32` which only takes ctx and keyName. The other would be the `KeyBech32` and have the 3 arguments.

examples/packet_forward_test.go
```diff
@@ -28,16 +28,19 @@ func TestPacketForwardMiddleware(t *testing.T) {
 
 	cf := ibctest.NewBuiltinChainFactory(zaptest.NewLogger(t), []*ibctest.ChainSpec{
 		{Name: "gaia", ChainName: "gaia-fork", Version: "bugfix-replace_default_transfer_with_router_module"},
-		{Name: "osmosis", ChainName: "osmosis", Version: "v7.3.0"},
-		{Name: "juno", ChainName: "juno", Version: "v6.0.0"},
+		{Name: "osmosis", ChainName: "osmosis", Version: "v11.0.1"},
+		{Name: "juno", ChainName: "juno", Version: "v9.0.0"},
```
DavidNix, line 32: Thanks for upgrading these. This would be something I'd probably put in a separate PR, though. That way via `git bisect` or similar, if this change were to introduce flaky/failing tests, we can zero in on the cause quicker. 

No changes needed in the PR. Just an FYI.

DavidNix (approved):
You mention "speeding up tests" in the PR description. I'm curious what gains we saw locally and on CI?

More concurrency is not always equal to more speed.

agouin:
> You mention "speeding up tests" in the PR description. I'm curious what gains we saw locally and on CI?
More concurrency is not always equal to more speed.

For tests that need to link multiple paths before the test can proceed the savings are significant. For example, the packet forward middleware test and the Stride ICA/ICQ test.

Before (sequential synchronous path links and faucet funding):
Screenshot from 2022-09-19 09-43-11

After (parallel path links and faucet funding):
Screenshot from 2022-09-19 09-37-48

This example shows test startup overhead decreasing from ~60 blocks down to ~40 blocks.

chain/cosmos/cosmos_chain.go
agouin: Good idea, will update

chain/cosmos/chain_node.go
agouin: done

examples/packet_forward_test.go
agouin: Understood. I updated these while working through fixing the packet forward middleware test for the events processor and decided to keep them in since these versions are isolated to that test.

PR #302, opened 2022-09-22
Break out modules for isolated concerns and add separate CI jobs

agouin:
Makes new modules:
- `cmd/ibctest` for the CLI and runs the conformance tests in their own CI job
- `examples/cosmos` for the cosmos specific tests and runs them in their own CI job
- `examples/ibc` for the ibc specific tests and runs them in their own CI job
- `examples/penumbra` for the penumbra test and does not add a CI job
- `examples/polkadot` for the polkadot composable test and does not add a CI job


DavidNix:
> We don't need to run the example tests in CI

I respectfully disagree on that point. By running these tests on CI, we know the examples are accurate and work. I often point others to the examples as a starting point and I would hate to point them to code that doesn't work or works but fails the tests.

boojamya:
@DavidNix you make a valid point.

I think the reasoning behind this was that the CI tests keep failing because many of these "example" tests are writing to the same sqlite DB which is causing too much strain on the runner. However, clearly (at least at the time of writing this) tests are still failing.

Not sure if this is the culprit, but it could be one of a few culprits.

Maybe instead of making this its own module, we choose to **not** write to a DB and simply comment out all the `BlockDatabaseFile: ibctest.DefaultBlockDatabaseFilepath()`'s when the interchain is built.

agouin:
> By running these tests on CI, we know the examples are accurate and work. I often point others to the examples as a starting point and I would hate to point them to code that doesn't work or works but fails the tests.

This is valid, but maybe they don't belong in this repo due to the strain on CI. Perhaps an `ibctest-examples` repo.

DavidNix:
> move the cosmos specific tests into their own module/repo.

I'd be in favor of this compromise. 

I'm not sure the root problem we're trying to solve here? Flaky or frequently failing tests?

DavidNix:
> @DavidNix you make a valid point.
> 
> I think the reasoning behind this was that the CI tests keep failing because many of these "example" tests are writing to the same sqlite DB which is causing too much strain on the runner. However, clearly (at least at the time of writing this) tests are still failing.
> 
> Not sure if this is the culprit, but it could be one of a few culprits.
> 
> Maybe instead of making this its own module, we choose to **not** write to a DB and simply comment out all the `BlockDatabaseFile: ibctest.DefaultBlockDatabaseFilepath()`'s when the interchain is built. ... at least for these example tests.

I'm thinking the same: Writing to no db or a different one may help alleviate this problem for the examples, if that is indeed the root cause.

mark-rushakoff:
David is correct: the point of including ibctest examples is to provide sample code that compiles and passes its tests.

If the example tests are flaky, then ibctest isn't working as intended. If the example tests are too taxing on CI, then we need to optimize/minimize the resources required, or otherwise we are justified in increasing the CI runner's hardware.

Moving examples outside of ibctest's CI will detract from their purpose. The other repository will need to be maintained/updated as ibctest evolves; and a separate repository will not receive the same attention as the primary ibctest repo.

agouin:
> I'm not sure the root problem we're trying to solve here? Flaky or frequently failing tests?

> If the example tests are flaky, then ibctest isn't working as intended. If the example tests are too taxing on CI, then we need to optimize/minimize the resources required, or otherwise we are justified in increasing the CI runner's hardware.

It's not necessarily that the example tests are flaky, but when all the tests run in parallel and they all use the `ibctest.DefaultBlockDatabaseFilepath()`, the sqlite database frequently has issues with the migrations running while the other tests are trying to use the DB. 

Maybe to keep the examples in repo and have reliable CI we should de-parallelize the tests and run them in separate CI tasks, or have a way for the `ibctest.DefaultBlockDatabaseFilepath()` to return a random test path only for the parallel CI runs, since it is still nice to see everything in the same database when running tests locally.

>  FWIW, maintaining another repo is a possible solution, but that seems like high effort/low value right now.

> Moving examples outside of ibctest's CI will detract from their purpose. The other repository will need to be maintained/updated as ibctest evolves; and a separate repository will not receive the same attention as the primary ibctest repo.

I agree on this, a separate repo will mean maintenance overhead and likely it will not receive the desired amount of maintenance.

I think I will modify this to move the Cosmos specific tests to their own module, and have that run as a separate CI task so that we are still maintaining them through CI.

DavidNix:
> It's not necessarily that the example tests are flaky, but when all the tests run in parallel and they all use the `ibctest.DefaultBlockDatabaseFilepath()`, the sqlite database frequently has issues with the migrations running while the other tests are trying to use the DB.

Ah, thanks for that RCA. This makes sense because the sql migrations are very MVP right now. Before we merge this, let me take a shot at fixing the migrations. Ideally, they should run as a single transaction (they don't right now), so this could cause the raciness. IIRC, the original design was to use the db only for conformance tests, but as we've seen we need the information the database captures in all tests.

agouin:
> Before we merge this, let me take a shot at fixing the migrations.

That will be very helpful, and yes the db is extremely useful for just about every test.

I still think we could benefit from breaking apart the CI jobs, something like:
- ibctest unit tests (don't need self-hosted runner)
- ibctest conformance (needs self-hosted runner)
- ibctest examples (likely needs self-hosted runner)
- ibctest cosmos examples (may not need self-hosted runner)

I think we will see a good speed up in CI from doing this as long as we have self-hosted runner capacity. We can probably scale down our current runner, spin up a second or maybe even third runner, and run these separate jobs to speed up CI and get better insight into flaky tests

DavidNix:
> Before we merge this, let me take a shot at fixing the migrations. Ideally, they should run as a single transaction (they don't right now)

I haven't seen this code for a while; I'm wrong about the above. Mark wisely refactored the migrations to use a single tx already. I still have other ideas to fix the root cause though. Hope to have a PR up today.

.github/workflows/tests.yml
```diff
@@ -24,19 +24,74 @@ jobs:
...
+      # run tests
+      - name: run unit tests
+        run: go test -race -timeout 30m -v -p 2 ./...
+  test-cmd-ibctest:
+    name: test-cmd-ibctest
+    runs-on: [self-hosted, linux]
+    steps:
+      # Install and setup go
+      - name: Set up Go 1.18
+        uses: actions/setup-go@v2
```
DavidNix, line 41: There's a setup-go@v3 now. We probably want to use the latest while we're modifying these files. https://github.com/actions/setup-go

.github/workflows/tests.yml
```diff
@@ -24,19 +24,74 @@ jobs:
...
+        with:
+          go-version: 1.18
+
+      - name: checkout ibctest
+        uses: actions/checkout@v2
```
DavidNix, line 46: There's also a checkout@v3 now: https://github.com/actions/checkout

agouin: done

.github/workflows/tests.yml
agouin: done

DavidNix (approved):
🎉

PR #306, opened 2022-09-25
packet-forward-middleware packet memo, retry on timeout, and atomic forwards

agouin:
Modifies the packet forward middleware test to use the packet-forward-middleware with retry-on-timeout, packet memo refactor, and refund-through-ack. Updated to include the following cases:
- multi-hop A->B->C->D happy path test with A native denom
- denom unwind D->C->B->A happy path test, results in A native denom on chain A.
- forward A->B->C with failed recv_packet on C, then refund with ack error proxy B->A.
- forward A->B->C with forced timeout on B->C, configured for 2 retries, refund with ack error B->A after max retries.
- multi-hop A->B->C->D with failed recv_packet on D, then refund with ack error proxied C->B and B->A.
- typical ibc transfer from B->A (to create IBC denom), then multihop A->B->C->D with failed recv_packet on D, then refund with ack error proxy C->B and B->A. On chain C, C->D escrow funds are burned. On chain B, since it is the native denom, escrow funds are transferred from the B->C escrow account to the B->A escrow account instead of burned.

https://github.com/strangelove-ventures/packet-forward-middleware/pull/36
https://github.com/strangelove-ventures/packet-forward-middleware/pull/33

image


examples/ibc/packet_forward_test.go
```diff
@@ -102,9 +102,6 @@ func TestPacketForwardMiddleware(t *testing.T) {
 	// Get original account balances
 	osmosisUser, gaiaUser, junoUser := users[0], users[1], users[2]
 
-	osmosisBalOG, err := osmosis.GetBalance(ctx, osmosisUser.Bech32Address(osmosis.Config().Bech32Prefix), osmosis.Config().Denom)
-	require.NoError(t, err)
-
 	// Send packet from Osmosis->Hub->Juno
 	// receiver format: {intermediate_refund_address}|{foward_port}/{forward_channel}:{final_destination_address}
```
akure, line 106: foward_port -> "forward_port"

chain/cosmos/chain_node.go
```diff
@@ -225,14 +225,28 @@ func (tn *ChainNode) SetTestConfig(ctx context.Context) error {
...
+	a := make(testutil.Toml)
+	a["minimum-gas-prices"] = tn.Chain.Config().GasPrices
+	return testutil.ModifyTomlConfigFile(
+		ctx,
+		tn.logger(),
+		tn.DockerClient,
+		tn.TestName,
+		tn.VolumeName,
+		"config/app.toml",
+		a,
```
jtieri, line 249: This isn't related to the forward middleware, but is needed to spin up a Gaia v8 chain due to changes in the SDK?

chain/cosmos/poll.go
```diff
@@ -58,3 +59,24 @@ func PollForMessage[T any](ctx context.Context, chain *CosmosChain, registry cod
 	bp := testutil.BlockPoller[T]{CurrentHeight: chain.Height, PollFunc: doPoll}
 	return bp.DoPoll(ctx, startHeight, maxHeight)
 }
+
+// PollForBalance polls until the balance matches
+func PollForBalance(ctx context.Context, chain *CosmosChain, deltaBlocks uint64, balance ibc.WalletAmount) error {
```
jtieri, line 64: This is really helpful!! I'm all for moving to a poll based approach throughout `ibctest` vs. waiting arbitrary block heights and hoping the timing works out.

ibc/relayer.go
```diff
@@ -95,6 +98,78 @@ type Relayer interface {
...
+	var srcConnectionID string
+	for _, connection := range srcConnections {
+		if connection.ClientID == srcClientID {
+			if srcConnectionID != "" {
+				return nil, fmt.Errorf("found multiple connections on %s for client %s", srcChainID, srcClientID)
+			}
+			srcConnectionID = connection.ID
+		}
+	}
+
```
jtieri, line 146: Shouldn't we also check if `srcConnectionID == nil` and return an error like we do for the client ID on L124?

examples/ibc/packet_forward_test.go
```diff
@@ -26,43 +44,53 @@ func TestPacketForwardMiddleware(t *testing.T) {
...
-	gaia, osmosis, juno := chains[0], chains[1], chains[2]
+	chainA, chainB, chainC, chainD := chains[0].(*cosmos.CosmosChain), chains[1].(*cosmos.CosmosChain), chains[2].(*cosmos.CosmosChain), chains[3].(*cosmos.CosmosChain)
 
 	r := ibctest.NewBuiltinRelayerFactory(
 		ibc.CosmosRly,
 		zaptest.NewLogger(t),
-	).Build(
-		t, client, network,
-	)
+		relayer.CustomDockerImage("ghcr.io/cosmos/relayer", "andrew-recv_packet_write_ack_separation", rly.RlyDefaultUidGid),
```
jtieri, line 64: I think we merged the upstream changes in the relayer so this could get changed back to main before merging.

jtieri (approved):
This is awesome, great work in adapting for the async acks and catching the issue with the escrow accounts 🔥

examples/ibc/packet_forward_test.go
agouin: The default ibctest rly is `v2.1.2`, which we should update soon to `v2.2.0-rc1`+. We will still need an override until we change that, but we I'll update this to `main` for now.

ibc/relayer.go
agouin: yes, good catch

chain/cosmos/chain_node.go
```diff
@@ -569,7 +583,14 @@ type CosmosTx struct {
...
 
-func (tn *ChainNode) SendIBCTransfer(ctx context.Context, channelID string, keyName string, amount ibc.WalletAmount, timeout *ibc.IBCTimeout) (string, error) {
+func (tn *ChainNode) SendIBCTransfer(
+	ctx context.Context,
+	channelID string,
+	keyName string,
+	amount ibc.WalletAmount,
+	timeout *ibc.IBCTimeout,
+	memo string,
+) (string, error) {
```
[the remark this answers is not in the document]
agouin: You're right, this is a good thing to tidy up. Happy to make the change now.

PR #317, opened 2022-10-04
Add test case for Interchain Accounts

### LINKED ISSUE #13: Add tests for interchain accounts

jtieri:
## Description

This PR adds an example test case to `examples/ibc` that simulates a basic flow with regards to Interchain Accounts.

The test case walks through the following:
- Creating a new ICA path between a Host and Controller chain
- Registering a new ICA on the host chain, on behalf of a user account on the controller chain
- Querying for the ICA on the host chain
- Sending a bank transfer message from the controller chain to the ICA on the host chain
- Letting a packet timeout and ensuring the channel is closed on packet timeout
- Reinitializing the channel following a channel closing

## Checklist

## Notes

This requires bumping the ibc-go version to v5 stable

Closes #13


examples/ibc/interchain_accounts_test.go
```diff
@@ -0,0 +1,314 @@
...
+			Relayer: r,
+			Path:    pathName,
+		})
+
+	require.NoError(t, ic.Build(ctx, eRep, ibctest.InterchainBuildOptions{
+		TestName:          t.Name(),
+		Client:            client,
+		NetworkID:         network,
+		SkipPathCreation:  true,
+		BlockDatabaseFile: ibctest.DefaultBlockDatabaseFilepath(),
```
boojamya, line 85: Is there a need to write to a db here? Maybe we can remove this line?

boojamya (approved):
🍾

chain/cosmos/codec.go
```diff
@@ -9,8 +9,10 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	authTx "github.com/cosmos/cosmos-sdk/x/auth/tx"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	icaTypes "github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/types"
```
agouin, line 12: import aliases should be all lowercase

chain/cosmos/codec.go
```diff
@@ -26,6 +28,9 @@ func DefaultEncoding() simappparams.EncodingConfig {
 	ibctypes.RegisterInterfaces(cfg.InterfaceRegistry)
 	transfertypes.RegisterInterfaces(cfg.InterfaceRegistry)
 
+	demoTypes.RegisterInterfaces(cfg.InterfaceRegistry)
```
agouin, line 31: Do you think we need these and the ica types in the default encoding? Or can we put them in an override encoding for the ICA test? For example with osmosis: 

https://github.com/strangelove-ventures/ibctest/blob/v3/examples/osmosis/osmosis.go#L10-L17

https://github.com/strangelove-ventures/ibctest/blob/f2a589c02156f2bf47382e670fde3264f024749e/examples/osmosis/cosmos_chain_expedited_proposal_test.go#L39

examples/ibc/interchain_accounts_test.go
```diff
@@ -0,0 +1,314 @@
...
+	}))
+
+	// Fund a user account on chain1 and chain2
+	const userFunds = int64(10_000_000_000)
+	users := ibctest.GetAndFundTestUsers(t, ctx, t.Name(), userFunds, chain1, chain2)
+	chain1User := users[0]
+	chain2User := users[1]
+
+	// Generate a new IBC path
+	err = r.GeneratePath(ctx, eRep, chain1.Config().ChainID, chain2.Config().ChainID, pathName)
```
agouin, line 95: Is there a reason to do the path client/conns/channels steps explicitly in the test rather than letting the `Interchain` `Build` do it?

agouin:
Woohoo ICA tests :fire: :robot:

chain/cosmos/codec.go
jtieri: Good call

examples/ibc/interchain_accounts_test.go
jtieri: I was skipping the automatic path initialization because we need to explicitly create an `ORDERED` channel. Also, in ICA the channel handshake is initiated by the user via the `intertx register` command which fires off `MsgRegisterInterchainAccount`

PR #324, opened 2022-10-12
expose cliContext

boojamya:
This needs to be exposed in order to query the state ("Frozen", "Active", "Expired") of a client. 

The cosmos/relayer will use this in its testing suite.


DavidNix:
Closing because I was able to build a client.Context in the relayer.

PR #325, opened 2022-10-14
feat: Add cosmos message poller

DavidNix:
Adds a new poll function for cosmos chains. The ability to poll for any arbitrary cosmos message. 

I will use this in the relayer codebase to help with a test regarding keeping light clients up to date.

### DIFF
--- a/examples/cosmos/versions_test.go
+++ b/examples/cosmos/versions_test.go
@@ -1,0 +1,6 @@
+package cosmos_test
+
+const (
+	gaiaVersion    = "v7.1.0"
+	osmosisVersion = "v12.2.0"
+)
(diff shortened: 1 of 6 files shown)


DavidNix:
@agouin Thanks for the quick review! But apologies, this was still in draft. I'm trying to add another poller function which will leverage this generic refactor. That's great all tests pass though.

examples/cosmos/versions_test.go
```diff
@@ -0,0 +1,6 @@
+package cosmos_test
+
+const (
```
agouin, line 3: :clap:

DavidNix:
Addresses https://github.com/cosmos/relayer/issues/853

PR #326, opened 2022-10-14
fix: Unable to build ibctest binary

DavidNix:
With a `go.mod` in dir `cmd/ibctest`, we are unable to build the `ibctest` artifact:

```
❯ make ibctest
go generate ./...
go test -ldflags "-X github.com/strangelove-ventures/ibctest/v6/internal/version.GitSha=cdb1669" -c -o ./bin/ibctest ./cmd/ibctest
main module (github.com/strangelove-ventures/ibctest/v6) does not contain package github.com/strangelove-ventures/ibctest/v6/cmd/ibctest
make: *** [Makefile:9: ibctest] Error 1
```

It's also uncommon to have a nested `go.mod` within the `cmd` directory. 

I was able to preserve the test partitioning and eliminate all the extra nested `go.mod`. Simply run the tests in a specific dir. (i.e. don't do `./...`).

## Use testing.Short()

I've taken the strategy that all e2e or integration tests must skip the test if `testing.Short()` is true. To me, this makes sense because conformance and other e2e tests may take a long time, > 1 minute. They are anything but short.

This will take developer discipline but the unit tests via `make test` will timeout after 60s. Therefore, that should catch where we miss a `testing.Short()` check.


Makefile
```diff
@@ -10,7 +10,7 @@ ibctest: gen ## Build ibctest binary into ./bin
 
 .PHONY: test
 test: ## Run unit tests
-	@go test -cover -short -race -timeout=60s $(shell go list ./... | grep -v /cmd/)
+	@go test -cover -short -race -timeout=60s ./...
```
DavidNix, line 13: `testing.Short()` and `-short` flag means we can run unit tests throughout the whole project. No need to filter out packages.

.github/workflows/tests.yml
```diff
@@ -31,9 +31,10 @@ jobs:
 
       # run tests
       - name: run unit tests
-        run: go test -race -timeout 30m -v -p 2 ./...
-  test-cmd-ibctest:
-    name: test-cmd-ibctest
+        # -short flag purposefully omitted because there are some longer unit tests
+        run: go test -race -timeout 10m -p 2 $(go list ./... | grep -v /cmd | grep -v /examples)
```
jtieri, line 35: It seems like we are expecting some longer running tests here but the timeout value changes from `30m` to `10m`, does that make sense or should we use the larger value still?

jtieri (approved):
This looks good to me 👍

PR #329, opened 2022-11-01
Get PenumbraChainStart test passing

boojamya:
The PR gets the `TestPenumbraChainStart` passing. 

Due to the command structure being re-organized in the `pcli` binary, some functions needed to be update.

Notes:
- only touches the functions necessary to spin up the chain. Other commands such as creating keys still need to be updated.
- helps with backwards compatibility when working with Tendermint Nodes. 
- downgrades user from `root` to `heighliner` user 
   - This will need to be merged to work on arm machines: https://github.com/strangelove-ventures/heighliner/pull/59)


chain/penumbra/penumbra_app_node.go
```diff
@@ -204,7 +205,7 @@ func (p *PenumbraAppNode) CreateNodeContainer(ctx context.Context) error {
 			Cmd:        cmd,
 
 			Hostname: p.HostName(),
-			User:     dockerutil.GetRootUserString(),
+			User:     dockerutil.GetHeighlinerUserString(),
```
agouin, line 208: this can probably use `p.Image.UidGid` now

chain/penumbra/penumbra_app_node.go
```diff
@@ -258,7 +259,7 @@ func (p *PenumbraAppNode) Exec(ctx context.Context, cmd []string, env []string)
 	opts := dockerutil.ContainerOptions{
 		Binds: p.Bind(),
 		Env:   env,
-		User:  dockerutil.GetRootUserString(),
+		User:  dockerutil.GetHeighlinerUserString(),
```
agouin, line 262: `p.Image.UidGid` here also so that non-heighliner images can also be used

chain/penumbra/penumbra_chain.go
```diff
@@ -428,7 +428,7 @@ func (c *PenumbraChain) start(ctx context.Context) error {
 		eg.Go(func() error {
 			return n.TendermintNode.CreateNodeContainer(
 				egCtx,
-				fmt.Sprintf("--proxy-app=tcp://%s:26658", n.PenumbraAppNode.HostName()),
+				fmt.Sprintf("--proxy_app=tcp://%s:26658", n.PenumbraAppNode.HostName()),
```
agouin, line 431: does this also need the conditional separator like you did for the config?

chain/internal/tendermint/tendermint_node.go
```diff
@@ -173,6 +177,26 @@ func (tn *TendermintNode) SetConfigAndPeers(ctx context.Context, peers string) e
 	)
 }
 
+// Tenderment deprecate snake_case in config for hyphen-case in v0.34.1
+// https://github.com/tendermint/tendermint/blob/main/CHANGELOG.md#v0341
+func (tn *TendermintNode) GetConfigSeparator() (string, error) {
```
agouin, line 182: This is fancy :star2:

chain/penumbra/penumbra_chain.go
boojamya: Yes! indeed it does. Fixed.

jackzampolin (approved):

PR #336, opened 2022-11-03
Make configutil public

### LINKED ISSUE #335: Confusion around ConfigFileOverrides
Hi, I am attempting to use the [ConfigFileOverrides](https://github.com/strangelove-ventures/ibctest/blob/26fb2566fcd5a998abab7ac4d5215e468f5334c9/ibc/types.go#L35) type to update some `toml` files in the tests. 

However, it appears that the `map[string]any` value [gets cast to an internal type](https://github.com/strangelove-ventures/ibctest/blob/26fb2566fcd5a998abab7ac4d5215e468f5334c9/chain/cosmos/cosmos_chain.go#L529) and so when the library is imported I am unable to use [this type alias](https://github.com/strangelove-ventures/ibctest/blob/26fb2566fcd5a998abab7ac4d5215e468f5334c9/internal/configutil/toml.go#L16) directly.

If I try to use a `map[string]any`, I hit the following error `failed to start chains: failed to start chain simapp-b: Provided toml override for file config/app.toml is of type (map[string]interface {}). Expected (DecodedToml)`

agouin:
Resolves #335


DavidNix (approved):
I'm not a fan of `github.com/strangelove-ventures/ibctest/v6/test` package. It was a short-term solution to break an import cycle. It's poorly named imo. The functionality in there would be better to be part of `ibctest` root package, but the import cycle prevents it.

In any case, I may opt for `github.com/strangelove-ventures/ibctest/v6/configutil` vs. `github.com/strangelove-ventures/ibctest/v6/test/configutil`.

agouin:
Played around with a few re-structuring ideas for the utils and the `test` rename. I'm not a huge fan of the utils being a bunch of dirs in the root dir, so landed on this. Thoughts?

util/util.go
```diff
@@ -0,0 +1 @@
+package util
```
DavidNix, line 1: It's customary to include a `doc.go` briefly stating the purpose of the package and any sub packages.

DavidNix (commented):
Thanks for the updates and apologies for my belated review. 

The Go developer in me really doesn't want a `util` package name even if it's just an "umbrella" package. It's incredibly rare in Go and each time I see one in a Go codebase, I immediately think "These authors aren't professional Go developers."  

However, that's just my perspective. 

I'd be in favor of the simpler getting rid of the top-level `util` package anyway. 

Just have `ibctest/chain` and `ibctest/config` packages. 

If there's a good reason for a package wrapping `chain` and `config`, then perhaps `testutil` is a better name.

agouin:
> perhaps `testutil` is a better name.

I like this idea. Seems like both util packages can be merged into this single package so that there is no umbrella package and then we don't need 2 root dirs for utils.

DavidNix:
> > perhaps `testutil` is a better name.
> 
> I like this idea. Seems like both util packages can be merged into this single package so that there is no umbrella package and then we don't need 2 root dirs for utils.

Works for me! Thanks for living with the churn on this one.

PR #338, opened 2022-11-03
Integrate coin-type configuration

boojamya:
The PR adds the ability to specify Coin Types when creating or restoring keys for chains. 

The coinType is invoked in the chainConfig but defaults to "118" (Cosmos SDK default) if the key:value pair is left blank.

### DIFF
--- a/ibc/types.go
+++ b/ibc/types.go
@@ -26,7 +26,7 @@ type ChainConfig struct {
 	// Denomination of native currency, e.g. uatom.
 	Denom string `yaml:"denom"`
 	// Coin type
-	CoinType uint32 `default:"118" yaml:"coin-type"`
+	CoinType string `default:"118" yaml:"coin-type"`
 	// Minimum gas prices for sending transactions, in native currency denom.
 	GasPrices string `yaml:"gas-prices"`
 	// Adjustment multiplier for gas fees.
@@ -52,18 +52,22 @@ func (c ChainConfig) Clone() ChainConfig {
 }
 
 func (c ChainConfig) VerifyCoinType() (uint32, error) {
-	if c.CoinType == 0 {
+	if c.CoinType == "" {
 		typ := reflect.TypeOf(c)
 		f, _ := typ.FieldByName("CoinType")
 		coinTypeS := f.Tag.Get("default")
-		u64, err := strconv.ParseUint(coinTypeS, 0, 32)
+		u64, err := strconv.ParseUint(coinTypeS, 10, 32)
 		if err != nil {
 			return 0, err
 		}
 		u32 := uint32(u64)
 		return u32, nil
 	} else {
-		return c.CoinType, nil
+		u64, err := strconv.ParseUint(c.CoinType, 10, 32)
+		if err != nil {
+			return 0, err
+		}
+		return uint32(u64), nil
 	}
 
 }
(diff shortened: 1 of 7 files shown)


interchain.go
```diff
@@ -442,9 +444,11 @@ type relayerChain struct {
 // BuildWallet will generate a random key for the key name in the provided keyring.
 // Returns the mnemonic and address in the bech32 format of the provided ChainConfig.
 func BuildWallet(kr keyring.Keyring, keyName string, config ibc.ChainConfig) ibc.Wallet {
-	// NOTE: this is hardcoded to the cosmos coin type.
-	// In the future, we may need to get the coin type from the chain config.
-	const coinType = types.CoinType
+	coinTypeU64, err := strconv.ParseUint(config.CoinType, 10, 32)
```
jtieri, line 447: The variable is named `coinTypeU64` but you are telling `ParseUInt` to use 32bits, and then on L451 there's a typecast to `uint32`. It looks like the coin type is a `uint32` so I'd say this parse call is fine but the type cast could be eliminated. 

Also the name could be shortened to just `coinType` for brevity

ibc/types.go
```diff
@@ -46,6 +51,27 @@ func (c ChainConfig) Clone() ChainConfig {
 	return x
 }
 
+func (c ChainConfig) VerifyCoinType() (string, error) {
+	// If coin-type is left blank in the ChainConfig,
+	// the Cosmos SDK default of 118 is used.
+	if c.CoinType == "" {
+		typ := reflect.TypeOf(c)
+		f, _ := typ.FieldByName("CoinType")
+		coinType := f.Tag.Get("default")
```
jtieri, line 60: This feels odd in the sense that I would expect the struct's field to take on the default value if the empty string is passed in. Idk that I've personally used the `default` tag before so maybe this is the idiomatic way to go about this?

Thoughts? cc @DavidNix @agouin

boojamya: Yes, I thought so too! After testing, if the value is empty, it takes on "0" which is Bitcoin's coin-type.
If there is a better way to go about this, I'm happy to learn.

relayer/rly/cosmos_relayer.go
```diff
@@ -114,10 +115,10 @@ func (commander) AddChainConfiguration(containerFilePath, homeDir string) []stri
 	}
 }
 
-func (commander) AddKey(chainID, keyName, homeDir string) []string {
+func (commander) AddKey(chainID, keyName, coinType, homeDir string) []string {
 	return []string{
 		"rly", "keys", "add", chainID, keyName,
-		"--home", homeDir,
+		"--coin-type", fmt.Sprint(coinType), "--home", homeDir,
```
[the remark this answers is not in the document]
boojamya: Thanks for this call out, it's not.

interchain.go
boojamya: `ParseUInt` outputs a `uint64` whether you specify 32bits or 64bits.
Regardless, for readability, I like what you're saying here and can do the type cast in the `hd.CreateHdPath` function (L456) with the renaming suggestion.

jtieri: Well TIL! Disregard my comment then 😛

ibc/types.go
jtieri: Seems very counterintuitive. I'll go ahead and approve, if there is a better way to do this we can always circle back up on it. Good job!

jtieri (approved):

PR #348, opened 2022-11-28
Icq wasm

KyleMoser:
ICQ WASM test case based on polymerdao's ibctest example (bo/icq_wasm branch)


chain/cosmos/cosmos_chain.go
```diff
@@ -335,7 +335,12 @@ func (c *CosmosChain) ExecuteContract(ctx context.Context, keyName string, contr
 	return c.getFullNode().ExecuteContract(ctx, keyName, contractAddress, message)
 }
 
-// QueryContract performs a smart query, taking in a query struct and returning a error with the response struct populated. 
+// ExecuteContract executes a contract transaction with a message using it's address.
+func (c *CosmosChain) ExecuteContractWithResult(ctx context.Context, keyName string, contractAddress string, message string) (txHash string, err error) {
```
agouin, line 339: Thanks for taking into account breaking changes here, but I'd rather make a breaking change to ExecuteContract to return the txHash instead of introducing a duplicate method

examples/ibc/wasm/wasm_icq_test.go
```diff
@@ -0,0 +1,447 @@
...
+	Models []kvPair
+}
+
+type statusResults struct {
+	SyncInfo struct {
+		Height string `json:"latest_block_height"`
+	} `json:"SyncInfo"`
+}
+
+// type icqResults struct {
```
agouin, line 370: same here re: commented code

DavidNix (approved):

PR #349, opened 2022-12-01
Changes for slim rly images

agouin:
- Remove dependency on `chown` in rly image. Use existing dockerutil `NewFileWriter` instead. 
- Use relayer user home dir


DavidNix (approved):
Nice refactor!

PR #354, opened 2022-12-07
version bumps for ibctest v6

faddat:
This PR upgrades ibctest from SDK v0.46.2 tro v0.46.6 and patches it for Dragonberry while annotating go.mod


faddat:
sure thing :).  

I think I'll close for now, and may reopen later :)

PR #357, opened 2022-12-07
fix: remove redundant ValidateBasic check in BroadcastTx

charleenfei:
this check is redundant because ValidateBasic on the messages will be run by the chain binary anyway, running it prematurely leads to strange version issues 

context: writing an e2e test for upgrading ibc-go v6 -> v7 runs automatic migrations which bump the client and consensusstate versions of the tendermint client and solo machine, we would need the msg to be broadcasted and validated by the v6 binary rather than the v7 code.

### DIFF
--- a/chain/cosmos/broadcaster.go
+++ b/chain/cosmos/broadcaster.go
@@ -182,12 +182,6 @@ func (b *Broadcaster) defaultTxFactory(clientCtx client.Context, accountNumber u
 // BroadcastTx uses the provided Broadcaster to broadcast all the provided messages which will be signed
 // by the User provided. The sdk.TxResponse and an error are returned.
 func BroadcastTx(ctx context.Context, broadcaster *Broadcaster, broadcastingUser User, msgs ...sdk.Msg) (sdk.TxResponse, error) {
-	for _, msg := range msgs {
-		if err := msg.ValidateBasic(); err != nil {
-			return sdk.TxResponse{}, err
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


charleenfei:
looks like the test failure is coming from an rpc not found in the interchain accounts test, but something went wrong with the docker container not starting :)

DavidNix:
> looks like the test failure is coming from an rpc not found in the interchain accounts test, but something went wrong with the docker container not starting :)

I'm re-running the failed one. As common with e2e tests, there's flakiness (especially when dealing with networks).

charleenfei:
interesting, it looks like the same test is failing. so something on the interchain account creation is failing -- but i am not sure how my changes would have caused that since the BroadcastTx function doesnt seem to be called in the test flow that is failing

DavidNix (approved):

DavidNix:
> interesting, it looks like the same test is consistently failing, even on your main branch. it looks like the interchain account creation is failing, which is why the rpc query returns an error.
> 
> looks like this error started with this commit: [9359b06](https://github.com/strangelove-ventures/ibctest/commit/9359b061fbe59910c193673e834c1189e3c71043)

You're right. See https://github.com/strangelove-ventures/ibctest/pull/354#issuecomment-1343007663

charleenfei:
@DavidNix is it possible to merge this PR before the failing test is fixed?

DavidNix:
> @DavidNix is it possible to merge this PR before the failing test is fixed?

Sure. I normally dislike merging when CI is failing. But this seems like a valid exception.

PR #360, opened 2022-12-14
Fix BroadcastSync for cosmos broadcaster

chatton:
@agouin This PR should fix the `BroadcastSync` issue. We need to wait for some time for the tx to be included. Then we can query for it and return the fully populated response.

Relevant upgrade steps [here](https://github.com/cosmos/cosmos-sdk/blob/main/UPGRADING.md#broadcast-mode).


testutil/wait.go
```diff
@@ -112,3 +113,27 @@ func (h *height) update(height uint64) {
...
+// period, or fn returns an error, the condition will not have been met.
+func WaitForCondition(timeoutAfter, pollingInterval time.Duration, fn func() (bool, error)) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeoutAfter)
+	defer cancel()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("failed waiting for condition after %f seconds", timeoutAfter.Seconds())
+		case <-time.After(pollingInterval):
```
DavidNix, line 128: Fyi, there is a resource leak here if `<-ctx.Done()` happens first. 

Per https://pkg.go.dev/time#After

> The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

I'm not sure if efficiency is a concern here. But to be safe, I typically `NewTimer` and `defer timer.Stop()` to be safe.

PR #363, opened 2023-01-02
Polkadot's relay chain and parachain improvements (wallet abstraction, chain interface implementations)

misko9:
1. Polkadot chain working with interchain process
2. Relay and faucet accounts are now added at genesis for relay chain and parachains
3. Added GetBalance() for relay chain and parachain
4. For polkadot chain, implemented: CreateKey(), RecoverKey(), GetAddress()
5. Replace custom DeriveSr25519FromName() to use RecoverKey()
6. Change ibc.Wallet to an interface so that each chain can call their own  implementation.
7. Add example test for pushing wasm client contract and pulling the code from chain


agouin (approved):
:clap:

PR #366, opened 2023-01-07
bring in changes from main branch to v3

jtieri:
The v3 branch was missing a few changes from the main branch like the fix for building the binary, some changes to the poller, and the changes for the packet forward middleware. This PR also bumps the ibc-go version to v3.4.0


jtieri:
CI job for the example test cases was failing due to using the `latest` tag for the interchain accounts demo. We'll need to look into this and update the test case to work with the most recent changes but that is outside the scope of this PR.

agouin (approved):

PR #374, opened 2023-01-20
Update penumbra key management

boojamya:
Since penumbra doesn't have "key names", we'll have the folder that houses the key act as the key name.


agouin (approved):
Nice way to make this interface consistent

PR #381, opened 2023-01-27
PreGenesis hooks

agouin:
Adds `PreGenesis` hook to execute a function prior to genesis file creation. Adds `SkipGenTx` to skip validator gentx flow.

This is utilized currently in our consumer chain tests.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -801,12 +801,12 @@ func (tn *ChainNode) ExportState(ctx context.Context, height int64) (string, err
 	tn.lock.Lock()
 	defer tn.lock.Unlock()
 
-	_, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
+	stdout, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
 	if err != nil {
 		return "", err
 	}
-	// output comes to stderr for some reason
-	return string(stderr), nil
+	// output comes to stderr on older versions
+	return string(stdout) + string(stderr), nil
 }
 
 func (tn *ChainNode) UnsafeResetAll(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


chain/cosmos/chain_node.go
```diff
@@ -801,12 +801,12 @@ func (tn *ChainNode) ExportState(ctx context.Context, height int64) (string, err
...
 	defer tn.lock.Unlock()
 
-	_, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
+	stdout, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
 	if err != nil {
 		return "", err
 	}
-	// output comes to stderr for some reason
-	return string(stderr), nil
+	// output comes to stderr on older versions
```
jtieri, line 808: glad to see that was finally fixed!

ibc/types.go
```diff
@@ -35,6 +35,10 @@ type ChainConfig struct {
 	TrustingPeriod string `yaml:"trusting-period"`
 	// Do not use docker host mount.
 	NoHostMount bool `yaml:"no-host-mount"`
+	// When true, will skip validator gentx flow
+	SkipGenTx *bool
```
jtieri, line 39: What's the reasoning for making this a pointer? Looks like we are just checking if this is false and if so we run the usual gentx logic. Couldn't we just use the fact that the zero value is false and gentx will only be skipped if explicitly set to true?

jtieri (commented):
Just had the one question but otherwise this looks good!

ibc/types.go
agouin: The reason for the pointer is so that during `MergeChainSpecConfig` and `applyConfigOverrides`, a zero value `false` would not override a prior intentional `true`. Having `nil` as the zero value will confirm override intent when the user has explicitly set the value to a non-nil bool.

With a non-pointer, it would probably work well enough in those mentioned methods to only use the incoming value if it's `true`, but this would mean it would not be possible to override a `true` value with `false` (any `true` value throughout means it's `true`). This is good enough for the current use case of this though, and does simplify the usage when declaring `ChainConfig` instances.

jtieri (approved):

ibc/types.go
jtieri: thanks for the reasoning! makes sense to me.

PR #382, opened 2023-01-27
Feat/hyperspace

misko9:
## Hyperspace relayer ibctest interface checklist
- [x] Write chain configs to docker container
- [x] Write core config to docker container
- [x] Name()
- [x] DockerUser()
- [x] AddChainConfiguration()
- [x] AddKey() [Unused / don't implement]
- [x] CreateChannel()
- [x] CreateClients()
- [x] CreateConnections()
- [x] FlushAcknowledgements() [Unused / don't implement]
- [x] FlushPackets [Unused / don't implement]
- [x] GeneratePath()
- [x] UpdatePath() [Unused / don't implement]
- [x] GetChannels()
- [x] GetConnections()
- [x] GetClients()
- [x] LinkPath()
- [x] RestoreKey() [Implemented in Hyperspace Relayer]
- [x] StartRelayer()
- [x] UpdateClients [Unused / don't implement]
- [x] ConfigContent()
- [x] DefaultContainerImage()
- [x] DefaultContainerVersion()
- [x] ParseAddKeyOutput() [Unused / don't implement]
- [x] ParseRestoreKeyOutput()
- [x] ParseGetChannelsOutput()
- [x] ParseGetConnectionsOutput()
- [x] ParseGetClientsOutput()
- [x] Init()
- [x] CreateWallet()
- [x] Add contract hash to substrate/parachain config (actually cosmos chain config, but may change)
- [x] Generate correct priv/pub keys for cosmos chain
- [x] ParachainNode: GetIbcBalance(), i.e. QueryBalanceWithAddress should work
- [x] Make relayer wallet amounts configurable for different denoms
- [x] Make cosmos chain block time configurable so that relayer doesn't work harder than necessary

## Needed Hyperspace binary changes
- [x] Take mnemonic as input for both chain configs (account used for respective txs) (Only needed for Cosmos, currently using KeyEntry which should be sufficient
- [x] Add contract hash to substrate/parachain config for wasm messages
- [ ] Optionally, add Hyperspace binary cmd for query clients and connections (if we want to be similar to rly interface)
- [x] Support either chain in chain slot a/b
- [x] Query channels should not panic. We need to be able to get the channel for each config.

## e2e test (examples/hyperspace/hyperspace_test.go)
- [x] Add contract hash to Hyperspace's substrate/parachain config  (actually cosmos chain config, but may change)
- [x] Add create client
- [x] Add create connection
- [x] Add create channel
- [x] Start relayer
- [x] Send ibc tx from cosmos -> substrate
- [x] Send ibc tx from substrate -> cosmos


jonathanpberger:
Thank you @joeabbey for volunteering to code review :-D

anhductn2001:
> NB: @agouin sez "notional is handling the rebase to get this up to date w/ `main`."

Yea, @jonathanpberger. We are doing that.

chain/cosmos/08-wasm-types/client_state.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+func (c ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, msg exported.ClientMessage) bool {
+	return true
+}
+
+// UpdateStateOnMisbehaviour should perform appropriate state changes on a client state given that misbehaviour has been detected and verified
+func (c ClientState) UpdateStateOnMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, clientMsg exported.ClientMessage) {
+
```
joeabbey, line 90: Should we panic here, since there's no implementation?

chain/cosmos/08-wasm-types/consensus_state.go
```diff
@@ -0,0 +1,19 @@
+package types
+
+import (
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
+)
+
+var _ exported.ConsensusState = (*ConsensusState)(nil)
+
+func (m ConsensusState) ClientType() string {
+	return ""
```
joeabbey, line 10: Is the client type `""`?

misko9: Same as above, the implementation of these 08-wasm interfaces does not matter, we just want the codecs.

relayer/hyperspace/hyperspace_commander.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+func (c *hyperspaceCommander) AddChainConfiguration(containerFilePath, homeDir string) []string {
+	fmt.Println("[hyperspace] AddChainConfiguration ", containerFilePath, homeDir)
+	//c.chainConfigPaths = append(c.chainConfigPaths, containerFilePath)
+	return []string{
+		"hyperspace",
+		"-h",
+	}
+}
```
joeabbey, line 51: Does this add the configuration or print help?

joeabbey (commented):
Overall looks really good, adding a few nits and comments.

Also, I wonder if we need to further decompose the "Commander" interface to avoid:
`panic("do not call me")`

joeabbey (approved):

PR #384, opened 2023-01-29
Add support for genesis sub commands

gjermundgaraba:
I did this change as a hack for myself to unblock myself because I am using the new genesis sub command from the SDK (https://github.com/cosmos/cosmos-sdk/pull/14149).

For reference: https://github.com/strangelove-ventures/ibctest/pull/359#issuecomment-1407785995


chainspec.go
```diff
@@ -28,10 +28,11 @@ type ChainSpec struct {
...
 	Version string
 
-	// GasAdjustment and NoHostMount are pointers in ChainSpec
+	// GasAdjustment, NoHostMount and UsingNewGenesisCommand are pointers in ChainSpec
 	// so zero-overrides can be detected from omitted overrides.
-	GasAdjustment *float64
-	NoHostMount   *bool
+	GasAdjustment          *float64
+	NoHostMount            *bool
+	UsingNewGenesisCommand *bool
```
agouin, line 35: Since `ChainSpec` embeds `ChainConfig`, I don't think this is necessary on the `ChainSpec` itself.

gjermundgaraba: The reason I added it was what I assumed was the same reason GasAdjusment and NoHosMount were there (which are also in ChainConfig), to be able to have them nullable and thus overridable. But I have to admit I didn't quite understand the config structure here so I just mimicked how it was done for the above mentioned values.

gjermundgaraba: I updated it (and added a test). It seems like it works just fine without being there :)

agouin (approved):
:rocket:

PR #389, opened 2023-01-31
Validate DockerImages in ChainConfigs

bigs:
This change introduces implementations of IsFullyConfigured and Validate for DockerImages and correspondingly adds a check to Validate ChainSpec's Config helper function. This arose out of a debugging session wherein I realized I wasn't passing a UidGid value in with my DockerImage configuration which was causing the mounted volume to become unreadable. Since this value (as is Repository) is required for operation, I decided to add some simple validation.

It seemed like Config was the easiest way to get this validation called everywhere it needs to be, but if there is another area I can add it, I'd be happy to.

Furthermore, I generally like to solve problems like this by exposing a public interface and hiding the struct, so that users must construct the struct with a NewDockerImage function that validates the inputs, but this wasn't easily done since we rely on serializers and deserializers into the struct which requires us to expose the struct definition publicly. Just food for thought!


bigs:
Bumping this!

ash-burnt:
@Reecepbcups added you to the project, let me know if there is anything else I can help with. @bigs is no longer with us

bigs:
Tagging in @mvid on this one.

Reecepbcups (approved):

PR #391, opened 2023-02-06
docs: point godoc to v7

cmwaters:
The current readme points to https://pkg.go.dev/github.com/strangelove-ventures/interchaintest@main which doesn't seem to exist. This PR modifies the path to point to [v7](https://pkg.go.dev/github.com/strangelove-ventures/interchaintest/v7)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 Formerly known as `ibctest`.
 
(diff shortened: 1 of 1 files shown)


boojamya:
Instead of pointing directly to v7 on the main branch,
Can we point change the url to: `https://pkg.go.dev/github.com/strangelove-ventures/interchaintest`
Thus removing the @main.

It should re-direct accordingly.

boojamya:
This seems to have resolved itself. Closing

PR #393, opened 2023-02-08
Param proposals, balance inquiries, and IBC transfer error handling

bigs:
Some things old, some things new...

## New Features

In writing some tests involving the impact of some module params on IBC transfers, we added a few new methods to the `CosmosChain`.

### ParamChangeProposal

Submits a parameter change proposal to the gov module. For simplicity's sake (and since a majority of the `ChainNode` methods leverage the cosmos-sdk command-line interface to interact with the chain anyway) we borrowed our argument type directly from the `params` module's client utils module. Fairly straightforward and exposed at the `CosmosChain` level as well.

### AllBalances

Uses the gRPC client (as `GetBalance` does) to query for all of the coins held by an account address. Almost a carbon-copy of `GetBalance`.

## Updated Features

After encountering some unexpected behaviors in our tests, we found what we believe to be a bug in the `SendIBCTransfer` command.

### SendIBCTransfer response code handling

We noticed that `SendIBCTransfer` does not inspect the transaction response's `Code` before populating the `Tx` object and returning it. This leads to situations where a transaction goes through unsuccessfully but the test author has no way of knowing this, since the `ibc.Tx` object doesn't include the response code. We thought it more correct to return an error in this case.

Our use case here was disabling a token from being sendable in the bank module (via params, hence the params proposal.) What we found was that even though we were getting a response code 7 indicating that the transfer was unsuccessful (as we expected), from our test's perspective everything was going along well.

**Edit**: We also added a helper function for writing files into the docker volume, since we saw the same pattern duplicated many times over.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	paramsutils "github.com/cosmos/cosmos-sdk/x/params/client/utils"
 	"hash/fnv"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


mvid (approved):

agouin (approved):
Thank you! Looks great

PR #396, opened 2023-02-14
Add support for the Hermes relayer

### LINKED ISSUE #3: Hermes support
Great initiative with the work here, love it!

I just wanted to quickly drop a couple of resources here detailing our current process of automatically creating docker images with every release of Hermes. Hoping these hints could help understanding how to support Hermes in the testing framework.

- Docker tags uploaded here: https://hub.docker.com/r/informalsystems/hermes/tags
- Images are based on this docker file: https://github.com/informalsystems/ibc-rs/blob/master/ci/hermes.Dockerfile

chatton:
This PR adds an `ibc.Relayer` implementation for the Hermes relayer.

closes #3

I ran into a few different issues with the Relayer/RelayerCommander interfaces when working on this implementation, but was largely able to work around them, I'll give a little bit of context about what these issues were to clarify the reasoning behind some of the implementation decisions made here.

1. The concept of a "path" does not exist as a first class citizen within hermes. i.e. where the golang relayer has cli commands like `rly tx link <path>`, hermes does not have a direct equivalent. In this case the "path" concept was implemented at the interchain test layer.
2. The combination of the `DockerRelayer` and `RelayerCommander`  interfaces assume it is possible to run a single cli command and directly parse the output of that command to extract the desired type. For example the [UpdateClients](https://github.com/strangelove-ventures/interchaintest/blob/5f573c9f63a76de925747ed023793d640ab19a07/relayer/docker.go#L336-L340). The golang relayer can do a lot of these things in a single command, however I found that in a lot of cases there was additional logic that needed to happen. For this reason a lot of the commander methods I added are implemented just to satisfy the interface but will panic if called, and the main logic is happening in the new hermes.Relayer type.
3. The `AddChainConfiguration` assumes a config file per chain config, however in hermes it is a single config that contains an array of chain configs within the single file. This method was re-implemented on the hermes relayer to account for this.

I think the `RelayerCommander` interface works perfectly for the go relayer, but I suspect there will be similar issues to the ones I encountered when adding new relayers.


internal/dockerutil/filewriter.go
```diff
@@ -48,10 +47,10 @@ func (w *FileWriter) WriteFile(ctx context.Context, volumeName, relPath string,
 			Cmd: []string{
 				// Take the uid and gid of the mount path,
 				// and set that as the owner of the new relative path.
-				`chown "$(stat -c '%u:%g' "$1")" "$2"`,
+				`chown -R "$(stat -c '%u:%g' "$1")" "$2"`,
 				"_", // Meaningless arg0 for sh -c with positional args.
 				mountPath,
-				path.Join(mountPath, relPath),
+				mountPath,
```
chatton, line 53: I made this change as the hermes relayer ended up making several files in other places in the home directory and was running into permissions issues. I thought it might be fine if we just set the home directory to be owned by the relayer user. If this is a problem we can figure something else out!

examples/ibc/interchain_accounts_test.go
```diff
@@ -268,12 +268,12 @@ func TestInterchainAccounts(t *testing.T) {
...
 	require.NoError(t, err)
 	require.Equal(t, 1, len(chain1Chans))
-	require.Equal(t, "STATE_CLOSED", chain1Chans[0].State)
+	require.Subset(t, []string{"STATE_CLOSED", "Closed"}, []string{chain1Chans[0].State})
 
 	chain2Chans, err := r.GetChannels(ctx, eRep, chain2.Config().ChainID)
 	require.NoError(t, err)
 	require.Equal(t, 1, len(chain2Chans))
-	require.Equal(t, "STATE_CLOSED", chain2Chans[0].State)
+	require.Subset(t, []string{"STATE_CLOSED", "Closed"}, []string{chain2Chans[0].State})
```
chatton, line 276: I noticed the go relayer and the hermes relayer return different states, this just accounts for either of them

relayer/docker.go
```diff
@@ -122,6 +132,21 @@ func NewDockerRelayer(ctx context.Context, log *zap.Logger, testName string, cli
...
 
+// WriteFileToHomeDir writes the given contents to a file at the relative path specified. The file is relative
+// to the home directory in the relayer container.
+func (r *DockerRelayer) WriteFileToHomeDir(ctx context.Context, relativePath string, contents []byte) error {
+	fw := dockerutil.NewFileWriter(r.log, r.client, r.testName)
+	if err := fw.WriteFile(ctx, r.volumeName, relativePath, contents); err != nil {
+		return fmt.Errorf("failed to write file: %w", err)
+	}
+	return nil
+}
```
chatton, line 143: Added this helper function as the hermes relayer required a few different files to be created.

misko9:
Agree with your points on the assumptions for the interfaces. Glad to see you could work around them until we figure out the correct refactor. And I like what you did with the` hermes.Relayer` overrides.

relayer/hermes/hermes_commander.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (c commander) CreateWallet(keyName, address, mnemonic string) ibc.Wallet {
+	return NewWallet(keyName, address, mnemonic)
+}
+
+func (c commander) UpdatePath(pathName, homeDir string, filter ibc.ChannelFilter) []string {
+	// TODO: figure out how to implement this.
+	panic("implement me")
+}
```
chatton, line 144: I don't know if this is required or not for hermes, the tests are passing without.

relayer/hermes/hermes_wallet.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Get mnemonic, only used for relayer wallets
+func (w *Wallet) Mnemonic() string {
+	return w.mnemonic
+}
+
+// Get Address
+func (w *Wallet) Address() []byte {
+	return []byte(w.address)
+}
```
chatton, line 42: this is essentially the same as the cosmos relayer wallet.

chatton:
I just realised that the the test matrix seems to always be defaulting to the in memory one (not loading from file). I will look into this tomorrow!

chatton:
Hey @jtieri / @misko9  , I'm seeing some strange failures in the conformance tests. I'm seeing them in the both the go relayer and the hermes implementaiton so I'm not 100% sure if I've broken something. It appears as if both relayers are failing [here](https://github.com/chatton/interchaintest/blob/0bdc194c2aa11aa32479f32b19e1c50304301981/conformance/test.go#L418). 

In this PR I've temporarily hard coded hermes in to run the conformance test CI, but the `test-cosmos-examples` job should still be using the go relayer.

Please let me know if anything is jumping out about them!

seanchen1991 (approved):

chatton:
Looks like all the CI is now passing with hermes plugged into the conformance tests, thanks for the review @seanchen1991 , and thanks for the support @misko9 / @jtieri  ❤️

misko9 (approved):

PR #311, opened 2022-09-26
Add more pre-configured chains

boojamya:
This adds more preconfigured chains to the configuredChains.yaml file

Closes: https://github.com/strangelove-ventures/ibctest/issues/27

PR #399, opened 2023-02-16
Remove `labels` Package

boojamya:
Before you had to manually register a label for each chain which hindered chains from easily being pre configured. 
This has been a blocker for https://github.com/strangelove-ventures/interchaintest/pull/311

Now, as tests are run, a chain label gets automatically registered for each chain. The chain label will be the same as the chain name.

Note: this fix is specific for running tests using a matrix file.


label/label.go
```diff
@@ -38,6 +39,7 @@ var knownRelayerLabels = map[Relayer]struct{}{
 	Rly:    {},
 	Hermes: {},
 }
+var mu sync.Mutex
```
jackzampolin, line 42: Having a global mutex is not ideal is there an object you can embed this? Maybe @mark-rushakoff has a good pattern here and I could easily be wrong.

boojamya:
Closing in favor of: https://github.com/strangelove-ventures/interchaintest/pull/528

PR #411, opened 2023-02-28
deps: migration to cometbft

crodriguezvega:
Bumping Cosmos SDK and ibc-go to commits on branches working on the migration to cometbft.

PR #412, opened 2023-03-01
Update to cometbft

chatton:
This PR updates the sdk version to the tip of the sdk 47 release branch and changes all tendermint imports to cometbft. I also needed to update to a branch of ibc-go which which has updated to comeftbft. This will be changed in a follow up once we are able to run our E2E tests using comefbft


chatton:
closing in favour of #411

Issue #402, opened 2023-02-17
Intermittent failures occur during chain startup


chatton:
We periodically notice startup errors in our tests when spinning up the chain containers. For example [this](https://github.com/cosmos/ibc-go/actions/runs/3530940914/jobs/5923576191) one.

The errors look like

```
Error:      	Received unexpected error:
        	            	failed to start chains: failed to start chain simapp-b: All attempts fail:
        	            	#1: post failed: Post "http://localhost:49162": read tcp [::1]:43896->[::1]:49162: read: connection reset by peer
        	            	#2: post failed: Post "http://localhost:49162": read tcp [::1]:46596->[::1]:49162: read: connection reset by peer
        	            	#3: post failed: Post "http://localhost:49162": EOF
        	            	#4: post failed: Post "http://localhost:49162": read tcp [::1]:46618->[::1]:49162: read: connection reset by peer
        	            	#5: post failed: Post "http://localhost:49162": read tcp [::1]:50548->[::1]:49162: read: connection reset by peer
        	            	#6: post failed: Post "http://localhost:49162": read tcp [::1]:50550->[::1]:49162: read: connection reset by peer
        	            	#7: post failed: Post "http://localhost:49162": EOF
        	            	#8: post failed: Post "http://localhost:49162": EOF
        	            	#9: post failed: Post "http://localhost:49162": read tcp [::1]:58102->[::1]:49162: read: connection reset by peer
        	            	#10: post failed: Post "http://localhost:49162": read tcp [::1]:58116->[::1]:49162: read: connection reset by peer
        	            	#11: post failed: Post "http://localhost:49162": EOF
        	            	#12: post failed: Post "http://localhost:49162": read tcp [::1]:48660->[::1]:49162: read: connection reset by peer
        	            	#13: post failed: Post "http://localhost:49162": read tcp [::1]:48672->[::1]:49162: read: connection reset 
```

In most cases, they will succeed after a retry.

Briefly looking into the code, it looks like something is going wrong [here](https://github.com/strangelove-ventures/interchaintest/blob/0b3a8a07775d19fe06ae3f2c0ed5023faae93464/chain/cosmos/chain_node.go#L964-L967). It's difficult to tell if the containers would succeed after some more time, or if they would need to be re-created from scratch.

The retry polls over 2 minutes which seems like a reasonable time, but something is getting stuck somewhere.

I wonder would be it worth attempting a removal / re-creation of the container before returning the error.

[...]

PR #415, opened 2023-03-02
fix: remove race condition on start container retries

colin-axner:
While investigating https://github.com/cosmos/ibc-go/issues/2827 I came across an issue locally. If I created an interchain test with many validators/full nodes, I would get indeterminate errors on configuration startup. One issue was actually an address being returned twice in the standard output and then being reused for gentx, `{address}\n{address}` was the string being passed around. Another was the genesis file already existing (trying to execute simd init twice)

It became apparent to me that on a rare occasion, some of the config commands would get executed twice. With the help of @chatton, we determined it was likely the retry in `StartContainer`. I believe while a node was processing a request, the timeout would occur. Since this is a request being sent to a server, there's no way to ensure that request doesn't get executed. A retry with the same request would be sent and then I would see an error roughly saying that this tx already executed. 

The proposed changes removes the retry entirely. I'm not sure what use it currently provides. I propose using the calling context's timeout value, if it does not exist, use a standard timeout (30 seconds in the current pr, fine with whatever, but should likely be a defined variable). The only scenario I think a retry could be useful is if the first request attempted is expected to fail and the second may succeed, but the current code doesn't seem to handler this case as the first error would be marked as unrecoverable. How likely is this? Should a defensive retry be used if the timeout hasn't been reached? Are there any common errors returned which warrant a retry? Like a connection being reset?

Is there any context I am missing on why the code is the way it is? Thanks for taking to review this, we would love to patch up these indeterminacy issues.  

ref #402


colin-axner:
Marking as r4r. The CI on ibc-go doesn't run into this issue which is why I haven't re-reviewed this until now, but I noticed this [failure](https://github.com/strangelove-ventures/interchaintest/actions/runs/4485401315/jobs/7886909620), which seems like it's very possible a command is being executed twice given the error:
> found multiple connections on chain-a for client 07-tendermint-0

agouin (approved):
Thank you! I like the approach of the longer timeout and no retries to prevent potential double execution.

PR #423, opened 2023-03-09
feat: add ReadFile method for reading files from docker fs

jtieri:
Expose a `ReadFile` method on `ChainNode` so that you can read files from the docker filesystem within test cases that exist in downstream repos.


agouin (approved):
Look good! Would be good to update the internal places where this code is repeated to call this method too

PR #427, opened 2023-03-13
automatically calculate gas for TXs

mvid:
Ran into this issue with wasm contract deployments, which far exceed the default gas costs for chains


boojamya:
Hey @mvid , I don't think it's necessary to make this a default value for all transitions, this could have a large affect on other tests.

Check out the this comment on how to work around this:
https://github.com/strangelove-ventures/interchaintest/issues/482#issuecomment-1507432996

Closing.

PR #444, opened 2023-03-16
generate explicit port bindings

### LINKED ISSUE #443: Create port bindings instead of relying on docker random port bindings
Based on a lot of [investigation](https://github.com/cosmos/ibc-go/pull/3300#issue-1625762543), we believe there to be a bug in [docker](https://github.com/moby/moby/issues/42442) which is causing non-deterministic failures in our e2e's due to incorrect packet routing. The problem arises when relying on docker to do the port bindings for the rpc/grpc/p2p ports. I propose we provide in the host config for creating a container, the [port bindings](https://github.com/moby/moby/blob/master/api/types/container/hostconfig.go#L384). 

We will need to determine a starting port address and then increment by X for each chain and by Y for each node. For example:

Start at 32000.
Increment by 1000 per chain
Increment by 5 per node

chain a val 0: 32000, 32001, 32002 etc
chain a val 1: 32005, etc
chain b val 0: 33000, etc

The cosmos chain node type can be modified to store the full port bindings (it currently stores the host rpc and grpc ports). I believe the chain interface would need to be modified to support incrementing the starting port by its index. Maybe there's an alternative approach?

agouin:
Generate explicit port bindings to workaround [docker bug](https://github.com/moby/moby/issues/42442). This retains automatic port allocation functionality by opening temporary listeners on the next available port and holding that listener open until the handoff to docker for the container to consume the port.

Resolves #443

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -935,6 +935,11 @@ func (tn *ChainNode) CreateNodeContainer(ctx context.Context) error {
 			zap.String("image", imageRef),
 		)
 
+	pb, err := dockerutil.GeneratePortBindings(sentryPorts)
+	if err != nil {
+		return fmt.Errorf("failed to generate port bindings: %w", err)
+	}
+
 	cc, err := tn.DockerClient.ContainerCreate(
 		ctx,
 		&container.Config{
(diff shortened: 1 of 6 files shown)


boojamya (approved):
Interesting 👏

Solid investigation @colin-axner and team!

PR #445, opened 2023-03-17
Add image repo to upgradeVersion

boojamya:
This PR...
Adds the option to pass an image repo when performing a chain upgrade via `versionUpgrade`.

This will help specifically for local testing.

For example: You are testing a chain upgrade process where you upgrade from remote repository to a local image. 

Currently, you can only pass in a `version` and the repository is inherited from the chain config. 
Being able to specify the repository solves the issue.


boojamya:
Closing. This is not necessary.
Workaround: https://github.com/strangelove-ventures/noble/blob/07b0d0015ff439af735fe74c2c412feaaefafcb0/interchaintest/upgradeauthority_test.go#L197-L201

Reecepbcups:
I still think this is needed for those  who are using images not on the main branch yes? 
Heighliner uses ghcr `juno`, but we use `juno-e2e`.

So I had to do a work around to get it to work properly:
https://github.com/CosmosContracts/juno/blob/2504cb2d25e17f6d37cb60b9eadd97fc0e7b4bb4/tests/interchaintest/chain_upgrade_test.go

(Your workaround did not work for me in this instance when moving off the main docker repo)

Took me forever to figure this out

chain/cosmos/cosmos_chain.go
```diff
@@ -506,13 +506,15 @@ func (c *CosmosChain) GetGasFeesInNativeDenom(gasPaid int64) int64 {
 	return int64(fees)
 }
 
-func (c *CosmosChain) UpgradeVersion(ctx context.Context, cli *client.Client, version string) {
+func (c *CosmosChain) UpgradeVersion(ctx context.Context, cli *client.Client, ContainerRepo, version string) {
```
agouin, line 509: var name should be lowercase `containerRepo`

agouin (approved):
One nit otherwise LGTM!

chain/cosmos/cosmos_chain.go
boojamya: 👍

PR #448, opened 2023-03-19
use correct grpc with sdk45

faddat:
In interchaintest v4, there was a bug preventing its use with chains that use sdk 45.  

SDK 45 mandates the use of an older grpc package, to be compatible with the regen modified protobuf 
compiler.  

This PR adds a replace statement and fixes this:

```go
	conn, err := grpc.Dial(grpcAddress, grpc.WithInsecure)
```

-- it's called differently in newer versions of grpc.


faddat:
I don't think that my changes caused the first test to fail, could you confirm?

agouin (approved):
Thank you!

PR #465, opened 2023-03-29
feat: add the ability to manage sidecar processes

jtieri:
There are cases where it is necessary to run some sidecar process on a per validator basis and/or on a chain basis. One example of this is the orchestrator needed to run a Sommelier validator. This PR introduces the concept of a `SidecarProcess` which can be scoped either at the chain level or at the validator level.

`SidecarProcesses` can optionally be started before the chain or validator node starts or explicitly started after the chain or validator is started.


chain/cosmos/chain_node.go
```diff
@@ -121,6 +125,46 @@ func (tn *ChainNode) NewClient(addr string) error {
...
+
+		VolumeName: v.Name,
+		ImageRef:   image.Ref(),
+		TestName:   testName,
+		UidGid:     image.UidGid,
+	}); err != nil {
+		return nil, fmt.Errorf("set volume owner: %w", err)
+	}
+
+	return s, nil
```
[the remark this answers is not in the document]
jtieri: Yeah I kinda fleshed this out yesterday without connecting the moving pieces. Just added the append here as well as the pieces we should need to instantiate sidecar processes from the chain config

chain/cosmos/sidecar.go
```diff
@@ -0,0 +1,172 @@
...
+	Chain ibc.Chain
+
+	// If true this process is scoped to a specific validator, otherwise it is scoped at the chain level.
+	validatorProcess bool
+
+	// If true this process should be started before the chain or validator, otherwise it should be explicitly started after.
+	preStart bool
+
+	// Set to true in StartContainer call so that we do not accidentally kick off the same process more than once.
+	started bool
```
agouin, line 32: maybe instead of this flag, this could be a method implemented in the ContainerLifecycle that inspects docker for the running container ID?

jtieri: much better idea! just made changes to account for this

ibc/types.go
```diff
@@ -42,6 +42,8 @@ type ChainConfig struct {
 	EncodingConfig *testutil.TestEncodingConfig
 	// Required when the chain uses the new sub commands for genesis (https://github.com/cosmos/cosmos-sdk/pull/14149)
 	UsingNewGenesisCommand bool `yaml:"using-new-genesis-command"`
+	// Configuration describing additional sidecar processes.
```
agouin, line 45: This covers chain-level sidecars, but can you include an additional slice for per-node sidecars?

aljo242:
@jtieri I think you will need to add the following logic to `ChainConfig.Clone()` in `ibc/types.go`

```go
func (c ChainConfig) Clone() ChainConfig {
	x := c

	images := make([]DockerImage, len(c.Images))
	copy(images, c.Images)
	x.Images = images

	sidecars := make([]SidecarConfig, len(c.SidecarConfigs))
	copy(sidecars, c.SidecarConfigs)
	x.SidecarConfigs = sidecars

	return x
}

```

ibc/types.go
jtieri: I was handling all sidcars from this single slice by using the bool field `ValidatorProcess`. the way I imagined this would work is people would instantiate the sidcars and add them to the slice, for any sidecar that should be scoped at a validator level `ValidatorProcess` will be `true`

Chain level sidecars are handled here https://github.com/strangelove-ventures/interchaintest/blob/7d8dff5bc89351f27ead0b24315fd5ed43572ad4/chain/cosmos/cosmos_chain.go#L697-L725

and then the validator sidecars are handled here 
https://github.com/strangelove-ventures/interchaintest/blob/7d8dff5bc89351f27ead0b24315fd5ed43572ad4/chain/cosmos/cosmos_chain.go#L587-L596

im indifferent to the path forward here and can definitely see the benefits of making this more explicit so if you think thats a better design i can make those changes

jtieri:
> 

grazie! I'm picking this back up so will get this added in there

aljo242:
@jtieri any updates on this?

boojamya:
@Mergifyio backport v6

PR #467, opened 2023-03-30
Adds ParamChangeProposal

### LINKED ISSUE #464: feat request: Param Change Proposal
## Feat
For juno, we would like to test multiple param change proposals. I will look into this later this week when I get time

## Reason
 The most important being: altering the IBC unbonding time from 28 days to less than the 2/3rd period and seeing how the chain handles.

I know both Juno & Nobile need GlobalFee E2E test too, and this would be helpful there as well
(Start chain with 0 gas fees, param change to 0.001token, try tx and expect fail with 0 fees)

Reecepbcups:
Closes #464 . Targeting v4 branch here

## Question
- Once approved, should I make another PR targeting main?

## Changes
- Adds a way to query a subspace param
- ParamChangeProposal
- Example test in action

## Reason
- We need this in v4 for Juno to test lowering unbonding period & the effects on IBC & GlobalFee
- Noble (v3) also has globalfees, so this is helpful to test that too with e2e per the todo https://github.com/strangelove-ventures/noble/pull/146

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -767,6 +767,20 @@ func (tn *ChainNode) QueryProposal(ctx context.Context, proposalID string) (*Pro
 	return &proposal, nil
 }
 
+// QueryParam returns the state and details of a subspace param.
+func (tn *ChainNode) QueryParam(ctx context.Context, subspace, key string) (*ParamChange, error) {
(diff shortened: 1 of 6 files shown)


agouin (approved):
Thank you for the example!

boojamya:
Nice! @Reecepbcups 
Pushing to `main` would be helpful.
`v3` branch too would be a bonus (for noble).

PR #468, opened 2023-03-31
chore: update `v5` dependencies

aljo242:
Update some core and minor dependencies of the `v5` branch of interchaintest:
- sdk and ibc-go
- use comet-bft

Fix the `make ibctest` command as done in https://github.com/strangelove-ventures/interchaintest/pull/326


aljo242:
bumping go to `1.19` seemed to mess with conformance test, so I reverted that change.

jtieri:
hey @aljo242 it looks like the unit tests are failing because this line is missing which filters out the tests in the `examples` directory

https://github.com/strangelove-ventures/interchaintest/blob/5baa47f70e73ebdd90d32d66c7a0b3e1d04d52a2/.github/workflows/tests.yml#L35

once that's in place this should pass CI and we can get this merged

Issue #474, opened 2023-04-04
Wrong value return in `ExportState` 


ThanhNhann:
## Context 

Refer to this: https://github.com/cosmos/ibc-go/pull/3282#discussion_r1155881792, I think the return of `ExportState` function should be stdout, not stderror

## Suggestion

Update the value return of `ExportState`

PR #475, opened 2023-04-04
Update `ExportState` 

ThanhNhann:
Close: #474

Update value return of `ExportState` function


chain/cosmos/chain_node.go
```diff
@@ -922,12 +922,12 @@ func (tn *ChainNode) ExportState(ctx context.Context, height int64) (string, err
...
 
-	_, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
+	stdout, _, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
 	if err != nil {
 		return "", err
 	}
-	// output comes to stderr for some reason
-	return string(stderr), nil
+
+	return string(stdout), nil
```
agouin, line 930: Can we concatenate stdout and stderr so that it is backwards compatible with prior binaries where the state export comes to stderr? We made this change in the v3 branch already but looks like it was not brought forward into main.

https://github.com/strangelove-ventures/interchaintest/blob/v3/chain/cosmos/chain_node.go#L879-L889

ThanhNhann: Updated, thank @agouin

ThanhNhann:
hey @agouin, can you plz check the CI, I checked the fails above but it runs ok in my local

agouin (approved):

PR #480, opened 2023-04-05
penumbra - pclientd integration

agouin:
This PR introduces a basic integration with `pclientd`, which is a standalone client daemon for interacting with private user data on Penumbra. We have ported over a [network integration](https://github.com/penumbra-zone/penumbra/blob/45bdbbeefc2f0d3ebf09e2f37d0545d8b1e094d8/crates/bin/pclientd/tests/network_integration.rs) test from the Penumbra repo to assert that basic functionality via `pclientd` is working as intended.


chain/penumbra/penumbra_app_node.go
```diff
@@ -132,47 +193,69 @@ func (p *PenumbraAppNode) GenerateGenesisFile(
...
 	fw := dockerutil.NewFileWriter(p.log, p.DockerClient, p.TestName)
 	if err := fw.WriteFile(ctx, p.VolumeName, "validators.json", validatorsJson); err != nil {
 		return fmt.Errorf("error writing validators to file: %w", err)
 	}
-	allocationsCsv := []byte(`"amount","denom","address"\n`)
+
+	allocationsCsv := []byte("\"amount\",\"denom\",\"address\"\n")
 	for _, allocation := range allocations {
-		allocationsCsv = append(allocationsCsv, []byte(fmt.Sprintf(`"%s","%s","%s"\n`, allocation.Amount.String(), allocation.Denom, allocation.Address))...)
+		allocationsCsv = append(allocationsCsv, []byte(fmt.Sprintf("\"%s\",\"%s\",\"%s\"\n", allocation.Amount.String(), allocation.Denom, allocation.Address))...)
```
agouin, line 204: this is logically the same, but the previous string with the ` is easier to understand

jtieri: this was my thought originally as well but we were seeing an error when attempting to parse the allocation csv file during start up.

```
Error Trace:	/Users/justintieri/go/src/github.com/strangelove-ventures/interchaintest/examples/penumbra/penumbra_chain_test.go:56
                 generating genesis file: failed to exec testnet generate: exit code 1: 2023-09-18T18:04:18.424475Z  INFO pd: Generating network config chain_id="penumbra-1"
        	Error: could not parse allocations file "/home/heighliner/allocations.csv"
        	            	
        	Caused by:
        	            CSV error: record 1 (line: 2, byte: 27): found record with 9 fields, but the previous record has 3 fields
```

chain/penumbra/penumbra_chain.go
```diff
@@ -109,56 +104,65 @@ func (c *PenumbraChain) Initialize(ctx context.Context, testName string, cli *cl
...
 
 // GetHostGRPCAddress returns the address of the gRPC server accessible by the host.
 // This will not return a valid address until the chain has been started.
 func (c *PenumbraChain) GetHostGRPCAddress() string {
-	return c.getRelayerNode().PenumbraAppNode.hostGRPCPort
+	return c.getFullNode().PenumbraAppNode.hostGRPCPort
 }
 
 func (c *PenumbraChain) HomeDir() string {
 	panic(errors.New("HomeDir not implemented yet"))
```
Reecepbcups, line 138: HomeDir() can now use - `return c.getFullNode().PenumbraAppNode.HomeDir()`

chain/penumbra/penumbra_chain.go
```diff
@@ -528,19 +503,23 @@ func (c *PenumbraChain) start(ctx context.Context) error {
...
 		sep, err := n.TendermintNode.GetConfigSeparator()
 		if err != nil {
 			return err
 		}
+
+		tmPort := strings.Split(rpcPort, "/")[0]
 		eg.Go(func() error {
 			return n.TendermintNode.CreateNodeContainer(
 				egCtx,
 				fmt.Sprintf("--proxy%sapp=tcp://%s:26658", sep, n.PenumbraAppNode.HostName()),
```
Reecepbcups, line 516: `26658` here should use the new `abciPort` const that was added

examples/penumbra/penumbra_chain_test.go
```diff
@@ -2,50 +2,127 @@ package penumbra_test
...
+
+	err = chain.SendFunds(ctx, alice.KeyName(), transfer)
+	require.NoError(t, err)
+
+	/*
+		TODO:
+		without this sleep statement we see intermittent failures where we will observe the tokens taken from alice's balance
+		but not added to bob's balance. after debugging it seems like this is because alice's client is in sync but bob's is not.
+		we may need a way to check if each client is in sync before making any assertions about chain state after some state transition.
+		alternatively, we may want to wrap penumbra related queries in a retry.
```
Reecepbcups, line 113: The newly added `testutil.WaitForBlocksUtil` method likely helps here

chain/penumbra/penumbra_app_node.go
Reecepbcups: Go's raw strings treat all values literally, so the escape character here throws it off

\`"%s","%s","%s"\`+"\n" - would resolve this

same for allocation if you want: \`"amount","denom","address"\` + "\n"

Reecepbcups (approved):
Looks good! A few minor comments but overall solid work! :fire:

chain/penumbra/penumbra_client_node.go
```diff
@@ -0,0 +1,396 @@
...
+	return dockerutil.CondenseHostName(p.Name())
+}
+
+// Bind returns the home folder bind point for running the node
+func (p *PenumbraClientNode) Bind() []string {
+	return []string{fmt.Sprintf("%s:%s", p.VolumeName, p.HomeDir())}
+}
+
+func (p *PenumbraClientNode) HomeDir() string {
+	return "/home/heighliner"
```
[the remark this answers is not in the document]
jtieri: good question, the actual full nodes in the network require `/home/heighliner` because of how the filesystem is setup via the Dockerfile that is used to build the penumbra images [here](https://github.com/strangelove-ventures/heighliner/blob/15a4dbc94689dce71efb0368b8bcfa2052cce526/dockerfile/cargo/Dockerfile#L308) but i think this occurrence could be any arbitrary string because the `PenumbraClientNode` describes a client daemon that is speaking to a full node.

I'll double check that the test still passes with any arbitrary string as the home dir and if so i'll change this to `/home/pclientd` since any files generated in this directory should specifically be related to the penumbra client daemon

chain/penumbra/penumbra_app_node.go
```diff
@@ -132,47 +193,69 @@ func (p *PenumbraAppNode) GenerateGenesisFile(
...
+	pdUrl := fmt.Sprintf("http://%s:8080", p.HostName())
+	cmd := []string{"pcli", "-d", keyPath, "-n", pdUrl, "view", "balance"}
+
+	stdout, _, err := p.Exec(ctx, cmd, nil)
+	if err != nil {
+		return 0, err
+	}
+
+	fmt.Printf("STDOUT BAL: '%s'\n", string(stdout))
+	return 0, nil
```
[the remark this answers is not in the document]
jtieri: I'm not 100% sure, they do technically return the same information the only difference being which piece of software is used to retrieve the balance (`pclientd` vs. `pcli`). 

This PR moves away from making queries and broadcasting txs from the `PenumbraAppNode` in favor of the `PenumbraClientNode` when possible, so technically this method is unused now within `interchaintest` but it's not immediately clear to me if it will be used moving forward or if it's beneficial to expose methods on both objects. 

I could remove this for now, change the method signature to accept a denom string and actually process the output properly so we can remove the todo, or open up an issue to come back to this later in a follow up PR. Thoughts?

chain/penumbra/penumbra_chain.go
jtieri: good catch! i'll update this

examples/penumbra/penumbra_chain_test.go
jtieri: great call out, this looks like it could be helpful. perhaps there is a way to check if a `pclientd` instance is in sync via a cmd that we can wrap in a function and pass into `testutil.WaitForBlocksUtil`

PR #489, opened 2023-04-11
Add ability to override default `--gas-prices` and `--gas-adjustment` flags

boojamya:
Before this PR... when construction or executing transaction commands,
`Interchain` test would just amend any string passed into `command` to the final broadcasted command.

Because `interchaintest` passes in values for `--gas-prices` and `--gas-adjustment` by default, it created the potential for duplicate flags with different values to be constructed.

This PR fixes this issue,
If the tester passes in one of these flags as part of the `command` argument, `interchaintest` will only use the "tester specified" value.


chain/cosmos/chain_node.go
```diff
@@ -376,14 +376,35 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx,
...
 // with the chain node binary.
 func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
 	command = append([]string{"tx"}, command...)
+	var gasPrices, gasAdjustment string = tn.Chain.Config().GasPrices, fmt.Sprint(tn.Chain.Config().GasAdjustment)
+	for i := 0; i < len(command); i++ {
+		switch command[i] {
+		case "--gas-prices":
+			gasPrices = command[i+1]
+			fmt.Println("Removing command:", command[i])
+			command = removeAtIndex(command, i, 2)
```
agouin, line 395: I know we went through this before, but thinking about it now, why don't we just only append the chain config value flags if we don't find the flags already in the list? That should be more concise than removing and re-adding.

chain/cosmos/chain_node.go
```diff
@@ -380,10 +380,19 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height uint64) ([]blockdb.Tx,
 // with the chain node binary.
 func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
 	command = append([]string{"tx"}, command...)
+	var gasPrices, gasAdjustment string = tn.Chain.Config().GasPrices, fmt.Sprint(tn.Chain.Config().GasAdjustment)
+	for i := 0; i < len(command); i++ {
```
agouin, line 384: With the most recent modification, it will potentially add the flags twice. I am not sure if it supports this. I was suggesting we have bools such as `gasPricesFound, gasAdjustmentFound := false, false` then in the for loop, if we find them, set the relevant bool to true. Then after the for loop, check each bool and append the chain config value for false

boojamya: Ahh!
Third times the charm? @agouin

agouin (approved):
:fuelpump:

Issue #482, opened 2023-04-05
Interchaintest v4 does not use gas adjustment from chain spec when executing transaction


0xekez:
To reproduce, check out the [`zeke/gas-limits`](https://github.com/DA0-DA0/polytone/tree/zeke/gas-limits/tests/strangelove) branch and run `just integrationtest`.

Note that the chain config used sets gas price and adjustment values:

```go
ibc.ChainConfig{
	Denom:          "ujuno",
	GasAdjustment:  2.0,
	GasPrices:      "0.00ujuno",
	EncodingConfig: wasm.WasmEncoding(),
}
```

But that in test output those values are not used. For example, when instantiating a contract the log output contains:

```
"command": "junod tx wasm instantiate 2 {\"proxy_code_id\":\"3\",\"block_max_gas\":\"100000000\"} --label wasm-contract --no-admin --from default-juno1-1-xqp --gas-prices 0.00ujuno --gas-adjustment 1.3 --keyring-backend test --output json -y --home /var/cosmos-chain/juno1 --node tcp://juno1-1-fn-0-TestOutOfGas:26657 --chain-id juno1-1"
```

Note specifically: `--gas-adjustment 1.3` in the output which differs from the gas adjustment specified in the configuration. The result of this is that I am unable to mock non-trivial contract interactions as the transaction runs out of gas.

---

PS: the contract instantiation running out of gas does not cause the expected error as [`ExecTx`](https://github.com/strangelove-ventures/interchaintest/blob/v4/chain/cosmos/chain_node.go#L707) does not error. Instead, the query for listing the instances of a code ID returns an empty list and the error printed to the console is `panic: runtime error: index out of range [-1]` while [indexing into the code ID list](https://github.com/strangelove-ventures/interchaintest/blob/v4/chain/cosmos/chain_node.go#L722) which has a length of zero. you can confirm that the error occuring is an out of gas error only by modifying the interchain test code to print the transaction output of `ExecTx` like so:

```go
	tx, err := tn.ExecTx(ctx, keyName, command...)
	if err != nil {
		return "", err
	}

	txout, _, err := tn.ExecQuery(ctx, "tx", tx)
	if err != nil {
		return "", err
	}
	tn.logger().Warn("Instantiate TX response", zap.String("out", string(txout)))
```

0xekez:
[...]

PR #496, opened 2023-04-13
InstantiateContract command should have the ability to intake extra args

boojamya:
Inside the `InstantiateContract` function, it  executes a transaction (`tn.ExecTx`). `ExecTx` has the ability to intake extra command line arguments. We should extend this ability to the `InstantiateContract` function so you can pass along an extra args when `ExecTx` is run.

A specific use case for this would be passing in specific gas related flags. 

Relevant issue: https://github.com/strangelove-ventures/interchaintest/issues/482

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -743,8 +743,9 @@ func (tn *ChainNode) StoreContract(ctx context.Context, keyName string, fileName
 }
 
 // InstantiateContract takes a code id for a smart contract and initialization message and returns the instantiated contract address.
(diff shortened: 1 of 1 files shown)


kogisin:
Shouldn't this [InstantiateContract](https://github.com/strangelove-ventures/interchaintest/blob/dan%2Fextraargs/chain/cosmos/cosmos_chain.go#L410) function also need to intake the extra argument?

boojamya:
> 

Yes! Good catch. TY.

jackzampolin (approved):

PR #498, opened 2023-04-15
Fix lint errors and dependency issues

misko9:


.github/workflows/lint.yml
```diff
@@ -17,4 +17,4 @@ jobs:
         with:
           version: latest
           only-new-issues: true
-          args: --timeout=3m
+          args: --timeout=5m
```
boojamya, line 20: I don't think this is necessary b/c when the linter runs successfully, it takes less than a minute. 

I do see it timeout occasionally, but I think it is due to a flaky runner or some other issue.

boojamya (approved):

PR #501, opened 2023-04-19
Bump rly default version to v2.3.1

agouin:
Bump rly default version to v2.3.1 and remove custom image for tm v0.37 boundary test


misko9 (approved):
Approved, if test-ibc-examples really passed on the first run!

PR #525, opened 2023-04-24
Add dependabot "dont_backport" label

boojamya:
This PR updates dependabot to add a `dont_backport` label when it updates dependancies on the `main` branch.

This PR also includes minor cosmetic update to the mergify PR's.


boojamya:
Since opening this PR, I now have a different stance. 
I don't think it is necessary to to have dependabot run on all maintained branches. It's a bit overkill.
Keeping it just on main should be fine.

We can update deps on maintained branches on an as-need basis.

PR #529, opened 2023-04-25
Modify CI Test Suite

boojamya:
This PR...
- Allows dependabot dependency PR's to be merged with out going through the entire E2E test suite.
- Adds the `-failfast` flag to end-to-end tests

Given how throughout and extensive our end-to-end tests are, I don't think it is necessary to run the entire test suite for dependabot's dependency PR's. The unit-tests and linter may suffice. 

Currently, it can take 30+ minutes for some tests to complete and while we continue to improve reliability of our E2E test suite, some tests still produce false positives. 

The PR will allow us to be more agile and merge PR's quicker.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,31 +1,11 @@
-name: Run Go Tests
+name: Run End-To-End Tests
 
 on:
   pull_request:
+    branches-ignore:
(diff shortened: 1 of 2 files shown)


boojamya:
This will also need to be integrated into branch protection.
Closing for now.

Issue #530, opened 2023-04-26
Broadcaster should provide account sequence by default


chatton:
See [this comment](https://github.com/cosmos/ibc-go/pull/3459#issuecomment-1523260804) for full details.

After sdk v0.47.1 it is required to provide the account sequence.

We have a temprorary hack to add it in ibc-go, but this makes the most sense to be included in [this](https://github.com/strangelove-ventures/interchaintest/blob/a06599d2eb8c9c59e70158ae0339cd7b5ab6377a/chain/cosmos/broadcaster.go#L169-L183) function.

boojamya:
Closing via: https://github.com/strangelove-ventures/interchaintest/pull/533

PR #533, opened 2023-04-26
Pass account sequence to default tx factory.

DimitrisJim:
Should fix up https://github.com/strangelove-ventures/interchaintest/issues/530 by passing the account sequence to the tx factory in the default case.


jtieri (approved):
lgtm, thanks for the PR!

PR #559, opened 2023-05-08
Modify CI

boojamya:
This PR...

- Adds the `-failfast` flag to all tests in the CI.
- Moves the unit tests to use the github runner instead of our own self-hosted runner


agouin (approved):

boojamya:
Unfortunately I've had to re-try the unit tests 6+ times now for this PR.

Putting it back on the self-hosted runner.

Issue #539, opened 2023-04-27
Maintained Branches should be in Sync and Up To Date with `main`


boojamya:
Our maintained branches (v3, v3-ics, v4, v5, v6) have gotten out of sync with each other and from `main`.

To combat this, the easiest solution may be to:

1) Ensure that `main` has all features and code additions that have been pushed to other branches.
2) Branch a `vx-old` for all of our maintained branches
3) Delete current maintained branches
4) Re-branch all maintained branches from `main`, downgrading to the proper version of `ibc-go` and the `sdk` 

I think we should drop support for `v3-ics` and gain support for a `v4-ics`.

From there, our mergify integration should properly backport anything merged into `main` to all our maintained branches. 

### Progress
- [x] v3
- [x] v4
- [x] v4-ics
- [x] v5
- [x] v6

boojamya:
**Prepped Branches:**

- v3: https://github.com/strangelove-ventures/interchaintest/tree/dan/v3_new
- v4: https://github.com/strangelove-ventures/interchaintest/tree/dan/v4_new
- v4-ics
- v5
- v6

boojamya:
Done. 
Readme describing newly maintained branches here: https://github.com/strangelove-ventures/interchaintest/pull/647

PR #562, opened 2023-05-09
Upstream #470 - Adds ParamChangeProposal

boojamya:
This PR...

(Partially) Upstreams #470. 

Unlike in `v3`, the `ParamChangeProposal` function was already implemented. However, in this branch ('main'), it intakes a type rather than an actual json file. If necessary, the json file can be unmarshalled during specific test cases if a user prefers. 

Happy to adjust this if intaking a file to the `ParamChangeProposal` is a better flow. 

Also... (unrelated to the upstream)

I didn't think it was necessary to have both `chain_upgrade_test.go` and `chain_upgrade_ibc_test`. 
They seem to be testing the same thing, except the IBC version ran conformance before and after the upgrade.
I kept the IBC version.


boojamya:
cc: @Reecepbcups 
As the OG author of #470 and #467, just want to make sure you see this. 

We will be branching new versions of our maintained branches off main (see issue: https://github.com/strangelove-ventures/interchaintest/issues/539)

If you have any concerns, let us know.

Reecepbcups:
@boojamya Thanks! yea the unmarshing is a good point - I was writing it from an idea of making it easy to show examples of pre formatted json files ready for chain use. 

Unmarshing will work for this

chain/cosmos/genesis.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/icza/dyno"
+	"github.com/strangelove-ventures/interchaintest/v7/ibc"
+)
+
+func ModifyGenesisProposalTime(votingPeriod string, maxDepositPeriod string) func(ibc.ChainConfig, []byte) ([]byte, error) {
```
misko9, line 11: This works for gov modules that are on v1beta1.
Gov v1 will look like this, so maybe break out into two functions with a Gov v1beta1 vs v1 name?
```
dyno.Set(g, votingPeriod, "app_state", "gov", "params", "voting_period")
dyno.Set(g, maxDepositPeriod, "app_state", "gov", "params", "max_deposit_period")
dyno.Set(g, chainConfig.Denom, "app_state", "gov", "params", "min_deposit", 0, "denom")
```

boojamya: Ahh. Good eye!
Will take a pass at this.

boojamya: @misko9
Ok, added gov module version check.

misko9 (approved):
LGTM!

PR #565, opened 2023-05-12
SDK v47 features from Juno

Reecepbcups:
Upstream of features needed to use ICtest for Juno on v47 working branch

I know you are working on streamlining gov proposal across versions, so unsure if any of this contradicts any of that. Happy to close if so

# New Features
- Generalized ModifyGenesis function
- Adds a BuildProposal method for govv1 types *(Is there a better location for this?)*

# Tweaks
- Uses gas for all Txs
- Makes SubmitProposal public so a user can submit any arbitrary govv1 proposal easily

### proposal in action:
- https://github.com/CosmosContracts/juno/blob/47-upgrade/interchaintest/contract_unity_submit_test.go#L49-L60

### genesis in action:
https://github.com/CosmosContracts/juno/blob/47-upgrade/interchaintest/chain_upgrade_test.go#L37-L59


Reecepbcups:
cc @boojamya I know you are refactoring gov proposals across versions so wanted to tag you here. happy to remove this SubmitProposal if you have other plans

chain/cosmos/chain_node.go
```diff
@@ -406,6 +406,7 @@ func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
 		"--from", keyName,
 		"--gas-prices", tn.Chain.Config().GasPrices,
 		"--gas-adjustment", fmt.Sprint(tn.Chain.Config().GasAdjustment),
+		"--gas", "auto",
```
boojamya, line 409: I think we should avoid adding this flag at this level.


I'm not sure if you caught these changes but we (somewhat) recently added the ability to pass extra args to the `InstantiateContract` function. So you should be able to pass this `"--gas", "auto"` flag there.

Is this where you are needing this flag?

chain/cosmos/genesis.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/icza/dyno"
+	"github.com/strangelove-ventures/interchaintest/v7/ibc"
+)
+
+type GenesisKV struct {
+	Key   string      `json:"key"`
+	Value interface{} `json:"val"`
+}
+
+func ModifyGenesis(genesisKV []GenesisKV) func(ibc.ChainConfig, []byte) ([]byte, error) {
```
boojamya, line 18: Huge fan of the generic way you implemented this 👏 

It beats the v1beta1 vs v1 check we are doing in the current `ModifyGenesisProposalTime` function.

Would you be able to bring this up to date with `main` use your `ModifyGenesis` instead of the current `ModifyGenesisProposalTime`

boojamya (changes requested):

chain/cosmos/genesis.go
Reecepbcups: :pray: Thank you! Resolved in 2b9025f84e739cfc0979766b093af37c00de77ca when I merged main in

chain/cosmos/chain_node.go
Reecepbcups: I missed that, I believe it was why I had this flag. It may affect my local-interchain impl so I will need to test more. b8ef5ef720abbcf3a45bd7b8bcbe1f509e2e78a5

boojamya (approved):
Awesome! Thank you Reece!

PR #571, opened 2023-05-18
Use ghcr registry for hermes, bump version.

DimitrisJim:
Changes the registry to ghcr which hermes is now using and bumps the version.


agouin (approved):
Thank you!

PR #608, opened 2023-05-30
CI Efficiency: Cancel in-progress jobs if a new commit is pushed to a PR

boojamya:
We have an elaborate CI pipeline.

Currently if a commit is pushed to a PR, the CI runner will finish out any in-progress jobs. Some of these jobs can take 10+ minutes.

Adding this concurrency feature will cancel any in-progress jobs to prioritize the new commit.


boojamya:
@agouin think this will be OK with the docker cleanup?

PR #609, opened 2023-06-01
Disable redundant caches in CI

boojamya:
We use the `actions/setup-go@v4` four times in one test. 

By default, the action enables caching...

I've noticed that `Post Setup Go 1.19` can *sometimes take 10+ minutes. 

Seeing if disabling caching on all but 1 of the 4 uses makes this quicker.

Screenshot 2023-06-01 at 3 44 50 PM


boojamya:
It seems to be working quite well. 
Un-drafting.

agouin (approved):

PR #610, opened 2023-06-02
More specific test Interface + enable chain REST API

Reecepbcups:
For https://github.com/Reecepbcups/local-interchain _(Per my talk with Jack, this will be upstreamed to this repo after the merge of this PR)_

- Squashes in my v47 changes from https://github.com/strangelove-ventures/interchaintest/pull/565 as this is required for the Genesis updates

Other than this, it is just interface tweaks so I can run without testing.T & compile into a binary. REST API is a nice to have so users can query and build off locally.


relayerfactory.go
```diff
@@ -13,11 +12,15 @@ import (
...
+type TestName interface {
+	Name() string
+}
+
 // RelayerFactory describes how to start a Relayer.
 type RelayerFactory interface {
 	// Build returns a Relayer associated with the given arguments.
 	Build(
-		t *testing.T,
+		t TestName,
```
boojamya, line 23: @DavidNix 
See any issues here? 
Reece notes why this is necessary in the description. Would be good to run it by you too.

DavidNix: This is fine. There is a new interface on line 15 `TestName` of which `*testing.T` implements. 

I'm betting this abstraction allows us to not care about creating an actual `*testing.T` when, in the context of this feature, you don't need because you are not making test assertions. Rather, you're trying to spin up a "mini-network" on your machine.

DavidNix (approved):

PR #647, opened 2023-06-27
Update Readme + Mergify Rule for Maintained Branches

boojamya:
This PR:

- Updates the `Maintained Branches` Table on the Readme

- Updates the Mergify rule 
   - Will not backport PR's to `v3` OR `v3-ics` anymore
   -  Instead of backporting all PR's EXCEPT if they have a `dont_backport` lable, this changes the rule to only backport PR's that have a `BACKPORT` label.
   - I think this makes more sense as there will most likely be more PR's pointed at specific branches.
   - It also updates the rule to not only backport PR's merged into `main`, but now allows for PR's to be targeted to any of our maintained branches to also be backported. EXAMPLE: PR targeted at `v5` branch with the backport label will also get "backported" into `main`.
   - Because of this rule change, dependabot does not need to label with `dont_backport` anymore

(Branch protection rules already have been updated)

I'm also hoping to use this PR as a test to ensure that backporting works correctly and everything in CI is passing across all branches.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,13 +11,11 @@ updates:
       interval: "weekly"
     open-pull-requests-limit: 10
     labels:
-      - dependencies
-      - dont_backport
-      
+      - dependencies      
   - package-ecosystem: github-actions
     directory: "/"
     schedule:
       interval: monthly
     open-pull-requests-limit: 10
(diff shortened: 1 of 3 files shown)


agouin (approved):
:rocket:

PR #667, opened 2023-07-17
add terpnetwork to configuredChains

hard-nett:


boojamya (changes requested):
Thanks @hard-nett,
Could you match the indentation as the other chains please?

boojamya (approved):

PR #671, opened 2023-07-17
Support localhost client/connection in conformance test

### LINKED ISSUE #670: Conformance test does not support localhost client
The current conformance test sets up a connection on a pair of chains and checks that [only one connection exists](https://github.com/strangelove-ventures/interchaintest/blob/cedf87f4624d3b65b1684632aa11adb3cb84358d/conformance/relayersetup.go#L94-L101) (i.e. only one was created). As of newer versions of ibc-go, chains now have the localhost connection as well, which is [created at genesis](https://github.com/cosmos/ibc-go/blob/v7.2.0/modules/core/03-connection/genesis.go#L23). This causes the one connection check to fail.

migueldingli1997:
Closes #670 

Quite a quick-and-dirty fix, if I'm being honest. Open to other ways of tackling this.


migueldingli1997:
I've just noticed that the scope of this is a bit larger than I expected, since interchaintest also makes use of an old version of the Go relayer, which does not seem to be compatible with the localhost client either.

boojamya:
Hey @migueldingli1997, this fix looks good to me. 
But you're right, we still need a relayer that has localhost support. Without it, I'm seeing this error:

`"no concrete type registered for type URL /ibc.lightclients.localhost.v2.ClientState against interface *exported.ClientState"`

We plan to cut a relayer release this week.
Then we can update the relayers `DefaultContainerVersion` in cosmos_relayer.go. 

I've confirmed that conformance will pass with a relayer that supports localhost with this matrix file:

```
{
  "Relayers": ["rly"],

  "ChainSets": [
    [
      {
        "Name": "ibc-go-simd",
        "Version": "v7.2.0"
      },
      {
        "Name": "ibc-go-simd",
        "Version": "v7.2.0"
      }
    ]
  ]
}
```

Thanks for the PR, we'll get this merged soon.

gjermundgaraba:
Btw: just tested this change with the latest version of rly (2.4.0) and it works for IBC 7.2 with localhost client.

boojamya (approved):

PR #678, opened 2023-07-26
e2e arbitrary passthrough fix

agouin:
Testing https://github.com/cosmos/ibc-apps/pull/71 PFM fix


Reecepbcups (approved):
My only question is if we really do need complex PFM test for examples with ictest?

I would feel its better to test this complexity in just in ibc-apps, and have a single hop transfer to show it works here, to eliminate flakey test (maybe using Juno's repo vs gaia is better, will see)

jtieri:
> My only question is if we really do need complex PFM test for examples with ictest?
> 
> I would feel its better to test this complexity in just in ibc-apps, and have a single hop transfer to show it works here, to eliminate flakey test (maybe using Juno's repo vs gaia is better, will see)

I'm actually all in favor of removing some of these examples that have now moved into other repos closer to the code bases that they seek to exercise. Providing links to complicated interchaintest examples in the wild somewhere in the docs should be sufficient

Reecepbcups:
Per the above comment and #772, closing

PR #679, opened 2023-07-29
refactor: use cosmos sdk Int type for balances/token amounts

jtieri:
Previously we were using `int64` to represent token amounts for things like transfers and balance checks. This becomes problematic when dealing with the EVM or chains like Penumbra where we may need to handle `int128` or `int256`. 

Since Go does not have native support for integer values larger than 64-bits we would need to use the stdlib's [big package](https://pkg.go.dev/math/big#Int). The Cosmos SDK's math package contains its own [Int](https://github.com/cosmos/cosmos-sdk/blob/8aaa776f822392bd506feb16cf454709059265fa/math/int.go#L75-L80) type that is a wrapper around Go's `big.Int` and offers useful helper methods and enforces a 257 bit range by checking overflow, underflow and division by zero.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -720,7 +720,7 @@ func (tn *ChainNode) SendIBCTransfer(
 ) (string, error) {
 	command := []string{
 		"ibc-transfer", "transfer", "transfer", channelID,
-		amount.Address, fmt.Sprintf("%d%s", amount.Amount, amount.Denom),
+		amount.Address, fmt.Sprintf("%s%s", amount.Amount.String(), amount.Denom),
 	}
 	if options.Timeout != nil {
 		if options.Timeout.NanoSeconds > 0 {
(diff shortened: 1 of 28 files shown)


boojamya (approved):
Nice! Looks good. 
A few nits on the zero int, but that's it!

boojamya:
Also, thinking we should probably backport this. 
... for consistency and for ease of future backports.

jtieri:
> Also, thinking we should probably backport this. ... for consistency and for ease of future backports.

I agree, we will definitely want to port this across the supported versions that we maintain. This is going to break things for anyone using `interchaintest` when upgrading though, since we don't cut releases here I'm wondering what would be the most effective way of communicating this breaking change to everyone downstream.

boojamya:
That's a good call out. 
The answer is we don't have any sort of system for that in place atm. 

FWIW, I made `breaking` label. 
My only idea is to make a changelog. But I'm debating if that is necessary.

PR #687, opened 2023-08-09
feat(cosmos): Additional Helpers

nivasan1:
## In This PR
 - I add functionality for retrieving the ContainerID of tendermint nodes in a cosmos-chain
      - This is useful for copying data / directories from specific nodes
 - I add functionality for Modifying the genesis delegation size / amounts for cosmos chains
      - This is necessary for chains that support tokens scaled by 1e18 (evm-chains)

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -146,6 +146,10 @@ func (tn *ChainNode) Name() string {
 	return fmt.Sprintf("%s-%s-%d-%s", tn.Chain.Config().ChainID, nodeType, tn.Index, dockerutil.SanitizeContainerName(tn.TestName))
 }
 
+func (tn *ChainNode) ContainerID() string {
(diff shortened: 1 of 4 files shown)


agouin (approved):
thank you!

chain/cosmos/chain_node.go
```diff
@@ -665,7 +669,7 @@ func (tn *ChainNode) AddGenesisAccount(ctx context.Context, address string, gene
 		command = append(command, "genesis")
 	}
 
-	command = append(command, "add-genesis-account", address, amount)
+	command = append(command, "add-genesis-account", address, amount, "--chain-id", tn.Chain.Config().ChainID)
```
boojamya, line 672: Is `chain-id` necessary here?

Some tests are showing this flag is not available.

boojamya (approved):

PR #692, opened 2023-08-11
feat: ability to pause and resume containers

debendraoli:
Halting container is helpful to check relay behaviour.


relayer/docker.go
```diff
@@ -445,6 +446,20 @@ func (r *DockerRelayer) StopRelayer(ctx context.Context, rep ibc.RelayerExecRepo
...
+	}
+	return r.client.ContainerPause(ctx, r.containerLifecycle.ContainerID())
+}
+
+func (r *DockerRelayer) ResumeRelayer(ctx context.Context) error {
+	if r.containerLifecycle == nil {
+		return nil
+	}
+	return r.client.ContainerUnpause(ctx, r.containerLifecycle.ContainerID())
+}
```
DavidNix, line 461: `r.containerLifecycle == nil` seems suspicious. Is nil a valid state for this struct member? Would we want to panic or return an error to fail fast?

debendraoli: yes, stop do the same, `r.containerLifecycle` is set to nil for stopped container.

boojamya (approved):
Thanks for the contribution!

PR #695, opened 2023-08-13
Update to go 1.21

faddat:
- update to go 1.21
- update ci to use go 1.21


DavidNix:
@faddat Sorry to close another PR. It's too early to upgrade to 1.21. Because interchaintest can be imported as a library, we want to support codebases which have not yet upgraded to the latest Go version. And 1.21 came out recently. 

We support the `n-1` Go version. Since 1.21 is now out, that means we can consider moving to 1.20.

faddat:
@DavidNix gotcha -- fyi for #696 -- I am not sure why it worked with 1.21 and not 1.20. 

If you have any thoughts there it would be greatly appreciated.

PR #710, opened 2023-08-16
fix: Local-interchain cleanup

DavidNix:
Fixes one compiler error. 

```
go build -ldflags '-X main.MakeFileInstallDirectory=/Users/davidnix/src/strangelove/interchaintest/local-interchain/' -o ../bin/local-ic ./cmd/local-ic
# github.com/strangelove-ventures/localinterchain/interchain
interchain/genesis.go:65:15: cannot use coin.Amount.Int64() (value of type int64) as "cosmossdk.io/math".Int value in struct literal
make: *** [Makefile:11: build] Error 1
```


local-interchain/cmd/local-ic/new_chain.go
```diff
@@ -185,7 +185,3 @@ func getOrDefault[T any](output string, defaultVal T) T {
 
 	return any(text).(T)
 }
-
-func init() {
-	rootCmd.AddCommand(newChainCmd)
-}
```
DavidNix, line 191: I avoid `init()` functions at all costs. Even though this is common in Viper and co-located with the subcommand file, I think it's an anti-pattern. It hides order of operations and hides functionality in the `main()` function. IOW, I want to look at `main()` and know exactly what's happening.

local-interchain/cmd/local-ic/root.go
```diff
@@ -9,10 +9,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	MakeFileInstallDirectory string
-)
```
DavidNix, line 14: Avoid global vars. In this case, wasn't needed at all.

local-interchain/interchain/genesis.go
```diff
@@ -62,7 +63,7 @@ func SetupGenesisWallets(config *types.Config, chains []ibc.Chain) map[ibc.Chain
 			for _, coin := range amount {
 				additionalWallets[chainObj] = append(additionalWallets[chainObj], ibc.WalletAmount{
 					Address: acc.Address,
-					Amount:  coin.Amount.Int64(),
+					Amount:  math.NewInt(coin.Amount.Int64()),
```
boojamya, line 66: Ah nice. Yea this makes sense because this got merged first:

https://github.com/strangelove-ventures/interchaintest/pull/679

Really surprised CI didn't catch this??

boojamya: Oh one more note... around backporting.
That PR is not backported yet: https://github.com/strangelove-ventures/interchaintest/pull/685

DavidNix: Oh good catch. We should get that back ported. I can look at it.

DavidNix:
> Thank you for the context around init.

Of course! Consider init functions harmful. They modify or create global state and their order of operations is not intuitive.

local-interchain/cmd/local-ic/root.go
Reecepbcups: This is required for the ldflags in `makefile` to work? or does the var in get directory work properly when you `make install`. This is the reason it is a globalvar

Reecepbcups (approved):
@DavidNix I reverted your change here to make the variable private (b87ff712f4c5523ef1078e18e73456e921493af8). The reason this is public is for the makefile to build to the proper directory (static link within the bin at compile time)

```make
ldflags = -X main.MakeFileInstallDirectory=$(CWD)
...
.PHONY: install
install:
	go install $(BUILD_FLAGS) ./cmd/local-ic ./interchain
```

Other than that it LGTM

Issue #673, opened 2023-07-21
Write docs for using `ictest` for E2E integration testing


boojamya:

boojamya:
Closing as duplicate issue: https://github.com/strangelove-ventures/interchaintest/issues/371

PR #718, opened 2023-08-22
Add ENV to show logs on fail, always, and never

### LINKED ISSUE #422: Allow logging outputted to stdout upon test failure to be disabled
Referencing [this code](https://github.com/strangelove-ventures/interchaintest/blob/main/internal/dockerutil/setup.go#L130).

I would like it to be only:
```go
if showContainerLogs {
```

or add another env variable to disable logging in the fail case.

We added logs to be uploaded upon failed gh action. It grabs all the node logs before the interchain test `dockerCleanup()` is called and dumps it into a directory which is then upload to the gh action. We noticed that a lot of the time, the tail printed to the stdout on the gh action is actually not included in the uploaded log file since it simply logs the shut down of the node after the failure. This causes a lot of noise when viewing a failed test, [see here](https://github.com/cosmos/ibc-go/actions/runs/4376766644/jobs/7659509842).

I would like to see only the initial failure and then open the full log file if I want more information

Reecepbcups:
Closes #422


internal/dockerutil/setup.go
```diff
@@ -92,6 +92,7 @@ func DockerSetup(t DockerSetupTestingT) (*client.Client, string) {
 func dockerCleanup(t DockerSetupTestingT, cli *client.Client) func() {
 	return func() {
 		showContainerLogs := os.Getenv("SHOW_CONTAINER_LOGS") != ""
```
Reecepbcups, line 94: Note: these env variables should be included in #673 docs for ictest

boojamya:
With the inclusion of https://github.com/strangelove-ventures/interchaintest/pull/725, (the part where it's not just dumping the cleanup phase of the nodes), I'm in the position where log dumps should be "on"  by default and opt out if the user prefers.

internal/dockerutil/setup.go
boojamya: Yes. Good call out. I started a doc for env vars in #725

internal/dockerutil/setup.go
```diff
@@ -92,6 +92,7 @@ func DockerSetup(t DockerSetupTestingT) (*client.Client, string) {
...
+		failedLogging := os.Getenv("DISABLE_FAILED_LOGGING") == ""
 		containerLogTail := os.Getenv("CONTAINER_LOG_TAIL")
```
boojamya, line 96: Ok, this is me being really nit picky I'm sorry! But these two ENV's do really similar things, but code readability it's really confusing. 

I also don't think this fix is necessary anymore. But if we want this feature it how about...

Maybe instead of just checking if `SHOW_CONTAINER_LOGS` is "", we check:
- "" -> default which only shows logs on failure
- "always" -> always shows logs, both for success and failure
- "never" -> never shows logs

Reecepbcups: I like this approach much more, will do this & add to docs

Reecepbcups: Completed in fe3d5d06e15780bfddc2660d9890259ce7087a0c

boojamya (approved):
BEAUTIFUL

PR #724, opened 2023-09-01
upgrade to go 1.21

faddat:
this pr:

* upgrades to go 1.21 which is required for sdk 50 and ibc 8
* upgrades to cosmos-sdk v0.47.5
* upgrades to ibc-go v7.3.0


agouin:
We'd like to keep v7 on 1.19 so that we can accomodate the largest number of chains, e.g. not require consumers of interchaintest-as-a-library to upgrade. We have not yet branched a v7, so can you please remove the go 1.21 upgrade from this? After that, we will gladly merge

faddat:
Ah okay, in that case I think I should close this, the dependabot PRs should work for the rest.

Issue #400, opened 2023-02-16
Interchaintest should have an option to "Keep Environment Alive After Test Completion".


boojamya:
Currently, after a test has run, `interchaintest` closes down all docker containers taking down the testnet. 

There should be an option to keep the environment alive to allow exploration of nodes.

chatton:
This would be a great addition! I wonder if [this mechanism](https://github.com/strangelove-ventures/interchaintest/blob/5ad0cec216e712fd3832ca2e65fc7a5cc1b45b8f/test_setup.go#L23-L31) could be expanded to allow for it?

PR #725, opened 2023-09-03
Add ENV variable option to keep testnet alive after test completion

boojamya:
This PR,

Adds the `KEEP_CONTAINTERS` env variable as an option to not tear down the testnet after completion.

This allows users to shell into the docker containers to inspect and execute upon the testnet.

I should also note that this moves the docker log dump (only happens on failed tests) to happen BEFORE the cleanup phase. This is on purpose. 
By default, the log dump logs the last 50 lines. This is great except that the cleanup phase usually logs 50+ lines. Therefore you never actually get to see the logs leading up to the failure.
This is also discussed here https://github.com/strangelove-ventures/interchaintest/issues/422.
Happy to move this back if others disagree.

Closes https://github.com/strangelove-ventures/interchaintest/issues/400


internal/dockerutil/setup.go
```diff
@@ -145,21 +123,50 @@ func dockerCleanup(t DockerSetupTestingT, cli *client.Client) func() {
 					b := new(bytes.Buffer)
 					_, err := b.ReadFrom(rc)
 					if err == nil {
-						t.Logf("Container logs - {%s}\n%s", strings.Join(c.Names, " "), b.String())
+						t.Logf("\n\nContainer logs - {%s}\n%s", strings.Join(c.Names, " "), b.String())
```
jtieri, line 126: why do we need the two newlines at the beginning of this log now?

internal/dockerutil/setup.go
```diff
@@ -145,21 +123,50 @@ func dockerCleanup(t DockerSetupTestingT, cli *client.Client) func() {
...
 			}
 		}
 
-		pruneVolumesWithRetry(ctx, t, cli)
-		pruneNetworksWithRetry(ctx, t, cli)
+		if !keepContainers {
+			pruneVolumesWithRetry(ctx, t, cli)
+			pruneNetworksWithRetry(ctx, t, cli)
+		} else {
+			t.Logf("KEEPING CONTAINERS - DOCKER CLEANUP SKIPPED")
```
jtieri, line 168: to be consistent with all other logging can we change this line so that it does not use all caps and follows similar conventions e.g. capitalize first letter of first word, no punctuation at the end, etc

jtieri (changes requested):
had a question and one nit but otherwise this looks good

internal/dockerutil/setup.go
boojamya: This is purely for readability. Without this, the log dump between each node bud up against each other and it is difficult to see where the logs from one node end and the other begins.

internal/dockerutil/setup.go
boojamya: Done!

jtieri (approved):

PR #716, opened 2023-08-21
sdk 50 v2

faddat:
This PR updates interchaintest to sdk v50.

PR #736, opened 2023-09-06
SDK 47+ unsafe-reset w/ backwards compatibility

Reecepbcups:
Per [Colin's comment](https://github.com/strangelove-ventures/interchaintest/pull/716#issuecomment-1705476348)

## Summary
With this, it is now possible to unsafe reset all in the legacy SDK (v45, 46) while also maintaining support for the newer `comet` namespace to reset all.

Since the new genesis command was also introduced at the same time, we just check that if it's on.

## Future considerations
We may want to change `UsingNewGenesisCommand` to be `UsingOldGenesisCommand` to opt in for older SDK versions (since all future versions will use this new format)


boojamya (approved):
LGTM.

Also, totally agree we should reword `UsingNewGenesisCommand`
Issue created: https://github.com/strangelove-ventures/interchaintest/issues/742

Reecepbcups:
@Mergifyio backport v7

PR #737, opened 2023-09-06
local-ic: Update to SDK v50

Reecepbcups:
Per [Dan K's comment](https://github.com/strangelove-ventures/interchaintest/pull/716#issuecomment-1703254210)

## Summary
- Adds local-ic build CI
- Updates local-ic to SDK v50 (uses a go quirk which replaces a null commit to parent)


.github/workflows/local-interchain.yaml
```diff
@@ -0,0 +1,30 @@
+name: Local Interchain
+
+on:
+  pull_request:
+  push:
+    branches:
+      - "main"
```
boojamya, line 7: If we are running on PR's, I don't think this is necessary.

local-interchain/interchain/genesis.go
```diff
@@ -62,7 +62,7 @@ func SetupGenesisWallets(config *types.Config, chains []ibc.Chain) map[ibc.Chain
 			for _, coin := range amount {
 				additionalWallets[chainObj] = append(additionalWallets[chainObj], ibc.WalletAmount{
 					Address: acc.Address,
-					Amount:  coin.Amount.Int64(),
+					Amount:  coin.Amount,
```
DavidNix, line 65: Are we sure this compiles? IIRC, I had to add the int64 method to fix a compile error. Ideally, CI catches any compile problems but it was missed last time.

DavidNix: The new GH action should catch any compile errors and fail.

.github/workflows/local-interchain.yaml
Reecepbcups: Resolved in dafd9c610d57cfb9fe8f144b1d67cc41690141aa

local-interchain/interchain/genesis.go
boojamya: FYI, this is the PR that made the change Nix is referring to. 
https://github.com/strangelove-ventures/interchaintest/pull/679

Off the top of my head, I'm not sure how this affects local-ic.

Reecepbcups: It was just a compile error, no logic broke. This PR fixes that and also ensures it does not happen again (build CI in this branch)

boojamya (approved):
I think it would be wise to get these changes backported into the `v7` branch. Specifically the workflow.

It's probably easier to do this with a separate PR rather than using mergify. But your call @Reecepbcups.

Reecepbcups:
@boojamya I will do manually now

PR #738, opened 2023-09-06
small bumps before v8 module paths

faddat:
This PR contains a set of small bumps ot be used before changing the module paths for ibc-go to github.com/cosmos-sdk/ibc-go/v8


Reecepbcups:
These are handled in #741. While typically I would push through the earlier PR, that one also bumps to IBC-v8 along side it

PR #740, opened 2023-09-06
Cleaner ExportState via document for SDK v47+

### LINKED ISSUE #721: Export to a document instead of reading from stdout
When calling `chain.ExportState`, log entries are being returned. I believe this might be because the stdout from [`node.ExecBin`](https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/chain_node.go#L1067) is being returned. I recommend using the `--output-document` flag so that the binary writes the exported state to a document and then that document is read and returning to the caller. This should hopefully avoid intermixing exports with logs

Reecepbcups:
Closes #721 

## Summary
If a chain is on SDK v47+, we should output exports to a file directly. Then we return the contents from this file instead of returning a messy stdout + err for parsing.

Tests both 45 and 47 usecases

### DIFF
--- a/examples/cosmos/chain_export_test.go
+++ b/examples/cosmos/chain_export_test.go
@@ -1,0 +1,118 @@
+package cosmos_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/strangelove-ventures/interchaintest/v8"
+	"github.com/strangelove-ventures/interchaintest/v8/chain/cosmos"
+	"github.com/strangelove-ventures/interchaintest/v8/ibc"
+	"github.com/strangelove-ventures/interchaintest/v8/testutil"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestJunoStateExport(t *testing.T) {
+	// SDK v45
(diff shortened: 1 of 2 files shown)


examples/cosmos/chain_export_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	err = chain.StopAllNodes(ctx)
+	require.NoError(t, err, "error stopping node(s)")
+
+	state, err := chain.ExportState(ctx, int64(haltHeight))
+	require.NoError(t, err, "error exporting state")
+
+	appTomlOverrides := make(testutil.Toml)
+
+	appTomlOverrides["halt-height"] = 0
```
boojamya, line 85: Can you explain why this is necessary? `halt-height` defaults to `0` yea? Does it get overwritten when a node is stopped?

boojamya (approved):
👏

examples/cosmos/chain_export_test.go
Reecepbcups: Oops! not necessary. Has no effect for this test. Removed in 4b6cadaca4a55f73848943d16e8e6bd8f94bd0c9

Reecepbcups:
@Mergifyio backport v7

PR #759, opened 2023-09-08
Remove `UsingNewGenesisCommand` config in favor of a runtime check

### LINKED ISSUE #742: Refactor `UsingNewGenesisCommand`
Specifically for the `SDK 50`, this is not new anymore and is actually the default.

We should change this to be something like `UsingLegacyGenesisCommand`

Reecepbcups:
Closes #742 

## summary
Removing the config option `UsingNewGenesisCommand` makes it easier to write test spanning multiple SDK versions (v45 - 50).


chain/cosmos/chain_node.go
```diff
@@ -1066,13 +1072,10 @@ func (tn *ChainNode) ExportState(ctx context.Context, height int64) (string, err
 	var (
 		doc     = "state_export.json"
 		docPath = path.Join(tn.HomeDir(), doc)
-
-		isNewGenesis = tn.Chain.Config().UsingNewGenesisCommand
-
 		command = []string{"export", "--height", fmt.Sprint(height), "--home", tn.HomeDir()}
 	)
 
-	if isNewGenesis {
```
agouin, line 1075: we can keep the var here so the lookup doesn't need to be performed twice within this method

examples/cosmos/sdk_boundary_test.go
```diff
@@ -37,9 +37,7 @@ func TestSDKBoundaries(t *testing.T) {
 				},
 				{
 					Name: "ibc-go-simd", ChainName: "simd-50", Version: "feat-upgrade-sdk-v0.50", //sdk 0.50 alpha
-					ChainConfig: ibc.ChainConfig{
-						UsingNewGenesisCommand: true,
-					},
+					ChainConfig: ibc.ChainConfig{},
```
agouin, line 40: The `ChainConfig` can be removed entirely now since there are no overrides

examples/cosmos/sdk_boundary_test.go
```diff
@@ -52,9 +50,7 @@ func TestSDKBoundaries(t *testing.T) {
...
```
agouin, line 53: same here

chain/cosmos/chain_node.go
```diff
@@ -830,10 +840,6 @@ func (tn *ChainNode) HasCommand(ctx context.Context, command ...string) bool {
 		return true
 	}
 
-	if err != nil {
-		return false
-	}
-
 	return true
```
agouin, line 843: should we `return false` here?

Reecepbcups: Yes good catch since I moved logic around, resolved 0ff41686af48eb81a4c2b57bb9152da891ffc1b7

examples/cosmos/sdk_boundary_test.go
Reecepbcups: Resolved: def97dfbd2b16619714919dfa9b0b0a17ee071bb

examples/cosmos/sdk_boundary_test.go
Reecepbcups: Resolved: def97dfbd2b16619714919dfa9b0b0a17ee071bb

chain/cosmos/chain_node.go
Reecepbcups: 27417615232c08a5ff32f5a83bee5ea6259383cd

agouin (approved):
:rocket:

PR #766, opened 2023-09-10
Adopt the function options pattern for relayer configuration

### LINKED ISSUE #403: Consider adopting the functional options pattern for advanced relayer configuration
Currently, there is an interface [RelayerOption](https://github.com/strangelove-ventures/interchaintest/blob/be8adbfb958f802b1f988f79a240ced032eaf930/relayer/options.go#L9) which defines an unexported function which is a no-op in in the implementations (I believe to narrow the set of types which implement this interface). 

This implementation means that the [exact types](https://github.com/strangelove-ventures/interchaintest/blob/5f573c9f63a76de925747ed023793d640ab19a07/relayer/docker.go#L75-L84) need to be known already before being able to apply configuration.

I think the functional options pattern would greatly simplify the workflow of configuring custom fields and also be more flexible in that it does not require the implementation (DockerRelayer) to specifically know about each type of configuration option.

```go
package relayer

import (
	"github.com/strangelove-ventures/interchaintest/v6/ibc"
)

type RelayerOpt func(relayer *DockerRelayer) // could act on docker relayer directly or some additional options type to be more flexible with non-docker implementations.

func CustomDockerImage(repository string, version string, uidGid string) RelayerOpt {
	return func(r *DockerRelayer) {
		r.customImage.Repository = repository
		r.customImage.Version = version
		r.customImage.UidGid = uidGid
	}
}

func HomeDir(homeDir string) RelayerOpt {
	return func(r *DockerRelayer) {
		r.homeDir = homeDir
	}
}


```

Sample usage could look like

```go
func NewDockerRelayer(ctx context.Context, log *zap.Logger, testName string, cli *client.Client, networkID string, c RelayerCommander, options ...RelayerOpt) (*DockerRelayer, error) {
	r := DockerRelayer{
           // default options
	}
        // any arbitrary configuration can happen here.
	for _, opt := range options {
		opt(&r)
	}
        //  rest of setup ... 
}
```

Reecepbcups:
Closes #403 

## summary
This moves to a more generic options configuration for setting up a relayer.


boojamya (approved):
💅

PR #772, opened 2023-09-12
Add more testing, remove redundant

### LINKED ISSUE #764: Add more chain examples & further testing with it
Showcase the features of ICtest that are in chain_node and other useful in other chains. Simplify duplicate code

This doubles as testing which we need anyways

Reecepbcups:
Closes #764 & #794

## summary
Adds more testing for critical pieces of which are not used anywhere else in interchaintest.
Test removed (ex: PFM) are linked to other outside repos

## other minor additions
- cosmos.SetSDKConfig(bech32) now sets the config of a running container for proper addr parsing
- RangeBlockMessages was made public
- NewGenesisKV for cleaner genesis items

## what this does not include
- gov proposals (sdk v50 parse issues, wasmd does not yet support)
- wasm (steve's Compile PR handles)


examples/cosmos/chain_miscellaneous_test.go
```diff
@@ -98,3 +127,204 @@ func BuildDependencies(ctx context.Context, t *testing.T, chain *cosmos.CosmosCh
...
+			Address: addr2,
+			Coins:   c2,
+		},
+	}
+
+	txResp, err := cosmos.BroadcastTx(
+		ctx,
+		b,
+		users[0],
+		banktypes.NewMsgMultiSend(in, out),
```
boojamya, line 272: Just for my own knowledge, could you explain a use case for this NewMultiMsgSend?

boojamya (approved):
Huge cleanup!
The tests added in miscellaneous are what we needed. 

Hopefully we can get those removed tests that you created gists for into ibc-apps sooner than later. 

One question for my own sanity, but approved.

examples/cosmos/chain_miscellaneous_test.go
Reecepbcups: This is useful in the future to use in the faucet.

I just wanted to ensure we tested non standard methods here for the broadcaster, and we do not yet use a MsgMultiSend anywhere. Just a good example for others on how to do it

Issue #763, opened 2023-09-08
CI to test in other languages: Python & Rust


Reecepbcups:
With the addition of #761 and #762 we need an example showing how to run test from other languages.

Successful completion of this issue would launch 2 IBC chains, relay between them, and execute on a contract forwarding data all through Python or the TS Clients.

# TODO
- [ ] Python (#775)
- [x] Rust (#785)

akc2267:
Rust client too

PR #775, opened 2023-09-12
local-ic: CI test with Python

Reecepbcups:
closes: #763 

## summary
Adds a python based local-interchain test CI pipeline

## features
- python CI pipeline w/ relaying
- addr & port override flags
- python poll_for_start utility
- moved from ./scripts to ./python

## todo
- [x] Test: contract<->contract relayer & api interaction assertions
- [x] Upload artifact, then run each test on its own.
- [x] Deterministically randomize the port for the server (both localic & script) so no runner overlap
- [x] Move Python to its own section
- [x] Re-add CW cache


Reecepbcups:
I am putting PR #785 ahead, so put this back on draft. Will fix after getting that in since it is the more feature complete version now

local-interchain/python/api_test.py
```diff
@@ -0,0 +1,119 @@
...
+        "fee-payer": "",
+        "note": "",
+    }
+
+    assert len(rb.binary("keys list --output=json")) > 0
+
+    # this query returns a dict with suply as the key, validate it is there
+    assert "supply" in rb.query("bank total")
+
+    rb.query("bank balances juno10r39fueph9fq7a6lgswu4zdsg8t3gxlq670lt0 --output=json")
```
boojamya, line 98: Is there supposed to be an assert after this?

Reecepbcups: Yes! Good catch

boojamya (approved):

PR #782, opened 2023-09-14
feat: cosmwasm contract build flow during e2e test execution

### LINKED ISSUE #370: Smart contract developer should have a CosmWasm build flow which allows easy re-running w/o manual steps.
It is currently possible to store, instantiate, and execute wasm smart contracts on a `CosmosChain` with the `StoreContract`, `InstantiateContract`, and `ExecuteContract` methods. `StoreContract` requires a `fileName` to a compiled wasm binary in order to upload the contract to the chain.

To enable a smart contract developer flow, it would be great to have a method to compile wasm contract source code so that the developer could modify the smart contract and re-run the test case without having to manually re-compile the binary and copy/move it to the correct path in between.

This will also allow contract source code to be committed with the test cases instead of compiled binaries.

example:

```golang
func (c *CosmosChain) CompileContract(
  workingDirectory string, // working directory of the smart contract project on local machine
  buildCommand string, // command to run to compile binary e.g. `cargo wasm`
  binaryPath string, // relative path to compiled binary within working directory inside docker container
  outputPath string, // path to store binary on local machine
) error {
  // launch docker container from a wasm build docker image

  // copy `workingDirectory` contents from
  // local machine to build folder inside container

  // execute `buildCommand`

  // use programmatic docker copy to copy compiled binary 
  // from `binaryPath` inside the container to `outputPath` on local machine

  return nil
}
```

misko9:
Adds a new `contract/cosmwasm` package to interchaintest. This package add support for compiling contract(s) during test execution. Both `cosmwasm/rust-optimizer` and `cosmwasm/workspace-optimizer` are supported.

**Single contract / rust-optimizer:**
1. early in test, call `cosmwasm.NewContract(relativePathOfContract).Compile()`, returns contract object
2. before `StoreContract()`, call `contract.WaitForCompile()`, returns wasm binary path (StoreContract input) or error

**Multiple/workspace contracts / workspace-optimizer:**
1. early in test, call `cosmwasm.NewWorkspace(relativePathOfWorkspace).Compile()`, returns workspace object
2. before `StoreContract()`, call `workspace.WaitForCompile()`, returns map of cargo names to wasm binary paths (StoreContract input) or error

**The examples/cosmwasm/ directory shows two very simple use cases:**
1. rust-optimizer for single contracts
2. workspace-optimizer for multiple contracts

Closes #370


contract/cosmwasm/compile.go
```diff
@@ -0,0 +1,119 @@
...
+		Mounts: []mount.Mount{
+			{
+				Type: mount.TypeBind,
+				Source: repoPathFull,
+				Target: "/code",
+			},
+			{
+				Type: mount.TypeVolume,
+				Source: filepath.Base(repoPathFull)+"_cache",
+				Target: "/target",
```
Reecepbcups, line 66: Just a note: this will break contracts using rust-optimizer v0.12.13 and below *(target moved from /code/target to /target in v13)*.

I mention this as I see many still building with the v0.12 suite in CI, including myself just last week.

contract/cosmwasm/rust_optimizer.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		// Form the path to the wasm binary
+		c.wasmBinPathChan <- filepath.Join(artifactsPath, strings.TrimSpace(wasmBin))
+	}()
+
+	return c
+}
+
+// WaitForCompile will wait until coyympilation is complete, this can be called after chain setup
```
Reecepbcups, line 76:
```suggestion
// WaitForCompile will wait until compilation is complete, this can be called after chain setup
```

Reecepbcups (approved):
Great work here!

contract/cosmwasm/compile.go
misko9: Thanks, updated!

PR #790, opened 2023-09-19
chore: move localhost -> 127.0.0.1

### LINKED ISSUE #402: Intermittent failures occur during chain startup

Reecepbcups:
closes #402

## summary
This PR moves any tcp references to localhost, and instead uses the direct IP, 127.0.0.1, in its place.

reasons:
- the /etc/hosts lookup itself is more sys calls when it is not required for an ipv4 enviroment
- localhost can reference multiple IP families (ipv6 & ipv4) on lookup. - This could cause unintended issues (overhead: query ipv6, fail, then try ipv4. Machines not routing and only doing v6)

### DIFF
--- a/internal/dockerutil/ports.go
+++ b/internal/dockerutil/ports.go
@@ -20,7 +20,7 @@ func (l Listeners) CloseAll() {
 
 // openListenerOnFreePort opens the next free port
 func openListenerOnFreePort() (*net.TCPListener, error) {
-	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	addr, err := net.ResolveTCPAddr("tcp", "127.0.0.1:0")
 	if err != nil {
 		return nil, err
 	}
--- a/internal/dockerutil/strings.go
+++ b/internal/dockerutil/strings.go
@@ -27,7 +27,7 @@ func GetHostPort(cont types.ContainerJSON, portID string) string {
 
 	ip := m[0].HostIP
 	if ip == "0.0.0.0" {
-		ip = "localhost"
+		ip = "127.0.0.1"
 	}
 	return net.JoinHostPort(ip, m[0].HostPort)
 }
--- a/internal/dockerutil/strings_test.go
+++ b/internal/dockerutil/strings_test.go
@@ -40,7 +40,7 @@ func TestGetHostPort(t *testing.T) {
 						},
 					},
 				},
-			}, "test", "localhost:3000",
+			}, "test", "127.0.0.1:3000",
 		},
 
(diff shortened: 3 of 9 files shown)


boojamya (approved):
Thanks for the explanation.

PR #798, opened 2023-09-25
Adds x/tokenfactory helper methods

### LINKED ISSUE #770: Add TokenFactory helper functions
Match the osmosis spec, pull from Juno repo work

main:
- burn
- change-admin
- create-denom
- mint
- modify-metadata

extra:
- burn-from
- force-transfer

Reecepbcups:
Closes #770 

## summary
Adds x/tokenfactory helper methods (matching the osmosis specification). Full coverage of new functions added

## todo
- [x] Add test


chain/cosmos/tokenfactory.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/strangelove-ventures/interchaintest/v8/ibc"
+)
+
```
jtieri, line 11: Godoc comments on all these functions could be useful. As someone not too familiar with the `tokenfactory` module it's not immediately obvious what the behavior of a few of these would be.

jtieri (approved):
I'm not very familiar with the `tokenfactory` so I can't speak to the validity of some of these but the code changes look alright. Had the one suggestion on adding comments but I don't think it's a blocker.

chain/cosmos/tokenfactory.go
Reecepbcups: Resolved in 27e80ece036f73a7fbb71659634274ae76536020

PR #810, opened 2023-10-07
add extraExecTxArgs to other CosmWasm methods

### LINKED ISSUE #809: allow override of cosmwasm commands
Something like:
```
type Override struct {
	overrideGas uint64
	overrideFee string // coin
	// amount to deposit (for execute contracts)
	// etc.
}
```

Without this I have to write methods such as ExecuteMsgWithFee, ExecuteMsgWithCoinAmount. But cannot do this with StoreContract easily. Same for InstantiateContract

Reecepbcups:
closes #809 

## summary
When writing CW test, I had to write helper methods to execute with other fees, amounts, and gas limits. This removes that for StoreCode's and contract Executes

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -779,14 +779,17 @@ type CodeInfosResponse struct {
 }
 
 // StoreContract takes a file path to smart contract and stores it on-chain. Returns the contracts code id.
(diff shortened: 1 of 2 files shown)


boojamya (approved):
Nice. Makes sense. Looks good.

PR #819, opened 2023-10-11
go workspace & deps bumps

Reecepbcups:
## summary
Adds a go workspace to make go.mod deps easier for both root & local-interchain syncing

Also bumps some other dependencies


Reecepbcups:
todo: bump to net 0.17

boojamya (approved):

PR #836, opened 2023-10-19
make `GetTransaction` public

Reecepbcups:
title

useful to decode a TxHash for debugging on test from the stdout []byte

May also add a method to easily decode Txs responses / builders (similar to my POA approach)


Reecepbcups:
Doing this in the wasm err PR

PR #838, opened 2023-10-20
fix: use ibc/v4 in v4 branch

johnletey:
When integrating the v4 release branch, the Noble team noticed that it was still using IBC v3.

This PR addresses this by using IBC v4 in the v4 release branch!


ibc/packet.go
```diff
@@ -5,15 +5,15 @@ import (
...
+	host "github.com/cosmos/ibc-go/v4/modules/core/24-host"
 	"go.uber.org/multierr"
 )
 
 type Nanoseconds uint64
 
 // Packet is a packet sent over an IBC channel as defined in ICS-4.
 // See: https://github.com/cosmos/ibc/blob/52a9094a5bc8c5275e25c19d0b2d9e6fd80ba31c/spec/core/ics-004-channel-and-packet-semantics/README.md
-// Proto defined at: github.com/cosmos/ibc-go/v3@v3.0.0/proto/ibc/core/channel/v1/tx.proto
+// Proto defined at: github.com/cosmos/ibc-go/v4@v3.0.0/proto/ibc/core/channel/v1/tx.proto
```
Reecepbcups, line 16: nit
```suggestion
// Proto defined at: github.com/cosmos/ibc-go/v4@v4.0.0/proto/ibc/core/channel/v1/tx.proto
```

Reecepbcups (approved):

ibc/packet.go
johnletey: Should we update this to even v4.5.1?

Reecepbcups: Yes! That works too

PR #840, opened 2023-10-24
sdk v50 proposal compatibility

Reecepbcups:
(used in x/POA module)

SDK v50 added more governance params in the proposal.json.

The new functions are suffixed with the IBC Go / interchaintest version (ex: BuildProposal and BuildProposalV8). Should we instead prefix with SDK versions?

## after merge todo
- tag v7 and v8 releases


Reecepbcups:
Ahhh we need to be backwards comptable for SDK v47 (v7) ->v50 (v8) upgrades

So multiple structs will be required. Suffix based off the IBC go version instead of SDK (Getv8Proposal)

chain/cosmos/cosmos_chain.go
```diff
@@ -439,7 +444,9 @@ func (c *CosmosChain) SubmitProposal(ctx context.Context, keyName string, prop T
 }
 
 // Build a gov v1 proposal type.
-func (c *CosmosChain) BuildProposal(messages []cosmosproto.Message, title, summary, metadata, depositStr string) (TxProposalv1, error) {
+//
+// The proposer field should only be set for IBC-Go v8 / SDK v50 chains.
```
boojamya, line 448: We should also add some sort of note that the `expedited` field is also not ingested for non-sdk50 chains.

boojamya (approved):
Nice, looks good!
One nit. Approving to not block.

chain/cosmos/cosmos_chain.go
Reecepbcups: this will not matter so long as depositStr is blank. Will error out if they try anything for legacy chains

boojamya: Wait how does depositStr come into play here?

Reecepbcups: oops! sorry I mis-copied `proposer`

boojamya: Copy that. It was more of a guidance thing for users. But yea, it's self explanatory, should be fine.

PR #842, opened 2023-10-25
wasm exec returns response & err

### LINKED ISSUE #823: Wasm executions do not produce errors even if the contract returns an error
When using `chain.ExecuteContract`,  we don't get an error even if the contract returns an error.

Reecepbcups:
Closes #823 

## summary
- When dealing with CW contracts, the error is nested in the RawLog. Currently we only return an Err if there is a stdout binary issue. This PR allows the CW contract to return the err if CW ahs a non 0 resp. code.
- Makes GetTransaction(txHash) public since it is useful

## considerations
after the v7 and v8 tags for ictest, I would like to do this across the entire codebase (instead of returning (txHash, err) we return (txResp, err')


chain/cosmos/chain_node.go
```diff
@@ -961,11 +961,25 @@ func (tn *ChainNode) InstantiateContract(ctx context.Context, keyName string, co
...
+		return &types.TxResponse{}, err
+	}
+
+	txResp, err := tn.GetTransaction(tn.CliContext(), txHash)
+	if err != nil {
+		return &types.TxResponse{}, fmt.Errorf("failed to get transaction %s: %w", txHash, err)
+	}
+
+	if txResp.Code != 0 {
+		return txResp, fmt.Errorf("error in transaction (code: %d): %s", txResp.Code, txResp.RawLog)
```
Reecepbcups, line 979: @srdtrk is this what you were looking for?

srdtrk: I need a v7 backport of this to try out with my test case. It'd be nice if you could provide me with the go.mod import i'd need to use

Reecepbcups: I have not tested it on v7 yet, but here is the cherry picked commit:

99ec15f2674d326aa0ab84138e189141c31b0204 - `github.com/strangelove-ventures/interchaintest/v7 v7.0.0-20231025163603-99ec15f2674d`

srdtrk: Tested it and it works!

srdtrk (approved):
Works on my machine XD

boojamya (approved):

PR #845, opened 2023-10-25
gh runners: ubuntu-latest

Reecepbcups:
issues with self hosting due to K8, removed test should let us do this again


Reecepbcups:
it may be slower than self hosted, but it works instead of 5-7 re-runs :weary:

PR #747, opened 2023-09-06
Re-add polkadot/hyperspace support and update TestHyperspace for ibc-go wasm light client changes

misko9:

PR #868, opened 2023-11-09
chore: expose validator key funcs, increase gas

johnletey:


boojamya (approved):
Thanks John!

chain/cosmos/chain_node.go
```diff
@@ -893,7 +893,7 @@ func (tn *ChainNode) QueryClientContractCode(ctx context.Context, codeHash strin
 func (tn *ChainNode) VoteOnProposal(ctx context.Context, keyName string, proposalID string, vote string) error {
 	_, err := tn.ExecTx(ctx, keyName,
 		"gov", "vote",
-		proposalID, vote, "--gas", "auto",
+		proposalID, vote, "--gas", "100000",
```
Reecepbcups, line 896: `ExecTx` under the hood uses `--gas-adjustment`. Did you attempt to increase your ibc.ChainConfig GasAdjustment? Maybe attempt 1.5 if you are on 1.3

auto should work as expected

boojamya: Good call. Thanks Reece. Overlooked this.

johnletey: I went as high as `5.0`, and it still didn't work. I can play around with it more if needed 🙌🏻

Reecepbcups: That seems odd, maybe the --gas-adjustment is not applying in the command?

check in the logs if it is showing --gas-adjustment=0 or something weird like that. or not enough fees or something

johnletey: It's not being picked up ...

Screenshot 2023-11-09 at 19 56 15

This is what I have as a configuration ...

Screenshot 2023-11-09 at 19 57 28

Reecepbcups: ahh I bet this is the issue Steve fixed in https://github.com/strangelove-ventures/interchaintest/pull/747#discussion_r1320159263 with `return int64(math.Ceil(fees))`

johnletey: What do you suggest we do?

I personally don't see why we can't hardcode the gas here. Could even make it higher at `250000` so we know it will always go through!

Reecepbcups: The `math.Ceil` patch did not work. Since this is only for v4-ics I am just going to approve the hardcoded value. Real Gas should never be above 300k for this sort of Tx anyways, so plenty of room for higher consume cost

Reecepbcups (approved):

Reecepbcups:
oh it's on self hosted that needs to be moved to ubuntu-latest before merge

PR #886, opened 2023-11-27
fix: correctly check transaction status

johnletey:
Starting in the `v0.47.x` release line, the `block` broadcast mode has been deprecated. I've updated the `ExecTx` logic to match what the SDK team [recommends](https://github.com/cosmos/cosmos-sdk/blob/main/UPGRADING.md#broadcast-mode).

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -521,17 +521,30 @@ func (tn *ChainNode) ExecTx(ctx context.Context, keyName string, command ...stri
 	if err != nil {
 		return "", err
 	}
+
 	output := CosmosTx{}
-	err = json.Unmarshal([]byte(stdout), &output)
+	err = json.Unmarshal(stdout, &output)
+	if err != nil {
+		return "", err
+	}
+
+	if err := testutil.WaitForBlocks(ctx, 1, tn); err != nil {
+		return "", err
+	}
+
+	stdout, _, err = tn.Exec(ctx, tn.QueryCommand("tx", output.TxHash), nil)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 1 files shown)


Reecepbcups:
@johnletey I am a bit confused, where is there --broadcast-mode=`block` logic per the description? interchaintest already supports `sync` broadcast & polling (tested on SDK v47 and main branch v50 chains)

I see you are re-decoding the Tx from the original Txhash to get all the data, but only the TxHash is returned so you don't get to use this extra processing

chain/cosmos/chain_node.go
```diff
@@ -521,17 +521,30 @@ func (tn *ChainNode) ExecTx(ctx context.Context, keyName string, command ...stri
...
 	output := CosmosTx{}
-	err = json.Unmarshal([]byte(stdout), &output)
+	err = json.Unmarshal(stdout, &output)
 	if err != nil {
 		return "", err
 	}
-	if output.Code != 0 {
-		return output.TxHash, fmt.Errorf("transaction failed with code %d: %s", output.Code, output.RawLog)
+
+	if err := testutil.WaitForBlocks(ctx, 1, tn); err != nil {
```
Reecepbcups, line 531: this should also remain at 2. Moving to 1 causes issues in multi val environments due to docker + tendermint

Reecepbcups:
Closing as the v7 branch matches v8 - SDK v47 and SDK v50 chains work as expected and use the `sync` mode during broadcast. Make sure any custom commands or config settings do not use the legacy `block` as mentioned in my above comment.

if the issue persist please re-tag me and I will look into it more. The proposed changes here would not resolve that even if broadcast-mode=block was used.

PR #890, opened 2023-11-30
Test Hermes Relayer during boundary test (v7)

boojamya:
With https://github.com/strangelove-ventures/interchaintest/pull/889, I realized we do have any e2e tests in the CI that test Hermes.

I figured the best place for this was in the boundary test.

I also removed `sdk_boundary_test.go` as that is only meant for `main/v8` branch.


boojamya:
Closing. 
I am just now realizing we do indeed test both relayers. 

I forgot we have CI setup to test conformance with the `example_matrix.json` file. Both relayers are passed in here.
We are covered, this is unnecessary.

PR #906, opened 2023-12-14
chore: spell check CI

Reecepbcups:


Reecepbcups:
hmm this does not catch words that use _ or other programming techniques. Only simple comments and single words...

need something deeper

Reecepbcups:
We do need this but it is not the right CI for the task. Need something with more control and options

Issue #795, opened 2023-09-22
EPIC: `local-ic` misc issue tracking ticket


Reecepbcups:
```[tasklist]
### Tasks
- [x] make install for go releaser instead of go install ./... so version is set https://github.com/strangelove-ventures/interchaintest/pull/1149
- [ ] [Mainnet interaction across networks](https://github.com/strangelove-ventures/interchaintest/pull/291) (cosmos dir RPCs),
- [x] Drivers: Bash (get/post examples, then allow curl down + [source from url](https://superuser.com/questions/255260/bash-source-from-url)) https://github.com/strangelove-ventures/interchaintest/pull/1207
- [ ] More IBC Abilities (Ex: ICA, Mesh Security, Polytone). Pending https://github.com/strangelove-ventures/interchaintest/issues/1094
- [ ] Support upgrades _(spawn a new fullnode & wait on that, not the validator else panic)_
- [ ] Allow mounting (ref main Ethereum chain start)
- [ ] Auto set keyring-backend=test for local-ic config
- [ ] pipe logs to a file, don't block on the end. On exit print out where (pipe logs to a file, don't block on the end. On exit print out where). Stream tailed logs from the container (docker logs -f -n 100 or something?)
- [x] on ctrl+c remove all running containers https://github.com/strangelove-ventures/interchaintest/pull/1206
- [x] Easy CW upload & init from within interaction
- [ ] Fix multiple tokens spreading out over many commands (some networks like osmosis do not have --append support)
- [ ] How to launch testnets (public & private). [ref spawn for CI](https://github.com/rollchains/spawn/blob/release/v0.50/simapp/.github/workflows/testnet-hetzner.yml)
- [ ] More docs on interaction (req by Steve & Ollie @ ICF)
- [ ] Polkachu snapshots -> mainnet with v50 in place testnets
- [ ] Add service file which we did for ethos for running
- [x] Docker compose support (running alongside), and sidecars
- [ ] https://github.com/strangelove-ventures/interchaintest/issues/1182
- [ ] [Kubernetes / Gitlab CI integration](https://github.com/strangelove-ventures/interchaintest/issues/1075#issuecomment-2284314398) (Done, just need to write docs / verify)
- [ ] use as bases possible https://github.com/strangelove-ventures/interchaintest/pull/1219#discussion_r1725972315
- [ ] add thorchain / utxo chain support
[...]

PR #908, opened 2023-12-14
feat!: local-ic remove relayer file dependency

Reecepbcups:
ref #795

Previously it was required for developers to maintain a relayer.json file with the image information. For most use cases this is excessive as the defaults work. Instead it should be a startup option to override default relayer image information.

If none are specific, the latest cosmos relayer image is used.

Only a breaking change if a team uses a custom relayer image for local-ic


local-interchain/python/helpers/file_cache.py
```diff
@@ -33,6 +34,7 @@ def get_chain_start_time_from_logs() -> int:
     def get_cache_or_default(contracts: dict, ictest_chain_start: int) -> dict:
         cache_time: str | int = 0
         if os.path.exists(contracts_json_path):
+            os.makedirs(os.path.dirname(contracts_json_path), exist_ok=True)
```
jtieri, line 37: i don't Python so can't speak to these changes but i will assume these two lines are correct

jtieri (approved):
changes look straightforward enough. 

what's up with the empty `.gitkeep` file? is there some assumption that this empty directory needs to exist for tests to run or something?

local-interchain/python/helpers/file_cache.py
Reecepbcups: yes, just creates a nested directory if it does not exist. if it does, it does nothing

Reecepbcups:
let's see

PR #917, opened 2023-12-19
feat: host port overrides

Reecepbcups:
@agouin brought up a good point for LocalInterchain, the host ports are always random when using interchaintest instead of deterministic. When spinning up infra as a dev/testnet, it would be nice to keep these standard (1317, 26657,...) if specified in the ibc.ChainConfig
 
This allows a map of map[internal]external port int's to force the host mapping.

Using the updated base.json example:
image


Reecepbcups:
closed in favor of PR 921 (Eth support) which uses this)

PR #918, opened 2023-12-19
Add anvil/ethereum support

misko9:
This PR allows spinning up an anvil ethereum chain with the minimum requirements for GetAndFundTestUser(). For Ethos, it also allows a state.json file to be loaded on start and the GetHostRPCAddress()/GetHostWSAddress() will then be used for their applications.


Reecepbcups (approved):
Amazing! :ship:

PR #919, opened 2023-12-20
feat: faucet api, auth key, GetHostPeerAddress

Reecepbcups:
## Summary
- **opens up ICtest to use 0.0.0.0 (all interfaces) instead of only localhost**. I am worried about this so others inputs here would be great. Seems fine for GH actions and locally running. Required for public testnets such as Ethos
- `--auth-key` start flag for permissioned/internal API access only
- faucet endpoint to fund accounts
- `GetHostPeerAddress` to query the chain and sync external node


Reecepbcups:
This allowed me to sync an external juno node to the local-interchaintest network from genesis :partying_face:

Reecepbcups:
PR 921 implements this since it is required for ETH support

PR #921, opened 2023-12-21
feat: local-interchain ethereum support

Reecepbcups:
Built for Ethos testnet

- PR features from #917, #919
- Eth Port map
- [Local IC Eth type support](f3d138c8efd8acf1f68ba26998861f5d53113c63)
- [Fix config file overrides](614f3c4b87c44fc957915bbe03ed10da5e41fe36)
- Hide /info endpoint mnemonics
- Allows genesis accounts without mnemonic


chain/cosmos/chain_node.go
```diff
@@ -1291,7 +1292,23 @@ func (tn *ChainNode) CreateNodeContainer(ctx context.Context) error {
 		cmd = []string{chainCfg.Bin, "start", "--home", tn.HomeDir(), "--x-crisis-skip-assert-invariants"}
 	}
 
-	return tn.containerLifecycle.CreateContainer(ctx, tn.TestName, tn.NetworkID, tn.Image, sentryPorts, tn.Bind(), nil, tn.HostName(), cmd, nil)
+	usingPorts := make(nat.PortMap)
+
+	if tn.Index == 0 && chainCfg.HostPortOverride != nil {
+		for intP, extP := range chainCfg.HostPortOverride {
```
misko9, line 1298: if you override a single port, do you need to override all ports?

local-interchain/README.md
```diff
@@ -112,6 +120,21 @@ Here is a base chain template with every feature the configuration accepts. Acco
     "debugging": true,
     "block_time": "500ms",
     "encoding-options": ["juno"],
+    "host_port_override": {
+        "26657": "26657",
+        "1317": "1320",
```
misko9, line 125: nit: it's just a template, but is 1320 a typo?

misko9 (approved):

chain/cosmos/chain_node.go
Reecepbcups: no, if a port is not directly overridden a random mapping is assigned. ex:
image

chain/ethereum/ethererum_chain.go
```diff
@@ -198,15 +197,31 @@ func (c *EthereumChain) Start(testName string, ctx context.Context, additionalGe
 		dockerJsonFile := c.HomeDir() + path.Base(loadState)
```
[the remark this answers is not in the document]
Reecepbcups: d36933207f514f525e2476eede3d2471acf5bf91

local-interchain/README.md
Reecepbcups: This was to show that internal 1317 maps to 1320. Agree it is confusing so changed to add a documentation note instead and use 1317:1317 mapping

14b34db35c3c201088b433b667645d046881c6f4

chain/cosmos/chain_node.go
Reecepbcups: Ahh - while this logic was correct, it was duplicating bindings to all interfaces and ipv6. So I force the host mappings to ipv4 only now (instead of a blank PortMap it will use the desired ports, then override)

5ea8cd29c2ee6857509e6e7579442cc35b18b2ac

PR #928, opened 2024-01-01
Add support for ethereum's forge script

misko9:


chain/ethereum/forge.go
```diff
@@ -0,0 +1,116 @@
...
+	"path/filepath"
+
+	"github.com/docker/docker/api/types/mount"
+	"github.com/strangelove-ventures/interchaintest/v8/internal/dockerutil"
+)
+
+type ForgeScriptOpts struct {
+	ContractRootDir  string   // required, root directory of the contract with all local dependencies
+	SolidityContract string   // required, contract script to run
+	Signature        string   // optional, signature function to run, empty string uses default run()
```
Reecepbcups, line 16: nit: Would `SignatureFn` be a better name here? or would ETH native developers understand this in the context of forge and not signature schemes (i.e. BLS)

Also a link to the docs where the `run()` info & options are would be nice

Reecepbcups (approved):

chain/ethereum/forge.go
misko9: I like `SignatureFn`, forge's flag is `--sig`, but not sure the level of familiarity for an eth native dev. There are many options that can be pieced together in various ways so I'll also add a link to their docs.

PR #938, opened 2024-01-09
fix: Errors in cosmos Broadcaster are sometimes not propagated to caller

### LINKED ISSUE #937: Errors in cosmos Broadcaster are sometimes are not propagated to caller
When working on a test, I noticed the following behaviour when calling [BroadcastTx](https://github.com/strangelove-ventures/interchaintest/blob/66cb76406407fbc931486165d4c8f257ff86b9bb/chain/cosmos/broadcaster.go#L188).

The response returned [here](https://github.com/strangelove-ventures/interchaintest/blob/66cb76406407fbc931486165d4c8f257ff86b9bb/chain/cosmos/broadcaster.go#L222) did not error, but included an error message in the `RawLogs` of the response.

The error in question I believe happened in the `ValidateBasic` of a message that was broadcast.

I think in the case of `RawLogs` being populated on the linked line, we can return an error with the contents (even though the actual `err` returned is nil ).

The happy path here is that all fields except for `txHash` should be zero values.

cc @crodriguezvega

Reecepbcups:
closes #937


chain/cosmos/broadcaster.go
```diff
@@ -141,6 +141,16 @@ func (b *Broadcaster) UnmarshalTxResponseBytes(ctx context.Context, bytes []byte
...
 	}
+
+	// persist nested errors such as ValidateBasic checks.
+	code := resp.Code
+	rawLog := resp.RawLog
+
+	// rawLog can be empty or just an empty array "[]"
+	if code != 0 || len(rawLog) > 2 {
+		return resp, fmt.Errorf("error in transaction (code: %d): %s", code, rawLog)
+	}
```
jtieri, line 152: why is `len(rawLog) > 2` indicative of an error if `code == 0`?

Reecepbcups: I was thinking for ValidateBasic(), but ictest catches this in the Simulation (in BroadcastTx) vs at broadcast time. So should never be the base

Good call

Reecepbcups: updated: 671f3ed1bd15313dff7fd2da2f8c7b5fe20d2a23

jtieri (approved):
🚢 🚢

PR #948, opened 2024-01-16
update penumbra protos for ics20 testing

avahowell:


Reecepbcups:
Justin already updated the latest protos it looks like

PR #955, opened 2024-01-27
feat: Support CometMock

### LINKED ISSUE #824: Implement CometMock (opt-in)
https://github.com/informalsystems/CometMock

- [x] Depends on: https://github.com/cosmos/cosmos-sdk/pull/18110. Need to update ICtest to this commit so gas sim & hermes works
- [ ] design spec / impl. PR as an opt-in for a network. 
    - 1 docker instance per chain-id / network
- [ ] Move our internal test over to it

Reecepbcups:
closes #824

Adds opt-in support for [CometMock](https://github.com/informalsystems/CometMock). This results in significantly lower e2e test time

## Rough benchmark
*(( Intel Core i9-13900K @ 32x 5.5GHz, 128GB DDR4, x86_64 Linux 6.1.69-1-MANJARO, TestCometMock, avg of 3 runs ))*
- Standard: `105.50s`
- CometMock: `55.16s` (52% reduction)

## Summary
- Runs a CometMock process (26658) beside the Validator, cross connecting both
- Uses the built in 22331 port as the RPC override
- Sidecar processes now auto pull down images if they are not found

# TODO
- [ ] Add to other test to reduce times (excluding Upgrade)
- [ ] Documentation
- [ ] Test with multiple validators?
- [ ] Cache startup 'maybe at time we want to save state, we clone the docker volume and keep that as a snapshot, cloning for start volume for next runs, rather than storing in a new docker image.' May be its own issue


ibc/types.go
```diff
@@ -22,6 +26,8 @@ type ChainConfig struct {
 	ChainID string `yaml:"chain-id"`
 	// Docker images required for running chain nodes.
 	Images []DockerImage `yaml:"images"`
+	// https://github.com/informalsystems/CometMock usage.
+	CometMockImage []DockerImage `yaml:"comet-mock-image"`
```
Reecepbcups, line 30: Can maybe just include this in the Images?
 ( dd an optional tag to the DockerImage for 'type' so forks of CometMock can be used while we set the correct values when iterating the sidecars start)

Reecepbcups: ictest does not like this when the standard ChainName is being used. Would be bad UX to require the image Repos and correct ordering

Reecepbcups:
github CI may not like de80961a2e880f2fe2747a2fcd719d046f4fdde9

Reecepbcups:
github CI is being weird. req close and open to get latest commits

Reecepbcups:
LETS GOOOO FINALLY

boojamya (approved):
Nice job. This is an incredible addition!

chain/cosmos/chain_node.go
```diff
@@ -1008,6 +1068,8 @@ func (tn *ChainNode) CreateNodeContainer(ctx context.Context) error {
 }
 
 func (tn *ChainNode) StartContainer(ctx context.Context) error {
+	rpcOverrideAddr := ""
+
 	for _, s := range tn.Sidecars {
 		err := s.containerLifecycle.Running(ctx)
```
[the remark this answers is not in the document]
Reecepbcups: If s.preStart is true, err should always be nil. If it is an error, then preStart would have been false (not satisfying the && statement) and everything running as normal. We could add a check to ensure `!s.perStart` but would be kind of duplicated. also don't want a real return here, so it would only log it anyways

*Running() method acts more like a boolean vs error, this logic checks out. Is a little confusing for sure*

PR #965, opened 2024-02-02
Fixes and wires up `CreateClient`

boojamya:
Prior to this PR, `createClient` was not wired up.

### DIFF
--- a/examples/ibc/client_creation_test.go
+++ b/examples/ibc/client_creation_test.go
@@ -2,7 +2,6 @@ package ibc_test
 
 import (
 	"context"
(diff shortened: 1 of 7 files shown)


Reecepbcups:
yea getting / setting values from the relayer is a big need as we get into more complex cases (run into this twice at Ethos, found workaround but ideally this gets easier).

boojamya:
Yea we should look into this.

For now, I just discovered that in the relayer config, the `chain-id` is the same as the `chain-name`

Reecepbcups (approved):

PR #979, opened 2024-02-16
refactor!: changing block height type from uint64 to int64

spoo-bar:
CometBFT [(here)](https://github.com/cometbft/cometbft/blob/c920404f6b56fc9c366aee8774c4d011610c564d/proto/cometbft/state/v1/types.proto#L77) and Cosmos-sdk [(here)](https://github.com/cosmos/cosmos-sdk/blob/74dfab268484d3606e9df8cbc66327a6c627f29d/proto/cosmos/base/tendermint/v1beta1/types.proto#L27) both use `int64` to denote block height.

Currently the interchaintest framework uses `uint64` for all height related values.
This PR changes that to use `int64` instead. This would help in preventing frequent typecasting and ensure consistency with upstream types.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -368,17 +368,17 @@ func (tn *ChainNode) SetPeers(ctx context.Context, peers string) error {
 	)
 }
 
-func (tn *ChainNode) Height(ctx context.Context) (uint64, error) {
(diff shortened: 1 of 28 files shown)


Reecepbcups:
Will be a bit until we get a 8.2 tagged, so use main for now if you need reference for testing :)

PR #989, opened 2024-02-23
feat[v7]: support interchain security in local-interchain

Reecepbcups:
request by timewave

## Summary
Adds ICS support to local-interchain in a user friendly way with only 1 configuration change needed. Supports multiple consumers -> 1 provider

## Integration
Add `"ics_consumer_link": "providerchainid-1",` to your consumer chain. No changes on the provider side is required

## Run
`make install && local-ic start interchainsecurity`

### DIFF
--- a/local-interchain/interchain/start.go
+++ b/local-interchain/interchain/start.go
@@ -121,7 +121,7 @@ func StartChain(installDir, chainCfgFile string, ac *AppConfig) {
 	client, network := interchaintest.DockerSetup(fakeT)
 
 	// setup a relayer if we have IBC paths to use.
-	if len(ibcpaths) > 0 {
+	if len(ibcpaths) > 0 || len(icsPair) > 0 {
 		rlyCfg := config.Relayer
 
 		relayerType, relayerName := ibc.CosmosRly, "relay"
@@ -150,9 +150,6 @@ func StartChain(installDir, chainCfgFile string, ac *AppConfig) {
 
 		// iterate icsPair
 		for provider, consumers := range icsPair {
-			fmt.Println("provider", provider)
-			fmt.Println("consumers", consumers)
-
 			var p ibc.Chain
 			var c ibc.Chain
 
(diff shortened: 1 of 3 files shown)


Reecepbcups:
Currently  `make install && local-ic start interchainsecurity` starts but then fails on the gov prop `failed to start provider chain localcosmos-1: proposal status did not change to passed in expected number of blocks: proposal status (PROPOSAL_STATUS_VOTING_PERIOD) does not match expected: (PROPOSAL_STATUS_PASSED)"`

I assume the genesis params are wrong given SDK v45. That should resolve it

Reecepbcups:
StartConsumer seems to override any consumer chain genesis configs passed. Should modify the gensis with our params after the ccv setup consumer side (interchaintest bug)

local-interchain/interchain/start.go
```diff
@@ -137,6 +144,40 @@ func StartChain(installDir, chainCfgFile string, ac *AppConfig) {
...
+			logger.Info("Adding ICS pair", zap.String("provider", p.Config().ChainID), zap.String("consumer", c.Config().ChainID), zap.String("path", pathName))
+
+			ic = ic.AddProviderConsumerLink(interchaintest.ProviderConsumerLink{
+				Provider: p,
+				Consumer: c,
+				Relayer:  relayer,
+				Path:     pathName,
+			})
+		}
+	}
```
jtieri, line 179: this could be its own function, which would increase readability and make it easier to maintain.

boojamya (approved):

jtieri (approved):
looks good to me! i like the one config change made to signal a chain is a consumer chain, may steal this for the relayer as i think i made two changes for ICS related functionality.

local-interchain/interchain/start.go
Reecepbcups: atm many places in the `StartChain` could use this. It's kinda a monolith for starting up. Need to do a refactor here in the near future

PR #999, opened 2024-02-28
ux: use relative path for chains if applicable

Reecepbcups:
## Summary
Currently when using local-ic, the user has to input if they want to use the current directory with `ICTEST_HOME=.`. User input for something we should be able to assume is less than ideal.

## Solution
We check if the user is in a folder with the 'chains' directory +1 file. If this matches, we use the local directory. If not, we use the standard interchaintest install location for the files

## Showcase
image


boojamya (approved):
Makes sense.

PR #1009, opened 2024-03-07
feat: replace broadcaster's account retriever to use chain's bech32 prefix when converting AccAddress to string

### LINKED ISSUE #1008: Broadcaster fails on submitting transaction when the bech32 prefix of the chain is different with sdk.Config
This issue is similar to #898; the TX broadcaster contains the following code:
```
account, err := clientContext.AccountRetriever.GetAccount(clientContext, sdkAdd)
```
which is failed since the `AccountRetriever` is the auth module's implementation and uses the following function:
```
res, err := queryClient.Account(context.Background(), &QueryAccountRequest{Address: addr.String()}, grpc.Header(&header))
```
where `addr.String()` uses the static bech32 prefix and leads to the failure of this query when the chain's prefix is different.

madrezaz:
Closes #1008
This PR adds a custom implementation of the `AccountRetriever` (copied from auth module) which is modified to convert an `AccAddress` to string using the target chain's bech32 prefix.

### DIFF
--- a/chain/cosmos/account_retriever.go
+++ b/chain/cosmos/account_retriever.go
@@ -1,0 +1,88 @@
+package cosmos
+
+import (
+	"context"
+	"fmt"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"strconv"
+
+	grpc "google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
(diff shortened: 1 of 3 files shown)


Reecepbcups (approved):

Reecepbcups:
@madrezaz The latest main needs to be merged in then this is good to go :)

PR #1021, opened 2024-03-14
feat: `GetTimeoutHeight`

### LINKED ISSUE #240: Add a default timeout to use when sending packets for CosmosChain
Add a function `GetTimeout()` to `CosmosChain` which returns a valid timeout which is not expected to timeout before the packet is relayed. This will be useful when test writers write tests which require sending packets that must specify timeouts. It is probably easier to specify the timeout height rather than the timeout timestamp

Reecepbcups:
closes #240


chain/cosmos/cosmos_chain.go
```diff
@@ -1183,3 +1184,15 @@ func (c *CosmosChain) VoteOnProposalAllValidators(ctx context.Context, proposalI
 	}
 	return eg.Wait()
 }
+
+// GetTimeoutHeight returns a timeout height of 1000 blocks above the current block height.
+// This function should be used when the timeout is never expected to be reached
+func (c *CosmosChain) GetTimeoutHeight(ctx context.Context) clienttypes.Height {
+	height, err := c.Height(ctx)
+	if err != nil {
+		c.log.Error("failed to get chain height", zap.Error(err))
```
jtieri, line 1193: [logs should begin with a capital letter](https://github.com/strangelove-ventures/interchaintest/blob/main/docs/logging.md#usage)

chain/cosmos/cosmos_chain.go
```diff
@@ -1183,3 +1184,15 @@ func (c *CosmosChain) VoteOnProposalAllValidators(ctx context.Context, proposalI
...
+		return clienttypes.Height{}
```
jtieri, line 1194: will returning an empty struct guarantee a failure for tests that rely on this method? if so i think it's better to return an error here so that tests can fail earlier instead of waiting for whatever relies on this height to fail because it expects a valid height

Reecepbcups: resolved in 5a64a9d6c32f14a7046b51d4080a329eb2cb84e4

jtieri (approved):

PR #1022, opened 2024-03-14
feat: support remote JSON configs

Reecepbcups:
## Summary
Adds remote JSON configuration files to local-ic for start

## Why
To support spawn testnets on hetzner, gcp, etc private repos must also be supported. To accomplish this, the remote config can be public (gist), the docker build binary can (ghcr), but the codebase remain private

This way local-ic has everything it needs (docker image + config) to run, without exposing the private code


local-interchain/interchain/config.go
```diff
@@ -45,14 +33,43 @@ func LoadConfig(installDir, chainCfgFile string) (*types.Config, error) {
...
 	log.Println("Using chain config:", cfgFilePath)
 
+	return setConfigDefaults(&config), nil
+}
+
+func LoadConfigFromURL(url string) (*types.Config, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		log.Fatalln(err)
+	}
```
jtieri, line 56: It is the callers responsibility to ensure the response body of an http request is closed after use, failure to do so can lead to resource leak, etc.

```suggestion
	resp, err := http.Get(url)
	if err != nil {
		log.Fatalln(err)
	}
	defer resp.Body.Close()
```

local-interchain/cmd/local-ic/start_chain.go
```diff
@@ -40,6 +45,28 @@ var startCmd = &cobra.Command{
...
+		var err error
+
+		if strings.HasPrefix(configPath, "http") {
+			config, err = interchain.LoadConfigFromURL(configPath)
+			if err != nil {
+				panic(err)
+			}
+
+			// last part of the URL to be the test name
+			configPath = configPath[strings.LastIndex(configPath, "/")+1:]
```
jtieri, line 58: `configPath` can become any arbitrary string depending on what URL the user passes in, is this okay? is it solely used for creating the name of the running test case?

jtieri (commented):
just had the one suggestion otherwise this looks good

local-interchain/interchain/config.go
Reecepbcups: great catch

local-interchain/cmd/local-ic/start_chain.go
Reecepbcups: yes, it is only used as the test name (and technically nothing here is also okay as other data is appended)

else you get a super long testname for gist which takes up the an entire terminal line. really messy

jtieri (approved):

PR #1028, opened 2024-03-19
feat(v7)!: validators can have different genesis and self-delegation amounts

### LINKED ISSUE #1027: feat: Allow unequal distribution of stake among validators
## Problem

Currently, all validators start up with the same amount of stake. It's possible to modify what this stake is via `ModifyGenesisAmounts` in the chain configuration, but that still assumes a single set of genesis and self-delegation amounts, with no way to set it per-validator. This makes it hard to reproduce certain bugs that happen when there's an unequal distribution of stake; one such bug was found on gaia `v15.1.0-rc0`, for instance.

## Proposal

`ModifyGenesisAmounts` should take a validator index; the returned amounts will be used for that specific validator.

If this sounds like a good idea, I can put up a PR.

fastfadingviolets:
Closes #1027 

## Summary

Some issues only arise when there's an unequal distribution of stake among validators. This PR allows interchaintest to create chains whose validators have different stake. It does this by changing the `ModifyGenesisAmounts` function in the chain config so that it takes in an index; the returned values will be used as the genesis and self-delegation amounts for the validator at that specific index.

### DIFF
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -872,28 +872,26 @@ func (c *CosmosChain) Start(testName string, ctx context.Context, additionalGene
 
 	decimalPow := int64(math.Pow10(int(*chainCfg.CoinDecimals)))
 
-	genesisAmount := types.Coin{
-		Amount: sdkmath.NewInt(10_000_000).MulRaw(decimalPow),
-		Denom:  chainCfg.Denom,
-	}
+	genesisAmounts := make([][]types.Coin, len(c.Validators))
+	genesisSelfDelegation := make([]types.Coin, len(c.Validators))
 
-	genesisSelfDelegation := types.Coin{
-		Amount: sdkmath.NewInt(5_000_000).MulRaw(decimalPow),
-		Denom:  chainCfg.Denom,
+	for i := range c.Validators {
+		genesisAmounts[i] = []types.Coin{{Amount: sdkmath.NewInt(10_000_000).MulRaw(decimalPow), Denom: chainCfg.Denom}}
+		genesisSelfDelegation[i] = types.Coin{Amount: sdkmath.NewInt(5_000_000).MulRaw(decimalPow), Denom: chainCfg.Denom}
+		if chainCfg.ModifyGenesisAmounts != nil {
+			amount, selfDelegation := chainCfg.ModifyGenesisAmounts(i)
+			genesisAmounts[i] = []types.Coin{amount}
+			genesisSelfDelegation[i] = selfDelegation
+		}
 	}
-
-	if chainCfg.ModifyGenesisAmounts != nil {
-		genesisAmount, genesisSelfDelegation = chainCfg.ModifyGenesisAmounts()
-	}
-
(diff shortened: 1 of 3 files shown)


fastfadingviolets:
I'll port this to `main` if it gets merged :)

ibc/types.go
```diff
@@ -42,8 +42,8 @@ type ChainConfig struct {
 	PreGenesis func(ChainConfig) error
 	// When provided, genesis file contents will be altered before sharing for genesis.
 	ModifyGenesis func(ChainConfig, []byte) ([]byte, error)
-	// Modify genesis-amounts
-	ModifyGenesisAmounts func() (sdk.Coin, sdk.Coin)
+	// Modify genesis-amounts for the validator at the given index
+	ModifyGenesisAmounts func(int) (sdk.Coin, sdk.Coin)
```
Reecepbcups, line 46: Should modify/add an e2e test to use this. Where 2 validators are used and the 2nd (index 1) has a different amount.

This is extremely used for Proof of Authority testing as well

chain/cosmos/cosmos_chain.go
```diff
@@ -872,28 +872,26 @@ func (c *CosmosChain) Start(testName string, ctx context.Context, additionalGene
...
+	genesisSelfDelegation := make([]types.Coin, len(c.Validators))
+
+	for i := range c.Validators {
+		genesisAmounts[i] = []types.Coin{{Amount: sdkmath.NewInt(10_000_000).MulRaw(decimalPow), Denom: chainCfg.Denom}}
+		genesisSelfDelegation[i] = types.Coin{Amount: sdkmath.NewInt(5_000_000).MulRaw(decimalPow), Denom: chainCfg.Denom}
+		if chainCfg.ModifyGenesisAmounts != nil {
+			amount, selfDelegation := chainCfg.ModifyGenesisAmounts(i)
+			genesisAmounts[i] = []types.Coin{amount}
+			genesisSelfDelegation[i] = selfDelegation
+		}
```
Reecepbcups, line 885: There should probably be some sanity checks here/before that 
- `len(genesisAmounts) <= len(c.Validators)`
- `i < len(c.Validators)` on each loop iteration 

if I am not mistaken.

fastfadingviolets: hmm, i'm not sure :) `genesisAmounts` is `make([][]types.Coin, len(c.Validators))` so it's guaranteed to be the same length as `c.Validators`, and `i` is `range c.Validators` so it's also gonna be `< len(c.Validators)`

this is why i made `ModifyGenesisAmounts` take in an index instead of just asking for it to return the list of `genesisAmounts` directly btw

ibc/types.go
fastfadingviolets: done; i didn't end up pulling StakingQueryDelegation just because there's no easy way (that I know) to get a node's validator address anyway. Instead i just called `query staking validators` directly and parsed that.

Reecepbcups (approved):

Reecepbcups:
@mergifyio backport main

Reecepbcups:
Thanks for your work on this! super appreciated :)

PR #1037, opened 2024-03-25
chore(deps): versions

Reecepbcups:


Reecepbcups:
i think SDK v0.50.5 was tagged wrong or something. Getting this same issue with the ibc-rate-limit repo I am upgrading.

Reecepbcups:
v0.50.6 SDK in a few weeks will get the patch in to fix

.gitignore
```diff
@@ -4,4 +4,6 @@ interchaintest.test
 .idea
 /bin
 vendor
-examples/hyperspace/exported_state.json
\ No newline at end of file
+examples/hyperspace/exported_state.json
+
+go.work.sum
```
Reecepbcups, line 9: This redundant work.sum is always incorrect vs the standard go.sum's. I can't see why it is needed since it always has issues, so ignoring should be a good way to fix constant merge conflicts since adding

jackzampolin (approved):
ACK

PR #1042, opened 2024-04-02
fix: dependency issues in v6

faddat:
For some time, there have been a set of complex replaces in interchaintest.  this PR:

* resolves the complex replaces, eliminating all of them
* bumps go to a supported version
* is replicable on v7


Reecepbcups (approved):
Looks fine to me, we don't support v6 anymore so doubt many will update if they are still on such legacy software.

PR #1044, opened 2024-04-03
feat: node-index, auth key interaction, and YAML support

Reecepbcups:
ref: #795

## Summary
Adds node index and auth key interaction commands for the local-ic action handler. Required for private testnet networks (rather than using bash).

Also now supports YAML for configs (comments are nice sometimes)


local-interchain/cmd/local-ic/start_chain.go
```diff
@@ -94,6 +97,27 @@ local-ic start https://pastebin.com/raw/Ummk4DTM
 	},
 }
 
+// GetConfigWithExtension returns the config with the file extension attached if one was not provided.
+// If "hub" is passed it, it will search for hub.yaml, hub.yml, or hub.json.
+// If an extension is already applied, it will use that.
+func GetConfigWithExtension(parentDir, config string) (string, error) {
```
Reecepbcups, line 103: before I just appended .json to the end of the filename if it was not found, instead it's better to try and find the file with possible extensions, then return that instead.

Reecepbcups (commented):
Tested locally with running `hub`, `hub.yml`, `cosmoshub`, `cosmoshub.json` and a remote github - all started up and read as expected

local-interchain/chains/hub.yml
```diff
@@ -0,0 +1,33 @@
+---
```
Reecepbcups, line 1: This is just cosmoshub.json -> yaml. tested and works the same

PR #1069, opened 2024-04-15
feat!: using ica controller instead of intertx

spoo-bar:
- Changing the existing ICA funcs to use `interchain-accounts controller` instead of `intertx` module as it is deprecated. 

With the moving of the module features to module_modname.go, open to moving the ica features to `module_ica_controller.go`


Reecepbcups (approved):
Looks good! This can be moved in another PR to it's own file once some gRPC queries are added on top. Not a priority just a nice to have.

Thanks for your work!

PR #1077, opened 2024-04-19
chore!: Parse packet data from the packet_data_hex attribute

DimitrisJim:
Reference https://github.com/cosmos/ibc-go/pull/6023

It seems like there's no other spot events are currently parsed? Either way, can leave as draft and trigger a run of our e2es to see the green ticks before opening.


Reecepbcups (approved):

Reecepbcups:
🫡

PR #1087, opened 2024-04-22
fix(v7)!: check transactions succeed after they're included in a block

fastfadingviolets:
## Summary

Sometimes a transaction returns code 0 initially, but then fails when a validator goes to actually include it in a block. This PR adds code to cosmos' ExecTx to make sure transactions are verified after their inclusion in a block; if they fail at any stage of proceedings, an error is returned.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -552,6 +552,19 @@ func (tn *ChainNode) ExecTx(ctx context.Context, keyName string, command ...stri
 	if err := testutil.WaitForBlocks(ctx, 2, tn); err != nil {
 		return "", err
 	}
+	// The transaction can at first appear to succeed, but then fail when it's actually included in a block.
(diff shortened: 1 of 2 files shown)


Reecepbcups:
Yea this is an SDK thing. It always returns code 0 so long as it passes message Validate(), but then once included is when the real failure can occur (like gas, etc)

I do worry that some users may want a tx which is successful on submit but fail upstream (I have this exact case checked with proof of authority). I need to think on this moer

fastfadingviolets:
hmm, i think we should maybe introduce different error types then; that way you get an `err` back and you can make sure it's of type `UpstreamFailure` or whatever

Reecepbcups:
Thinking about it, I think it makes sense to for ExecTx. If someone wants to bypass they can use `Exec` / `ExecBin` to ignore

Reecepbcups (approved):

Reecepbcups:
Ty!

PR #1093, opened 2024-04-24
refactor(v7): get consumer genesis from provider chain and transform it

fastfadingviolets:
## Summary

Instead of building the ccv genesis for the consumer, we can fetch it from the provider chain. This is preferred, as it guarantees that it works across different consumer/provider versions of ICS

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -38,6 +38,7 @@ import (
 	dockerclient "github.com/docker/docker/client"
 	"github.com/docker/go-connections/nat"
 	"go.uber.org/zap"
+	"golang.org/x/mod/semver"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
@@ -719,6 +720,24 @@ func (tn *ChainNode) IsAboveSDK47(ctx context.Context) bool {
 	// so we use this to switch between new and legacy SDK logic.
 	// https://github.com/cosmos/cosmos-sdk/pull/14149
 	return tn.HasCommand(ctx, "genesis")
+}
+
+// ICSVersion returns the version of interchain-security the binary was built with.
+// If it doesn't depend on interchain-security, it returns an empty string.
+func (tn *ChainNode) ICSVersion(ctx context.Context) string {
+	if strings.HasPrefix(tn.Chain.Config().Bin, "interchain-security") {
+		// This isn't super pretty, but it's the best we can do for an interchain-security binary.
+		// It doesn't depend on itself, and the version command doesn't actually output a version.
+		// Ideally if you have a binary called something like "v3.3.0-my-fix" you can use it as a version, since the v3.3.0 part is in it.
+		return semver.Canonical(tn.Image.Version)
+	}
+	info := tn.GetBuildInformation(ctx)
+	for _, dep := range info.BuildDeps {
+		if strings.HasPrefix(dep.Parent, "github.com/cosmos/interchain-security") {
+			return semver.Canonical(dep.Version)
+		}
+	}
+	return ""
 }
 
 // AddGenesisAccount adds a genesis account for each key
--- a/chain/cosmos/cosmos_chain.go
+++ b/chain/cosmos/cosmos_chain.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"math"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"sync"
@@ -17,8 +18,6 @@ import (
 
 	govv1beta1 "cosmossdk.io/api/cosmos/gov/v1beta1"
 	sdkmath "cosmossdk.io/math"
-	abcitypes "github.com/cometbft/cometbft/abci/types"
-	"github.com/cometbft/cometbft/proto/tendermint/crypto"
 	"github.com/cosmos/cosmos-sdk/codec"
 	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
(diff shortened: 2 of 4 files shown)


examples/ibc/ics_test.go
```diff
@@ -31,64 +32,83 @@ func TestICS(t *testing.T) {
 		},
 	}
 
-	for _, tt := range tests {
-		tt := tt
-		testname := tt.name
+	icsVersions := []string{"v3.1.0", "v3.3.0", "v4.0.0"}
```
Reecepbcups, line 35: is there a difference between v3.1 and v3.3? if not we can just do 


```suggestion
        icsVersions := []string{"v3.3.0", "v4.0.0"}
```

chain/cosmos/chain_node.go
```diff
@@ -721,6 +722,24 @@ func (tn *ChainNode) IsAboveSDK47(ctx context.Context) bool {
...
 }
 
+// ICSVersion returns the version of interchain-security the binary was built with.
+// If it doesn't depend on interchain-security, it returns an empty string.
+func (tn *ChainNode) ICSVersion(ctx context.Context) string {
+	if strings.HasPrefix(tn.Chain.Config().Bin, "interchain-security") {
+		// This isn't super pretty, but it's the best we can do for an interchain-security binary.
+		// It doesn't depend on itself, and the version command doesn't actually output a version.
+		// Ideally if you have a binary called something like "v3.3.0-my-fix" you can use it as a version, since the v3.3.0 part is in it.
+		return semver.Canonical(tn.Image.Version)
```
Reecepbcups, line 732: I do not quite understand this, if interchain-security as a binary is used. What if we used Gaia and/or neutron. This logic would never flow?

What about if we check the BuildDeps first. If nothing, then return `semver.Canonical(tn.Image.Version)` instead of "" ?

like
```go
func (tn *ChainNode) ICSVersion(ctx context.Context) string {
	info := tn.GetBuildInformation(ctx)
	for _, dep := range info.BuildDeps {
		if strings.HasPrefix(dep.Parent, "github.com/cosmos/interchain-security") {
			return semver.Canonical(dep.Version)
		}
	}
	return return semver.Canonical(tn.Image.Version)
}
```

chain/cosmos/cosmos_chain.go
```diff
@@ -1263,83 +1300,21 @@ func (c *CosmosChain) StartConsumer(testName string, ctx context.Context, additi
...
-		ccvconsumertypes.DefaultGenesisState().GetParams(),
-	)
-
-	ccvState.Params.Enabled = true
-
-	ccvStateMarshaled, err := c.cfg.EncodingConfig.Codec.MarshalJSON(ccvState)
-	c.log.Info("HERE STATE!", zap.String("GEN", string(ccvStateMarshaled)))
+	consumerICS := c.GetNode().ICSVersion(ctx)
+	providerICS := c.Provider.GetNode().ICSVersion(ctx)
+	ccvStateMarshaled, err = c.transformCCVState(ctx, ccvStateMarshaled, consumerICS, providerICS)
```
Reecepbcups, line 1314: I do like how this abstracts out the setup. My worry though is that as ICS upgrades to newer versions, teams using this will have to get upstream patches in vs being able to fix in their own test. May not be an issue, we just need to keep refernce of the old way to do it as well

chain/cosmos/cosmos_chain.go
```diff
@@ -1197,6 +1193,47 @@ func (c *CosmosChain) StartProvider(testName string, ctx context.Context, additi
 	return nil
 }
 
+func (c *CosmosChain) transformCCVState(ctx context.Context, ccvState []byte, consumerVersion, providerVersion string) ([]byte, error) {
```
Reecepbcups, line 1196: This func confuses me a bit. Check my logic

- If provider & consumer are the same OR < v3.3.0, no modify
- if provider >3.3.0 and provider is newer than ther consumer, see if newer thjan v4.0.0 and use that. Else use v3.X
- else use the consumer version
- create the consumer genesis json from state. Then use the ics-cd to transfer the genesis to the provider version? This requires `interchain-security-cd` to have an official release rather than using something like gaia. Which means if we wanted to use the ICS v5.0.0-alpha.1 tag, we can't currently 

hmm

Reecepbcups:
This may help me with some of my [ICS v5](https://github.com/strangelove-ventures/interchaintest/pull/1088) integration. Will play around with it there as well to understand it better.

examples/ibc/ics_test.go
fastfadingviolets: The two are different, yeah, v3.3.0 is the first one to need an explicit transform when coming over from 3.1

chain/cosmos/cosmos_chain.go
fastfadingviolets: yeah, i think this'll depend on how ICS breaks as it upgrades to other versions; ideally the basic flow remains the same, and maybe we have to keep track of how the transform exactly happens

chain/cosmos/cosmos_chain.go
fastfadingviolets: Almost right :)

If provider >3.3.0 and provider is newer than ther consumer, see if newer thjan v4.0.0 and use that. Else use **v4.0.0 and tell it to transform --to v3.x.x**. This is sadly needed because if I've got a v3.3.0 provider and a v3.1.0 consumer, the v3.3.0 interchain-security-cd binary doesn't know how to transform the genesis to v3.1.0, but the v4.0.0 binary does...

> create the consumer genesis json from state. Then use the ics-cd to transfer the genesis to the provider version?

Almost, use it to transfer the genesis to the consumer version. How this is done depends on the provider version --if the provider is newer, then we need an `ics-cd` that's at least as new as the provider, since it needs to be able to read the new provider's genesis

> This requires interchain-security-cd to have an official release rather than using something like gaia. Which means if we wanted to use the ICS v5.0.0-alpha.1 tag, we can't currently

Correct. I find this extremely annoying. My strong preference would be that consumer chains compile themselves with the `genesis transform` command, which they should be able to get from the ICS version they're linked against. AFAIK no current consumer chain does that, so we do need an `ics-cd` binary. I suppose we could make which docker image it grabs it from configurable, and then if you need, say, an alpha tag, you can build it yourself and configure that image.

Reecepbcups (approved):

Reecepbcups:
This fixed an issue with an older version of neutron I was using originally on the ICS v5 WIP PR. Great work! really appreciated a ton here

Issue #1091, opened 2024-04-23
Feature Request: Allow exposing custom ports on all nodes of a chain


pirtleshell:
Originally mentioned in #1089 

**Problem**: Currently, exposing custom ports on a node's container is only supported through the `HostPortOverride` property of chain configs. Because it defines the host port, only one container can have the port exposed (multiple exposures conflict on the port binding). Right now, this is handled by only exposing the port on the first validator node. Ideally, we'd be able to expose ports on all nodes for the chain, not just one validator.

**Proposal**: Add a new chain config property to define what ports we want exposed (inside the container). Expose those ports to the for all chain containers on any available port (similar to how GRPC / etc are currently exposed).

For cosmos chains (probably the others as well), this can be implemented by injecting the user-defined port ids into the `usingPorts` map here: https://github.com/strangelove-ventures/interchaintest/blob/c8b741617cd8da95d4377c938c411c452490d16f/chain/cosmos/chain_node.go#L1057

**Example**: Ethermint chains run Ethereum JSON-RPC APIs on port 8545. This port could be exposed, and the endpoint used, by using a new config property & helper func to get the host address:
```go
chainConfig := ibc.ChainConfig{
	// chain config details here...
	ExposeAdditionalPorts: []string{"8545/tcp"}
}
// ...
// get address for the exposed ports on a node
fullNode0EvmUrl := chain.FullNodes[0].GetHostAddress(ctx, "8545/tcp")
validator0EvmUrl := chain.Validators[0].GetHostAddress(ctx, "8545/tcp")
```

The benefit of a new implementation like this is that we can expose ports on all validators and full nodes, instead of only the first validator.

I have tested an implementation like this for cosmos chains. Would a PR for that be accepted? I do not have capacity to fully test it for the other chain types. 🥲

Reecepbcups:
@pirtleshell Yes, a PR would be accepted even if only supporting Cosmos chains at first. Feel free to push a draft and allow maintainer edits so I can modify as needed to get into the next v8 release ideally

PR #1098, opened 2024-04-24
feat: add ExposeAdditionalPorts config option

pirtleshell:
for cosmos chains, `ExposeAdditionalPorts` takes a list of port ids that will be exposed on all validators and full nodes for the chains on random host ports.

extends the ethermint test by exposing the EVM & verifying the port actually gets exposed.

Resolves cosmos chain portion of #1091 

Happy to make any changes or refactors. I _think_ I updated all the config things that need updated (`Clone` & `MergeChainSpecConfig`), but possible I missed something.


pirtleshell:
hey @Reecepbcups, i'm pretty sure the test failure is unrelated to my changes and it may pass if given another try.

PR #1099, opened 2024-04-25
fix(v7): prevent concurrent hermes operations on the same chain

fastfadingviolets:
## Summary

Allowing concurrent ops can result in `hermes create ...` commands failing because there's sequence numbers being modified by simultaneous commands. This PR adds per-chain locks that must be acquired to e.g. create a connection or channel


Reecepbcups:
After multiple re-runs I can not get the unit test or test-ibc-examples w/ hermes to pass here

Reecepbcups (approved):

PR #1102, opened 2024-04-28
refactor(v7): make dockerutil `internal/` public

### LINKED ISSUE #869: (v7) make functions and packages public
Since this is a framework for testing, we should attempt to make all functions (within reason) public.

# Versions
- [ ] v7
- [x] v8 (main)

- ex: `PrivValFileContent` and `OverwritePrivValFile` in v7.

this may also be useful with the `internal/` repo. For example, the local-interchain lense clone could use docker internal functions, but it can not be accessed outside.

Reecepbcups:
closes #869 (part 2)


jtieri:
Before i actually review the changes i just wanted to ask, does it make sense for all of this code to become public? 

I can see the argument for making the `internal/dockerutil` package public because there are actually a lot of scenarios where you may want lower level access to the containers or volumes for container lifecycle management or reading/writing data, but i don't necessarily see a case where the `blockdb` or `mocktesting` packages are useful for developers working outside of the `interchaintest` repo. 

iirc `mocktesting` is used for internal testing inside `interchaintest`, so i can't see why anyone would need to import code from this package. I also can't think of a reason devs would need to do anything with `blockdb` if `interchaintest` is already handling all of the database specific logic for you e.g. connecting to DB, the schemas, extracting chain related data, inserting data, etc. I can actually imagine making the `blockdb` package public could lead to weird behavior if devs try to call certain functions and also have `BlockDatabaseFile` set in `Interchain.Build`, for example you could end up trying to call `ConnectDB` explicitly after the internal logic in `interchaintest` has already initialized a connection and this would lead to resource contention due to sqlite not supporting more than 1 open connection per process very elegantly. 

tl;dr: I'm in favor of making the Docker related code public but the rest of it seems useless from the perspective of developers making use of the testing framework and i think it would also create additional footguns in the case of `blockdb` since this code wasn't written in a way that it should be invoked outside of internal `interchaintest` functionality.

Reecepbcups:
These are good points. I was thinking developers could extend BlockDB, but this again could lead to more confusion when new contributors try using it directly improperly. Ill change this now

version.go
```diff
@@ -1,4 +1,4 @@
-package version
+package interchaintest
```
jtieri, line 1: Should we leave this as an internal package also? I don't see why someone would need access to this outside of `interchaintest`

test_setup.go
```diff
@@ -74,7 +73,7 @@ func StartChainPair(
 		TestName:          t.Name(),
 		Client:            cli,
 		NetworkID:         networkID,
-		GitSha:            version.GitSha,
+		GitSha:            GitSha,
```
jtieri, line 76: see comment about leaving `version` internal

cmd/interchaintest/interchaintest_test.go
```diff
@@ -65,7 +64,7 @@ func TestMain(m *testing.M) {
 		}
 		os.Exit(0)
 	case "version":
-		fmt.Fprintln(os.Stderr, version.GitSha)
+		fmt.Fprintln(os.Stderr, interchaintest.GitSha)
```
jtieri, line 67: see comment about leaving `version` as an internal package

jtieri (approved):
had the one comment but otherwise this looks good to me! i'll just approve now so you aren't blocked on me

version.go
Reecepbcups: reverted: https://github.com/strangelove-ventures/interchaintest/pull/1102/commits/7be6bd7100724cc8f11f901f1d7789236e5c6885

PR #1111, opened 2024-05-02
[BP v7 <- #1098] feat: add ExposeAdditionalPorts config option

pirtleshell:
for cosmos chains, ExposeAdditionalPorts takes a list of port ids that will be exposed on all validators and full nodes for the chains on random host ports.

extends the ethermint test by exposing the EVM & verifying the port actually gets exposed.

this is a backport of #1098 into the v7 branch.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -1404,6 +1404,9 @@ func (tn *ChainNode) CreateNodeContainer(ctx context.Context) error {
 	for k, v := range sentryPorts {
 		usingPorts[k] = v
 	}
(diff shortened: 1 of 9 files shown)


Reecepbcups (approved):

pirtleshell:
hello @Reecepbcups, i'm not sure what would be causing the local interchain CI to fail. could you kick it off again to see if the problem resolves itself?

Reecepbcups:
I think something with the package registry is being weird, re-running. I'll look more into it if it fails again

Reecepbcups:
8 month old upstream Cargo package decided to break lol

PR #1118, opened 2024-05-08
feat!(sidecar): add start check callback with rollkit example

gjermundgaraba:
I don't know what the best way to deal with dependencies between sidecars and chains, but since I used Celestia as a sidecar for the Rollkit example it needed some time to wait for the sidecar to start up before starting the chain itself.

Putting this here for discussion.


examples/cosmos/rollkit_test.go
```diff
@@ -0,0 +1,221 @@
...
+			Name:      "gm",
+			ChainName: "gm",
+			Version:   "tutorial-local-da",
+			ChainConfig: ibc.ChainConfig{
+				Type:    "cosmos",
+				Name:    "gm",
+				ChainID: "gm",
+				Images: []ibc.DockerImage{
+					{
+						Repository: "ghcr.io/gjermundgaraba/gm",
```
Reecepbcups, line 57: Is this just the standard `https://github.com/rollkit/gm` or are there changes made to work with this?

ibc/types.go
```diff
@@ -256,6 +256,7 @@ type SidecarConfig struct {
 	StartCmd         []string
 	Env              []string
 	PreStart         bool
+	StartCheck       func(int) error
```
Reecepbcups, line 259: This is a good addition

examples/cosmos/rollkit_test.go
```diff
@@ -0,0 +1,221 @@
...
+				Denom:            "stake",
+				CoinType:         "118",
+				SigningAlgorithm: "",
+				GasPrices:        "0stake",
+				GasAdjustment:    2.0,
+				TrustingPeriod:   "112h",
+				NoHostMount:      false,
+				SkipGenTx:        false,
+				PreGenesis:       nil,
+				ModifyGenesis: func(config ibc.ChainConfig, bytes []byte) ([]byte, error) {
```
Reecepbcups, line 73: Ahh I see now. Ideally we would abstract this away and have a RollKit section which does this but does not have the user be informed of all the setup complexities. Everything here looks reasonable imo

great work!

chain/cosmos/sidecar.go
```diff
@@ -48,6 +49,7 @@ func NewSidecar(
 	log *zap.Logger,
 	validatorProcess bool,
 	preStart bool,
+	startCheck func(int) error,
```
Reecepbcups, line 52: not to self, API breaking for sidecars (minor)

Reecepbcups:
Ref https://github.com/rollchains/tiablob/blob/main/interchaintest/setup_celestia.go as well vs using a combined instance. I should probably get this moved here and help clean  up some setup

examples/cosmos/rollkit_test.go
```diff
@@ -0,0 +1,221 @@
...
+							"26650/tcp",
+							"26657/tcp",
+							"26658/tcp",
+							"26659/tcp",
+							"9090/tcp",
+						},
+						StartCmd: []string{"/bin/bash", "/opt/entrypoint.sh"},
+						Env:      nil, // Here we could set CELESTIA_NAMESPACE if needed
+						PreStart: true,
+						StartCheck: func(index int) error {
```
Reecepbcups, line 143: Maybe to be more inline with https://github.com/strangelove-ventures/interchaintest/pull/1123, we rename this to PostStart() on the sidecar process. Just so it more closely matches that PRs API conventions and also is more versatile vs just a standard check.

Does not seem the `index` here is used? Could put the CosmosChain here or something

gjermundgaraba: That naming makes sense.

Yeah, the idea was originally to send in the SidecarProcess but that gets to a cyclical dependency so ended up with an index and then in the end didn't use it. 

In the code below I just do Sidecars[0], but I guess the idea was to access the right one. But I suppose I would always know the order they are added (being the same as the SidecarConfigs slice order). So could send in the chain like you said instead.

examples/cosmos/rollkit_test.go
gjermundgaraba: Yeah, it would be great to automagically sort this somehow without all the manual stuff for sure :)

examples/cosmos/rollkit_test.go
gjermundgaraba: Just standard, but there is no updated docker builds so I just took the tutorial-local-da branch and built a docker image of that

Reecepbcups:
Hey, we have still not decided on a priority level for this PR. It is something we want to get in, but have to juggle with other priorities right now as well.

Review / impl time is still unknown

PR #1124, opened 2024-05-10
chore(env)!: `IBCTEST_` to `ICTEST_`

### LINKED ISSUE #1049: chore: rename ENV from IBCTEST_ -> ICTEST_

Reecepbcups:
closes #1049

Will make note at the top of the release logs for this as priority (v8.4.0)

Also bumps deps


.github/dependabot.yml
```diff
@@ -8,10 +8,10 @@ updates:
   - package-ecosystem: gomod
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
```
jtieri, line 11: what's the motivation for moving this from `weekly` -> `monthly`?

jtieri (approved):
changes look good, just need to be sure we make note of this in the next release like you mentioned.

.github/dependabot.yml
Reecepbcups: we just depend on upstream for types. Not as high a priority vs standard chains depending. Easier to maintainer / less PR spam

Issue #998, opened 2024-02-27
Wasm IBC test ought not fail.


peterblockman:
I ran TestWasmIbc on the main branch and got this error: "Error:  Should NOT be empty, but was"
```
                Error Trace:   /Projects/test/interchaintest/examples/ibc/wasm/wasm_ibc_test.go:173
                Error:          Should NOT be empty, but was 
                Test:           TestWasmIbc
```
looks like it failed to reflect the account on juno2

full log: 
[ibc_test_log.txt](https://github.com/strangelove-ventures/interchaintest/files/14417648/ibc_test_log.txt)

Reecepbcups:
Thank you for the report! I will be taking a look into why this is occurring and moving it to our main testing CI suite to ensure it does not happen again

jonathanpberger:
Calling in the CosmWasm calvary @kakucodes

PR #1131, opened 2024-05-13
chore(test): migrate wasm ibc to cw_ibc_example from reflect_send 

kakucodes:
Spoke to Reece about the difficulties in getting the wasm ibc test to pass and we agreed it would be sensible to use the `cw-ibc-example` contract to test wasm ibc as it's more straight forward and more up to date.

Additionally I added a helper for querying the ContractInfo from a chain so that we can get the wasm ports directly instead of assuming the implementation detail of how portIds are constructed.

https://github.com/0xekez/cw-ibc-example

resolves: #998


chain/cosmos/module_cosmwasm.go
```diff
@@ -225,3 +225,19 @@ func (tn *ChainNode) DumpContractState(ctx context.Context, contractAddress stri
...
+func (tn *ChainNode) QueryContractInfo(ctx context.Context, contractAddress string) (*ContractInfoResponse, error) {
+	stdout, _, err := tn.ExecQuery(ctx,
+		"wasm", "contract", contractAddress,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	res := new(ContractInfoResponse)
+	if err := json.Unmarshal([]byte(stdout), res); err != nil {
```
Reecepbcups, line 239: nit: stdout is already a []byte

examples/ibc/wasm/wasm_ibc_test.go
```diff
@@ -110,48 +110,45 @@ func TestWasmIbc(t *testing.T) {
 	juno1Chain := juno1.(*cosmos.CosmosChain)
 	juno2Chain := juno2.(*cosmos.CosmosChain)
 
-	// Store ibc_reflect_send.wasm contract
-	ibcReflectSendCodeId, err := juno1Chain.StoreContract(
-		ctx, juno1User.KeyName(), "sample_contracts/ibc_reflect_send.wasm")
+	// Store ibc_reflect_send.wasm contract on juno1
+	juno1ContractCodeId, err := juno1Chain.StoreContract(
+		ctx, juno1User.KeyName(), "sample_contracts/cw_ibc_example.wasm")
```
Reecepbcups, line 115: can we add a README.md (in sample_contracts) that links to where `sample_contracts/cw_ibc_example.wasm` upstream is? and the commit used for this compile

Reecepbcups (approved):
2 minor nits, otherwise LGTM ! Approving so it is not blocked on me

PR #1141, opened 2024-05-24
feat(local-ic): genesis config builder

Reecepbcups:
## Reason
Building the configs is not currently possible & something spawn should support (for rapid generation & development). Especially with added complexity via multiple ICS partners.  Find and replace no longer is enough, especially with multiple bech32 prefixes in 1 file. ref:  https://github.com/rollchains/spawn/issues/112 / https://github.com/rollchains/spawn/pull/142

## Summary
- This add a builder pattern for new chains with customization and some abstraction.
- Removes unused struct values
- Fix local-interchain go.mod namespace for external import (<- #1142)
- Already integrated upstream


local-interchain/interchain/types/genesis.go
```diff
@@ -11,11 +21,73 @@ type GenesisAccount struct {
...
+
+	}
+
+	return accounts
+}
+
+func MnemonicToAddress(mnemonic, prefix string, coinType uint32) string {
+	e := moduletestutil.MakeTestEncodingConfig()
+
+	kr := keyring.NewInMemory(e.Codec)
```
Reecepbcups, line 74: This keyring is GC'ed after return. So deletion of keys does not have to occur as there is no overlap

local-interchain/interchain/types/types.go
```diff
@@ -25,18 +34,156 @@ type RestServer struct {
...
+	DockerImage  ibc.DockerImage `json:"docker_image" yaml:"docker_image"`
+	StartupFlags []string        `json:"startup_flags" yaml:"startup_flags"`
 }
 
 type IBCChannel struct {
 	ChainID string             `json:"chain_id" yaml:"chain_id"`
 	Channel *ibc.ChannelOutput `json:"channel" yaml:"channel"`
 }
+
+// ConfigFileOverrides overrides app toml configuration files.
```
Reecepbcups, line 47: Everything below was just copied from the other files

chain/cosmos/ics.go
```diff
@@ -365,7 +365,21 @@ func (c *CosmosChain) transformCCVState(ctx context.Context, ccvState []byte, co
...
+		imageVersion = icsCfg.ProviderVerOverride
+	}
+
+	if icsCfg.ConsumerVerOverride != "" {
+		toVersion = icsCfg.ConsumerVerOverride
+	}
+
+	if imageVersion == toVersion && imageVersion != "" {
+		return ccvState, nil
+	}
```
Reecepbcups, line 381: allowing override is useful in some cases for special ICS forks (i.e. Ethos) or partial set security

chain/cosmos/genesis.go
```diff
@@ -42,7 +42,7 @@ func ModifyGenesis(genesisKV []GenesisKV) func(ibc.ChainConfig, []byte) ([]byte,
 			}
 
 			if err := dyno.Set(g, values.Value, path...); err != nil {
-				return nil, fmt.Errorf("failed to set value (index:%d) in genesis json: %w", idx, err)
+				return nil, fmt.Errorf("failed to set key '%s' as '%+v' (index:%d) in genesis json: %w", values.Key, values.Value, idx, err)
```
Reecepbcups, line 45: I found debugging these to be kinda a pain with only index, key add is a UX improvement.

cmd/interchaintest/example_matrix_custom.json
```diff
@@ -14,7 +14,7 @@
           {
             "Repository": "ghcr.io/strangelove-ventures/heighliner/gaia",
             "Version": "latest",
-            "UidGid": "1025:1025"
+            "uid-gid": "1025:1025"
```
Reecepbcups, line 17: This does not break anything upstream, only internal test

local-interchain/interchain/types/chains_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestChainsGeneration(t *testing.T) {
+	t.Run("single chain runs", func(t *testing.T) {
+		require.NoError(t, ChainEthereum().SaveJSON("chains/gen-ethereum.json"))
+		require.NoError(t, ChainStargaze().SaveJSON("chains/gen-stargaze.json"))
+		require.NoError(t, ChainCosmosHub("localcosmos-1").SaveJSON("chains/gen-cosmoshub.json"))
+		require.NoError(t, ChainOsmosis().SaveJSON("chains/gen-osmosis.json"))
+		require.NoError(t, ChainJuno("localjuno-1").SaveYAML("chains/gen-juno.yml"))
```
jtieri, line 15: why is this the only one that uses yaml?

local-interchain/interchain/types/types.go
```diff
@@ -1,9 +1,20 @@
...
+	"path/filepath"
+
+	"github.com/go-playground/validator"
+	"github.com/strangelove-ventures/interchaintest/v8/chain/cosmos"
 	"github.com/strangelove-ventures/interchaintest/v8/ibc"
+	"github.com/strangelove-ventures/interchaintest/v8/testutil"
+	"gopkg.in/yaml.v3"
 )
 
+// === Local-IC Application ===
```
jtieri, line 16: what's up with this comment? is it here for a reason?

local-interchain/interchain/types/chain_builder.go
```diff
@@ -0,0 +1,203 @@
...
+		},
+		Accounts: append(
+			[]GenesisAccount{
+				NewGenesisAccount("acc0", c.Bech32Prefix, "25000000000%DENOM%", c.CoinType,
+					"decorate bright ozone fork gallery riot bus exhaust worth way bone indoor calm squirrel merry zero scheme cotton until shop any excess stage laundry",
+				),
+				NewGenesisAccount("acc1", c.Bech32Prefix, "24000000000%DENOM%", c.CoinType,
+					"wealth flavor believe regret funny network recall kiss grape useless pepper cram hint member few certain unveil rather brick bargain curious require crowd raise",
+				),
+			},
```
jtieri, line 134: why do these two accounts get created here? is there a reason the genesis amounts are different?

jtieri (approved):
i had a few questions but the code changes all look good. i don't think there is any reason to block this so i'll approve it for you.

local-interchain/chains/base_ibc.json
```diff
@@ -35,7 +35,6 @@
                 "version": "v17.0.0"
             },
             "block_time": "2s",
-            "encoding-options": ["juno"],
```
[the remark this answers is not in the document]
Reecepbcups: Yes it was removed in the past and I never went back through and removed from all the configs / docs

local-interchain/cmd/local-ic/new_chain.go
```diff
@@ -43,53 +34,41 @@ var newChainCmd = &cobra.Command{
 			}
 		}
 
-		var config Chains
+		var config types.ChainsConfig
 		var chains []ictypes.Chain
 
-		for i := 1; i < 1000; i++ {
+		for i := 1; i < 20; i++ {
```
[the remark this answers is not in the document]
Reecepbcups: Arbitrary number, 1000 was far too many. 10 - 20 seems more realistic for a really complex setup

local-interchain/interchain/types/chain_builder.go
Reecepbcups: These are specific deterministically generated accounts found in node's test_node.sh as well. So if a user wants pre-funded accounts to be compatible with their keyring adds, this adds those in

Different amounts just to make it easier to debug & not accidentally use one when you mean to use the other (have had this happen many times, and doing this helps a ton)

local-interchain/interchain/types/chains_test.go
Reecepbcups: Just to ensure it works as expected. If this failed the rest of them would fail due to how the logic is setup

local-interchain/interchain/types/types.go
Reecepbcups: no, I was planning something with the format of this file and never did. Removed

PR #1149, opened 2024-06-02
chore(deps): june bumps + fix release CI

Reecepbcups:


.github/workflows/release.yml
```diff
@@ -20,9 +20,8 @@ jobs:
         with:
             go-version: ${{ env.GO_VERSION }}
 
-      # This must be go install vs make install as to not statically link the binary
-      # to the worker node.
-      - run: cd local-interchain && go mod tidy && go install ./...
+      # NOTE: make install must not be statically linked to the MakeFileInstallDirectory
+      - run: cd local-interchain && go mod tidy && IGNORE_STATIC_LINK=true make install
```
Reecepbcups, line 24: ik this is not a version bump but is required for the next release (so versions are actually shows in local-ic version)

proof of work:
image

joelsmith-2019 (approved):

PR #1175, opened 2024-07-15
feat: allow consumers to have different keys than providers

fastfadingviolets:
This reflects how ICS chains are launched in the wild, with validators setting their consensus keys before the chain launches

### DIFF
--- a/chain/cosmos/ics.go
+++ b/chain/cosmos/ics.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	sdkmath "cosmossdk.io/math"
@@ -20,6 +21,7 @@ import (
 	"github.com/strangelove-ventures/interchaintest/v7/ibc"
 	"github.com/strangelove-ventures/interchaintest/v7/testreporter"
 	"github.com/strangelove-ventures/interchaintest/v7/testutil"
+	"github.com/tidwall/gjson"
 	"go.uber.org/zap"
 	"golang.org/x/mod/semver"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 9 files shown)


chain/cosmos/ics.go
```diff
@@ -55,23 +57,27 @@ func (c *CosmosChain) FinishICSProviderSetup(ctx context.Context, r ibc.Relayer,
 
 	providerVal := stakingVals[0]
 
-	beforeDel, err := c.StakingQueryDelegationsTo(ctx, providerVal.OperatorAddress)
```
fastfadingviolets, line 58: i don't really get what happened here, but as soon as I added a second validator to the ICS test, StakingQueryDelegationsTo started returning delegation responses that only had the second delegation and lacked the first self-delegation. So the balance after was less than the balance before; it seems if you just query the validators and get the balance from there it works out, so that's what I did.

chain/cosmos/ics.go
```diff
@@ -230,13 +239,37 @@ func (c *CosmosChain) StartConsumer(testName string, ctx context.Context, additi
...
-		if err := c.Validators[i].OverwritePrivValFile(ctx, privVal); err != nil {
-			return err
-		}
+		i := i
+		val := val
+		eg.Go(func() error {
+			copy := true
+			if c.cfg.InterchainSecurityConfig.ConsumerCopyProviderKey != nil {
+				copy = c.cfg.InterchainSecurityConfig.ConsumerCopyProviderKey(i)
+			}
```
Reecepbcups, line 248: minor nit
```suggestion
			copy = c.cfg.InterchainSecurityConfig.ConsumerCopyProviderKey(i) != nil
```

chain/cosmos/ics.go
```diff
@@ -148,7 +157,7 @@ func (c *CosmosChain) StartProvider(testName string, ctx context.Context, additi
 			InitialHeight: clienttypes.Height{RevisionNumber: clienttypes.ParseChainID(consumer.cfg.ChainID), RevisionHeight: 1},
 			GenesisHash:   []byte("gen_hash"),
 			BinaryHash:    []byte("bin_hash"),
-			SpawnTime:     time.Now(), // Client on provider tracking consumer will be created as soon as proposal passes
+			SpawnTime:     time.Now().Add(2 * time.Minute), // Needed so that there's time for consumer keys to be assigned
```
Reecepbcups, line 160: Does it take 2 minutes to get keys copied?

Maybe instead something like X seconds * number of chains you are starting or something? this way it scales with complexity instead of a hardcoded value. Seems long for a simple ICS <-> Consumer each with 1 val

or maybe an ENV_VAR to override this to make it longer as needed? (would need to add to the docs/env*.md section)

Reecepbcups:
Once above are resolved I'll approve and get this in thank you so much for this! :saluting_face:

chain/cosmos/ics.go
fastfadingviolets: yeah, I like the env var solution

Reecepbcups (approved):

Reecepbcups:
Thanks Violet! :saluting_face:

PR #1180, opened 2024-07-18
feat: allow consumers to have different keys than providers (#1175)

fastfadingviolets:
I ended up needing this backport, so I did it myself 😄 It's the same as #1179 but i fixed the proposals query. I figure there won't be hundreds of proposals in a test case (??) so hopefully we don't have to worry about pagination


Reecepbcups (approved):
:fire:

PR #1183, opened 2024-07-22
fix: upgrade proposals for SDK >= 0.47

fastfadingviolets:
There's no more `submit-proposal software-upgrade` command, so it has to go through a full proposal.json

### DIFF
--- a/chain/cosmos/module_gov.go
+++ b/chain/cosmos/module_gov.go
@@ -57,13 +57,16 @@ func (tn *ChainNode) UpgradeProposal(ctx context.Context, keyName string, prop S
 	if tn.IsAboveSDK47(ctx) {
 		cosmosChain := tn.Chain.(*CosmosChain)
 
-		authority, err := cosmosChain.GetGovernanceAddress(ctx)
-		if err != nil {
-			return "", err
+		if prop.Authority == "" {
+			authority, err := cosmosChain.GetGovernanceAddress(ctx)
+			if err != nil {
+				return "", err
+			}
+			prop.Authority = authority
 		}
 
 		msg := upgradetypes.MsgSoftwareUpgrade{
-			Authority: authority,
+			Authority: prop.Authority,
 			Plan: upgradetypes.Plan{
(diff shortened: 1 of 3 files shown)


chain/cosmos/module_gov.go
```diff
@@ -52,6 +54,29 @@ func (tn *ChainNode) GovSubmitProposal(ctx context.Context, keyName string, prop
 
 // UpgradeProposal submits a software-upgrade governance proposal to the chain.
 func (tn *ChainNode) UpgradeProposal(ctx context.Context, keyName string, prop SoftwareUpgradeProposal) (string, error) {
+	if tn.IsAboveSDK47(ctx) {
+		cosmosChain := tn.Chain.(*CosmosChain)
+
+		authority, err := cosmosChain.GetGovernanceAddress(ctx)
```
Reecepbcups, line 60: hmm, we can't really assume gov based authority with more PoA networks coming on.

`SoftwareUpgradeProposal` could add an authority string, but that may break json proto marshaling without having to write a custom decoder to ignore that key. or break the function signature to add a new authority string.

or leave it as is and fix it when it comes and issue or just override it on their side for ChainNode. Thoughts?

Reecepbcups (approved):

chain/cosmos/module_gov.go
fastfadingviolets: i think i'm okay to put it in the `SoftwareUpgradeProposal` --does it get marshalled right now? I don't find anywhere that does so

Reecepbcups: if not amazing that would work

Reecepbcups:
```go
		if prop.Authority == "" {
			authority, err := cosmosChain.GetGovernanceAddress(ctx)
			if err != nil {
				return "", err
			}
			prop.Authority = authority
		}
```

I tried to push but no perms. This works

Reecepbcups:
ty!!

PR #1184, opened 2024-07-22
feat: Allow configuring default --gas limit for transactions.

fastfadingviolets:
It's useful to just set "--gas auto" for all transactions. Right now, --gas auto will be set as the default for gaia chians.

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -504,7 +504,7 @@ func (tn *ChainNode) FindTxs(ctx context.Context, height int64) ([]blockdb.Tx, e
 // with the chain node binary.
 func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
 	command = append([]string{"tx"}, command...)
-	var gasPriceFound, gasAdjustmentFound, feesFound = false, false, false
+	var gasPriceFound, gasAdjustmentFound, gasFound, feesFound = false, false, false, false
 	for i := 0; i < len(command); i++ {
 		if command[i] == "--gas-prices" {
 			gasPriceFound = true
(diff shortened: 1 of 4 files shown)


ibc/types.go
```diff
@@ -44,6 +44,8 @@ type ChainConfig struct {
 	GasPrices string `yaml:"gas-prices"`
 	// Adjustment multiplier for gas fees.
 	GasAdjustment float64 `yaml:"gas-adjustment"`
+	// Default gas limit for transactions. May be empty, "auto", or a number.
+	Gas string `yaml:"gas"`
```
Reecepbcups, line 48: If gas is empty, should we default to `auto` rather than having to specify it?

```go
Gas string `yaml:"gas" default:"auto"`
```
or something like that

chain/cosmos/chain_node.go
```diff
@@ -515,13 +515,19 @@ func (tn *ChainNode) TxCommand(keyName string, command ...string) []string {
...
 	}
 	if !gasPriceFound && !feesFound {
 		command = append(command, "--gas-prices", tn.Chain.Config().GasPrices)
 	}
 	if !gasAdjustmentFound {
 		command = append(command, "--gas-adjustment", strconv.FormatFloat(tn.Chain.Config().GasAdjustment, 'f', -1, 64))
 	}
+	if !gasFound && !feesFound && tn.Chain.Config().Gas != "" {
+		command = append(command, "--gas", tn.Chain.Config().Gas)
+	}
```
Reecepbcups, line 530: How does this affect tx commands which are not actual txs? (i.e. appd tx decode <amino_encoded>

May want to add a test for this in [chain_miscellaneous_test.go](https://github.com/strangelove-ventures/interchaintest/blob/9b9c2db326d6ec628db7b035be23eb1e7edfc441/examples/cosmos/chain_miscellaneous_test.go#L27)

fastfadingviolets: done!

ibc/types.go
fastfadingviolets: so it turns out that the yaml parser doesn't actually interpret the `default` flag 🙃 
I added it anyway fwiw, and I added a test with it; I guess we could go set `gas: auto` to a bunch of the bigger chains, like i did with gaia here

Reecepbcups (approved):

PR #1187, opened 2024-07-23
fix: keep Provider/ConsumerVerOverride from being wiped by ICS version

fastfadingviolets:
IMO if we're manually overriding versions for the purposes of CCV transforms, we should honour whatever the user gave us as the version and not modify it.


Reecepbcups (approved):
Oops! I thought I did this in the past, must have not pushed or been in a dream :joy: Good catch

PR #1204, opened 2024-08-06
feat: local chain registry endpoint

Reecepbcups:
ref: https://github.com/rollchains/spawn/pull/194

## Summary

if the chain_registry.json file is found when running `local-ic start <name>`, it will showcase this in an easy to query way. This makes it easy to connect wallets and build existing infra on top. Better UX

image

### DIFF
--- a/local-interchain/interchain/router/router.go
+++ b/local-interchain/interchain/router/router.go
@@ -52,6 +52,14 @@ func NewRouter(
 		log.Printf("chain_registry.json not found in %s, not exposing endpoint.", wd)
 	}
 
+	chainRegistryAssetsFile := filepath.Join(wd, "chain_registry_assets.json")
+	if _, err := os.Stat(chainRegistryAssetsFile); err == nil {
+		crH := handlers.NewChainRegistry(chainRegistryAssetsFile)
+		r.HandleFunc("/chain_registry_assets", crH.GetChainRegistry).Methods(http.MethodGet)
+	} else {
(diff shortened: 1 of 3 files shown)


local-interchain/interchain/router/router.go
```diff
@@ -36,6 +39,27 @@ func NewRouter(
...
+	}
+
+	chainRegistryAssetsFile := filepath.Join(wd, "chain_registry_assets.json")
+	if _, err := os.Stat(chainRegistryAssetsFile); err == nil {
+		crH := handlers.NewChainRegistry(chainRegistryAssetsFile)
+		r.HandleFunc("/chain_registry_assets", crH.GetChainRegistry).Methods(http.MethodGet)
+	} else {
+		log.Printf("chain_registry_assets.json not found in %s, not exposing endpoint.", wd)
+	}
+
```
jtieri, line 62: we call the same handler here so is it just handling the case where the file could be named either `chain_registry.json` or `chain_registry_assets`? is there any reason for this explicit handling of both?

jtieri (approved):
i think this looks good. had one question but the answer seems mostly obvious so approving

local-interchain/interchain/router/router.go
Reecepbcups: it uses the same as it just returns the JSON object. Since its specific to chain registry and only do that 1 function, it makes sense to just re-use it vs duplicate code

PR #1205, opened 2024-08-06
feat: support gordian consensus

Reecepbcups:
## Summary

Adds consensus support for Gordian chains

## Status
- [x] Chain starts
- [x] Queries current height
- [ ] (upstream) Tx hash integration / callbacks
- [ ] gRPC ify / import types
- [ ] ...

## TODO
- [ ] Make a new consensus interface type (so we do not have to do the if tn.IsConsensus() ) for the different interactions needd
- [ ] Cleanup debugging printers
- [ ] Use proper tn.Hostname()'s instead of the 0.0.0.0 host port hack
- [ ] Upstream simapp w/ proper docker image (ghcr). Either in gordian or spawn chain export or something


Reecepbcups:
closing in favor of 1265 as it also migrates cometbft

PR #1206, opened 2024-08-07
feat(local-ic): catch os.Interrupt / sigterm signals and cleanup

Reecepbcups:
# Summary

Since local-ic was created, ctrl+c has been on the back burner for a year. It now properly cleans up a running instance.

image

### Breaking Consideration
`ict-` namespaces had to be added to containers so we don't kill non ict based networks. This MAY be breaking for some users of ICT if they hardcode expect values as hostnames. To fix, they would have to add `ict-` to the start of their container name.

This verification actually broke one of my rust test [here](https://github.com/strangelove-ventures/interchaintest/pull/1206/commits/a340c5256193e21689e1ca5cedb4c1b2457abdbe), but this is likely only due to internal verifications. I don't see why someone would do this outside of this repo.

## Manual Testing
- make install && local-ic start juno_ibc
- Wait some time (sometimes right away, sometimes delayed, someone after full start)
- run other containers like `docker run -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4`
- ctrl +c
- docker ps

### DIFF
--- a/local-interchain/interchain/handlers/actions.go
+++ b/local-interchain/interchain/handlers/actions.go
@@ -70,7 +70,7 @@ func (a *actions) PostActions(w http.ResponseWriter, r *http.Request) {
 
 	action := ah.Action
 	if action == "kill-all" {
-		KillAll(a.ctx, a.ic, a.vals, a.relayer, a.eRep)
+		KillAllICTContainers(a.ctx)
 		return
 	}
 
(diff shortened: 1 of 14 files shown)


local-interchain/interchain/handlers/actions.go
```diff
@@ -216,21 +218,49 @@ func (a *actions) relayerCheck(w http.ResponseWriter, r *http.Request) error {
 	return err
 }
 
-func KillAll(ctx context.Context, ic *interchaintest.Interchain, vals map[string][]*cosmos.ChainNode, relayer ibc.Relayer, eRep ibc.RelayerExecReporter) {
-	if relayer != nil {
-		if err := relayer.StopRelayer(ctx, eRep); err != nil {
+// KillAllICTContainers kills all containers that are prefixed with interchaintest specific container names.
+// This is required as you can not ctrl+c kill a local-ic instance before the ic.Build(), else containers will remain running.
+// (Since the vals map does not have context to these containers yet - due to still being built)
+func KillAllICTContainers(ctx context.Context) {
```
Reecepbcups, line 224: This could be a useful interchaintest cleanup method. May move to ict vs being in local-ic. Thoughts?

local-interchain/interchain/handlers/actions.go
```diff
@@ -216,21 +218,49 @@ func (a *actions) relayerCheck(w http.ResponseWriter, r *http.Request) error {
...
+
+		if len(container.Names) == 0 {
+			continue
+		}
+
+		name := strings.ToLower(container.Names[0])
+		name = strings.TrimPrefix(name, "/")
+
+		// leave non ict relayed containers running
+		if !(strings.HasPrefix(name, "ict-") || strings.HasPrefix(name, "interchaintest-") || strings.HasPrefix(name, "ictrelay-")) {
```
joelsmith-2019, line 245: It might make sense to pull out "ict-" "interchaintest-" and "ictrelayer-" as constants somewhere. That should help limit typos or mistakes if we were ever to rename the containers. We could also use those constants when we create the containers.

local-interchain/interchain/handlers/actions.go
joelsmith-2019: Does the current method not successfully close all of the ICT containers? What benefits would this have outside of local-ic

joelsmith-2019 (approved):

local-interchain/interchain/handlers/actions.go
Reecepbcups: It does not during ic.Build(...), if it fails within the containers persist. This would allow for someone to run into that, and then clean up.

Though keeping the containers can be useful for debugging, so we would not force this on ICT users, just allow them to use it if they see the need (like I did).

local-interchain/interchain/handlers/actions.go
Reecepbcups: 1897a72fec60a4f4e8e8f06694747a6287343289

local-interchain/interchain/handlers/actions.go
Reecepbcups: I moved to dockerutil since that is a better place for it anyways. Just added returning a []string of the instances removed rather than printing out (so its up to the app dev)

d10ae057221274a222e2b634258b5c5aea30db20

Reecepbcups:
re-tested with enw changes (despite no logical differnces). confirmed working still as expected

It's a used enough feature I do not think a test case for it makes sense to invest that time. So unless we have problems, manual is the best way

image

PR #1207, opened 2024-08-07
feat(local-ic): bash driver

Reecepbcups:
ref https://github.com/strangelove-ventures/interchaintest/issues/795
## Summary
Basic external bash driver which can be imported to interact with local-ic. Why? it's easier sometimes and I find myself wanting a helper constantly.

---
- [x] Add to main readme (depends on #1203 first)
- [x] External Sourcing
- [x] PollForStart
- [x] Arbitrary execution (binary, queries, or the container shell itself)
- [x] Relayers (Stop, Start, Exec, Get Channels)
- [x] CosmWasm (Store file, dump state)
- [x] Faucet
- [x] Add Full Node
- [x] Recover & View key/address
- [x] Store any file

image

### DIFF
--- a/local-interchain/bash/source.bash
+++ b/local-interchain/bash/source.bash
@@ -1,0 +1,62 @@
+# IMPORT ME WITH: source <(curl -s https://github.com/strangelove-ventures/interchaintest/tree/main/local-interchain/bash/source.bash)
+
+# CONTENT_HEADERS="Content-Type: application/json"
+
+# MAKE_GENERAL_REQUEST http://127.0.0.1:8080 '{"chain_id":"localjuno-1","action":"q","cmd":"bank total"}'
+# Call this standalone, not from another function
+# MAKE_GENERAL_REQUEST() {
+#     API=$1
+#     DATA=$2
+
+#     curl "$API" --include --header "\"Content-Type: application/json\"" -X POST --data $DATA
+# }
+
+# MAKE_REQUEST http://127.0.0.1:8080 localjuno-1 q "bank total"
+MAKE_REQUEST() {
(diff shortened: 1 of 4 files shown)


AntiTyping (approved):

AntiTyping:
Looks good. One thing that I would add in the future would be some help message if a bash function is called with the wrong number of arguments - but that depends on feedback from users. The scripts are easy to understand and anyone who has knowledge of cosmos should not have an issue with reading them.

local-interchain/bash/source.bash
```diff
@@ -0,0 +1,184 @@
...
+    AMOUNT=$3
+    ADDRESS=$4
+    ICT_MAKE_REQUEST $API $CHAIN_ID "faucet" "amount=$AMOUNT;address=$ADDRESS"
+}
+
+# ICT_ADD_FULL_NODE http://127.0.0.1:8080 "localjuno-1" "1"
+ICT_ADD_FULL_NODE() {
+    API=$1
+    CHAIN_ID=$2
+    AMOUNT=$3
```
AntiTyping, line 158: I would consider using `local`

`local API=$1, CHAIN_ID=$2, AMOUNT=$3`

in this way variables will not leak into outside context which might make debugging harder especially if multiple functions are called in succession with different args

Reecepbcups: good call

Reecepbcups: resolved in: 92e712f80dc0db60ce8cf058c7c0e9c5d331f58a

joelsmith-2019 (approved):

Issue #1153, opened 2024-06-24
Support addition of relayers to existing chain pair outside of interchain.Build


chatton:
in the ibc-go E2Es, we are working on a refactor with a goal of reducing the number of runners we use in our CI, by re-using a pair of chains across an entire test suite.

We are effictively doing

1. in `SetupSuite` we create the `interchain` calling `ic.Build()` (with `SkipPathCreation` set to true)
2. In `SetupTest` we establish a client/connection/channel etc for the test to act on.
3. Each test has a relayer to act on only the created channel for that test to avoid relayers relaying packets for other tests.

We are having an issue with part 3.

Due to the nature of how we construct interchain instances, we need to provide the relayers/links before chain creation. I.e. know details of test execution before the test is running. We would like to be able to lazily create a new relayer instance and wire it up (fund wallet accounts, add chains to config files etc ) in the `SetupTest` phase of each test. 

Looking at some of the code, it looks like the `AddChainConfiguration` and `RestoreKey` functions may be what we need, and are already public, but it seems that there are some [unexported types](https://github.com/strangelove-ventures/interchaintest/blob/fa08aa0d58fbe62498a0acabb25d9e55b534a756/interchain.go#L434) being leveraged here.

I think we would also need access to [this](https://github.com/strangelove-ventures/interchaintest/blob/fa08aa0d58fbe62498a0acabb25d9e55b534a756/interchain.go#L374-L381) functionality, however this is at genesis and of course could not be executed after the fact, we would need to be able to create and fund relayer accounts separately. Something similar to `GetAndFundTestUsers` but for relayers basically.

A temporary fix that I am experimenting is just pre-loading a bunch of relayers and then fetching one of these for a test. It's not great since it adds a bunch of time to start up, and you will likely create more than you need. 

Please let me know if this is already a possibility, or there is some workaround to lazily create relayers and wire them up to an `interchain` that is already built!

PR #1208, opened 2024-08-07
refactor: faster

Reecepbcups:
## Summary

Make Interchaintest Fast like zooooooooooooooomies

## Benchmarks
```
// TODO: time to beat:
 // Before this PR: 103.31s
 // With faster blocks (in this PR) & faster relayer polling: 70.39s / 67.03s 
// Without startup Locks: 62.8s
```

### DIFF
--- a/chain/cosmos/chain_node.go
+++ b/chain/cosmos/chain_node.go
@@ -109,7 +109,7 @@ type ChainNodes []*ChainNode
 
 const (
 	valKey      = "validator"
-	blockTime   = 2 // seconds
+	blockTime   = 500 * time.Millisecond
 	p2pPort     = "26656/tcp"
 	rpcPort     = "26657/tcp"
 	grpcPort    = "9090/tcp"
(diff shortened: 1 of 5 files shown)


Reecepbcups:
Build (including relayer setup) takes 50s of the total 60s test time. If we can speed up the ic.Build we could get test down to an absolute best case of ~15s (down from 103s)

Reecepbcups:
https://github.com/strangelove-ventures/interchaintest/issues/1153 is probably the next big performance improvement. Getting relayers to auto provision for planned testing is ideal, lot of dead time waiting.

We already know the chain configs to `chains add`, `rly keys restore`, and `rly paths new` (config). Then we just `rly tx link` as soon as the chain is at height capable.

Must allow relayers pre ic.Build() standalone

Reecepbcups:
ideas from Andrew:
- sleep inifinite then just exec into the container on start

Reecepbcups:
export genesis state after IBC start, then can load back from that in future runs

Reecepbcups:
I have been taking things from this PR and putting into others as it seems right.

PR #1213, opened 2024-08-09
refactor: docker label cleanup for running containers

### LINKED ISSUE #1212: docker: use labels instead of ict- prefix for containers cleanup
ref: where it is impl: https://github.com/strangelove-ventures/interchaintest/pull/1206

Reecepbcups:
closes #1212

## Summary
The proper way to perform https://github.com/strangelove-ventures/interchaintest/pull/1206 (using hidden labels instead of name prefixes). Reverted back to previous formats, keeping the next release non api breaking)

Confirmed to work with relayers
image

### DIFF
--- a/chain/cosmos/sidecar.go
+++ b/chain/cosmos/sidecar.go
@@ -93,10 +93,10 @@ func NewSidecar(
 // on a per validator level.
 func (s *SidecarProcess) Name() string {
 	if s.validatorProcess {
-		return fmt.Sprintf("%s-%s-%s-val-%d-%s", dockerutil.ICTDockerPrefix, s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
+		return fmt.Sprintf("-%s-%s-val-%d-%s", s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
 	}
 
-	return fmt.Sprintf("%s-%s-%s-%d-%s", dockerutil.ICTDockerPrefix, s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
+	return fmt.Sprintf("%s-%s-%d-%s", s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
 }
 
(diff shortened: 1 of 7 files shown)


chain/cosmos/sidecar.go
```diff
@@ -93,10 +93,10 @@ func NewSidecar(
 // on a per validator level.
 func (s *SidecarProcess) Name() string {
 	if s.validatorProcess {
-		return fmt.Sprintf("%s-%s-%s-val-%d-%s", dockerutil.ICTDockerPrefix, s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
+		return fmt.Sprintf("-%s-%s-val-%d-%s", s.Chain.Config().ChainID, s.ProcessName, s.Index, dockerutil.SanitizeContainerName(s.TestName))
```
joelsmith-2019, line 96: Was this leading '-' intentional?

dockerutil/kill.go
```diff
@@ -2,24 +2,36 @@ package dockerutil
...
 )
 
 // KillAllInterchaintestContainers kills all containers that are prefixed with interchaintest specific container names.
 // This can be called during cleanup, which is especially useful when running tests which fail to cleanup after themselves.
 // Specifically, on failed ic.Build(...) calls.
 func KillAllInterchaintestContainers(ctx context.Context) []string {
+	if os.Getenv("KEEP_CONTAINERS") != "" {
+		return nil
+	}
+
```
joelsmith-2019, line 20: Would it make sense to make this != "false"? Might also be worth adding this to the docs

joelsmith-2019 (approved):

chain/cosmos/sidecar.go
Reecepbcups: I specifically checked this 3 times and still missed it :cry: great catch ty

dockerutil/kill.go
Reecepbcups: ICT has this other check where this is the standard
image

evmOptions.md explains it
image

chain/cosmos/sidecar.go
Reecepbcups: 9b096e47900a9d4684f2346c6be74d9935318368

dockerutil/kill.go
joelsmith-2019: Oh perfect! Sounds good with me

PR #1216, opened 2024-08-14
fix: tx out of gas

joelsmith-2019:
Fixes: gas rounding issue from https://github.com/strangelove-ventures/interchaintest/pull/747#discussion_r1320159263
Adds `gas` to ChainConfig.
Updates TxCommand to auto set `gas`.

### DIFF
--- a/examples/cosmos/chain_miscellaneous_test.go
+++ b/examples/cosmos/chain_miscellaneous_test.go
@@ -51,6 +51,8 @@ func CosmosChainTestMiscellaneous(t *testing.T, name, version string) {
 				CoinType:       "118",
 				ModifyGenesis:  cosmos.ModifyGenesis(sdk47Genesis),
 				EncodingConfig: wasmEncoding(),
+				GasAdjustment:  1.5,
+				Gas:            "auto",
 			},
 			NumValidators: &numVals,
 			NumFullNodes:  &numFullNodes,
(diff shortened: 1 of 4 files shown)


Reecepbcups (approved):

examples/cosmos/chain_miscellaneous_test.go
```diff
@@ -440,6 +443,40 @@ func testTXFailsOnBlockInclusion(ctx context.Context, t *testing.T, chain *cosmo
...
+	// check if generate contains "--gas"
+	require.Contains(t, generate, "--gas")
+
+	txJson, _, err := chain.GetNode().Exec(ctx, generate, nil)
+	require.NoError(t, err)
+
+	// encode
+	err = chain.GetNode().WriteFile(ctx, txJson, "tx.json")
+	require.NoError(t, err)
+	encode := chain.GetNode().TxCommand(users[0].KeyName(), "encode", chain.GetNode().HomeDir()+"/tx.json")
```
Reecepbcups: should use path.Join vs +, resolved with push

Reecepbcups (approved):

examples/cosmos/chain_miscellaneous_test.go
joelsmith-2019: Good catch!

PR #1220, opened 2024-08-15
feat(utxo): Add support for utxo chains

misko9:


chain/cosmos/cosmos_chain.go
```diff
@@ -345,6 +345,11 @@ func (c *CosmosChain) SendFunds(ctx context.Context, keyName string, amount ibc.
 	return c.getFullNode().BankSend(ctx, keyName, amount)
 }
 
+// Implements Chain interface
+func (c *CosmosChain) SendFundsWithNote(ctx context.Context, keyName string, amount ibc.WalletAmount, note string) (string, error) {
+	return c.getFullNode().BankSendWithNote(ctx, keyName, amount, note)
+}
```
Reecepbcups, line 351: Should we just modify BankSend with an `extraFlags ...string`?

I am good with adding this, but wonder if we really need 2 bank sends with just a note memo variant. UX wise, I can just mention in the release notes for next ICT v8.

chain/ethereum/ethererum_chain.go
```diff
@@ -345,6 +346,38 @@ func (c *EthereumChain) SendFunds(ctx context.Context, keyName string, amount ib
...
+	cmd := []string{"cast", "send", amount.Address, hexutil.Encode([]byte(note)), "--value", amount.Amount.String(), "--json"}
+	if keyName == "faucet" {
+		cmd = append(cmd,
+			"--private-key", "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
+			"--rpc-url", c.GetRPCAddress(),
+		)
+
+	} else {
+		cmd = append(cmd,
+			"--keystore", c.keystoreMap[keyName],
```
Reecepbcups, line 363: UX nit: keyName in c.keystoreMap check?

chain/utxo/utxo_chain.go
```diff
@@ -0,0 +1,520 @@
...
+
+func (c *UtxoChain) SendFunds(ctx context.Context, keyName string, amount ibc.WalletAmount) error {
+	_, err := c.SendFundsWithNote(ctx, keyName, amount, "")
+	return err
+}
+
+func (c *UtxoChain) SendFundsWithNote(ctx context.Context, keyName string, amount ibc.WalletAmount, note string) (string, error) {
+	partialCoin := amount.Amount.ModRaw(int64(math.Pow10(int(*c.Config().CoinDecimals))))
+	fullCoins := amount.Amount.Sub(partialCoin).QuoRaw(int64(math.Pow10(int(*c.Config().CoinDecimals))))
+	sendAmountFloat := float64(fullCoins.Int64()) + float64(partialCoin.Int64())/math.Pow10(int(*c.Config().CoinDecimals))
```
Reecepbcups, line 412: We love some go math

Reecepbcups (approved):

chain/cosmos/cosmos_chain.go
misko9: I talked with @agouin about this and I, at least, had a preference to not break the SendFunds API for everyone else. But if we should bite the bullet and just break it now, I'm okay with that too. There just could be many calls to it in test cases out there.

Reecepbcups: If we add `extraFlags ...string` it's not function signature breaking :) since its just optional

I think we add this & can reduce. It should only break our internal implementations which need updating

User facing functions wont change or require migration work

Reecepbcups: I can take this on in another PR if you prefer

chain/ethereum/ethererum_chain.go
misko9: Here's the current error that would be returned.
exit code 1:  Error: Keystore file `""` does not exist

But I can add that check in the various areas to make it a little clearer we don't know that keyname since it only resolves to an empty string for the keystore file.

chain/cosmos/cosmos_chain.go
misko9: I like that, but also forgot to include that we need to return the tx hash too... If the note/memo results in external processes, that tx hash is used to check the state. Again, something SendFunds() should probably do, but that will break the API.

Reecepbcups: gotcha, Ill make a ticket then and can decide in a future release. Or we wait to do that in the v52 refactor

PR #1225, opened 2024-08-21
fix: error returns for queries in cosmos staking & slashing

fastfadingviolets:
the queries tend to reply with a nil result when they fail, so trying to access that result will cause a panic


Reecepbcups (approved):

Reecepbcups:
Good catch, will be tagging a new release of ICT soon

PR #1227, opened 2024-08-21
feat(v7): override sdk v47 version check with env variable

joelsmith-2019:
Manually overrides the logic to determine if a chain is running SDK 47 or greater.


Reecepbcups (approved):

chain/cosmos/chain_node.go
```diff
@@ -722,6 +722,29 @@ func (tn *ChainNode) RecoverKey(ctx context.Context, keyName, mnemonic string) e
...
+	}
+
+	// Check OS environment variables for the SDK version.
+	env := os.Getenv("ICT_ABOVE_SDK_47")
+	if env != "" {
+		if env == "true" {
+			return true
+		} else {
+			return false
+		}
```
Reecepbcups, line 745: FYI could technically do `return env == "true"`

Reecepbcups (approved):

PR #1233, opened 2024-08-23
ci: add new CI jobs and bring existing ones into compliance

jtieri:
This PR revises the existing CI jobs to be conformant with our OSS guidelines. 

All repo settings should also be in compliance with the OSS template now, EXCEPT for the required status checks. It seems like the GitHub workflow files need to me merged before it allows you to update the settings to use them.

The changes also include linter fixes that were able to be applied via `golangci-lint run ./... --fix`

Edit: we decided to stick with the existing workflow for releases due to issues we were having with the cross-arch compilation of `local-ic`. We have decided that since `interchaintest` is a library and that we are only building the binary for `local-ic` it does not make sense to try and force a pattern that doesn't translate well.


.github/workflows/lint.yml
```diff
@@ -1,41 +1,34 @@
...
-      with:
-        profile: minimal
-        toolchain: stable
-        components: rustfmt, clippy
-    - name: Install clippy
-      run: rustup component add clippy
-    - name: Update
-      run: cargo update
-    - name: Run clippy
-      run: make lint
```
Reecepbcups, line 40: accidental removal?

.goreleaser.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - go mod tidy
+
+builds:
+  - id: "local-ic"
+    main: "./local-interchain/cmd/local-ic"
+    binary: "local-ic"
+    env:
+      - CGO_ENABLED=0
+    ldflags:
+      - -X main.Version=$(VERSION)
```
Reecepbcups, line 23: for spawn I use 

```
- -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser
```
so I think this needs to be

```suggestion
      - -X main.Version={{.Version}}
```

.goreleaser.yaml
```diff
@@ -0,0 +1,75 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
+
+# The lines below are called `modelines`. See `:help modeline`
+# Feel free to remove those if you don't want/need to use them.
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
+# vim: set ts=2 sw=2 tw=0 fo=cnqoj
+
+version: 2
```
Reecepbcups, line 9: Get the following error. Got it on the old binary & latest install from goreleaser. 
```
goreleaser build --snapshot --clean -f .goreleaser.yaml
  • starting build...
  • loading config file                              file=.goreleaser.yaml
  ⨯ build failed after 0s                    error=yaml: unmarshal errors:
  line 9: field version not found in type config.Project
```

Does not happen with spawn, so not a binary issue. Something with config

.github/workflows/lint.yml
jtieri: Yepp, will get this added back in

.goreleaser.yaml
jtieri: Good catch, this is exactly what the docs say as well. Not sure how i missed this given i was just working on this

.goreleaser.yaml
jtieri: Interesting 🤔 

I updated `goreleaser` this morning so should be using the most recent version available in `brew`

```
brew info goreleaser
==> goreleaser: stable 2.3.0 (bottled), HEAD
Deliver Go binaries as fast and easily as possible
https://goreleaser.com/
Installed
/opt/homebrew/Cellar/goreleaser/2.3.0 (9 files, 61.2MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-09-13 at 12:02:57
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/g/goreleaser.rb
License: MIT
==> Dependencies
Build: go ✘
==> Options
--HEAD
	Install HEAD version
==> Caveats
Bash completion has been installed to:
  /opt/homebrew/etc/bash_completion.d
==> Analytics
install: 1,322 (30 days), 3,273 (90 days), 14,028 (365 days)
install-on-request: 1,313 (30 days), 3,239 (90 days), 13,778 (365 days)
build-error: 0 (30 days)
```

Running the same command on macOS I see the following output.

```
goreleaser build --snapshot --clean -f .goreleaser.yaml
  • skipping validate...
  • cleaning distribution directory
  • loading environment variables
  • getting and validating git state
    • git state                                      commit=7f2f6cd83d557c1b6540f181049983ccf965d924 branch=justin/ci current_tag=v8.7.0 previous_tag=v8.6.1 dirty=false
    • pipe skipped                                   reason=disabled during snapshot mode
  • parsing tag
  • setting defaults
  • snapshotting
    • building snapshot...                           version=8.7.0-SNAPSHOT-7f2f6cd8
  • running before hooks
    • running                                        hook=cd local-interchain
    • running                                        hook=go mod tidy
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • building                                       binary=dist/local-ic_linux_amd64_v1/local-ic
    • building                                       binary=dist/local-ic_linux_arm_6/local-ic
    • building                                       binary=dist/local-ic_windows_arm_6/local-ic.exe
    • building                                       binary=dist/local-ic_windows_amd64_v1/local-ic.exe
    • building                                       binary=dist/local-ic_darwin_amd64_v1/local-ic
    • building                                       binary=dist/local-ic_linux_arm64/local-ic
    • building                                       binary=dist/local-ic_darwin_arm64/local-ic
    • building                                       binary=dist/local-ic_windows_arm64/local-ic.exe
```

jtieri: It was hanging on the build step for me but after enabling `cgo` it appears to work, however, it looks like builds fail for linux and windows at that point due to various issues 

Linux:
```
 ⨯ build failed after 16s                   error=failed to build for linux_amd64_v1: exit status 1: # runtime/cgo
linux_syscall.c:67:13: error: call to undeclared function 'setresgid'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
linux_syscall.c:67:13: note: did you mean 'setregid'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:593:6: note: 'setregid' declared here
linux_syscall.c:73:13: error: call to undeclared function 'setresuid'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
linux_syscall.c:73:13: note: did you mean 'setreuid'?
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/unistd.h:595:6: note: 'setreuid' declared here
```

Windows:
```
 ⨯ build failed after 13s                   error=failed to build for windows_amd64_v1: exit status 1: # runtime/cgo
gcc_libinit_windows.c:6:10: fatal error: 'windows.h' file not found
```

Not sure if there is a good way to fix that unfortunately and it appears that we rely on C libs due to the fact `blockdb` imports `modernc.org/sqlite` which imports `modernc.org/libc`

Reecepbcups: as far as I can tell with source graph, no one is using the blockdb. We don't even use it internally due to those CI issues we were getting.

If we don't have any internal reliance on it, what if we were to just, yeet it from the codebase? Would anyone care or notice? This is a bigger discussion that needs its own issue. May help us get off the SDK import as well? or limit scope of where it touches.

Reecepbcups: starting a draft here to test it & see how feasible it is https://github.com/strangelove-ventures/interchaintest/pull/1257

It looks like the FindTxs() methods can just be refactored to query consensus directly vs needing some complex indexing thing we don't use. This may get rid of codec requirements as well? Unsure yet. But sounds like the SDK reliance was a big part of this feature (iirc from a convo with jack 1 year ago)

Reecepbcups:
todo: remove goreleaser

Reecepbcups (approved):

PR #1235, opened 2024-08-24
fix(local-ic): multiple ics, faster relayer

Reecepbcups:
## Summary

Timewave has been using local-interchain extensively and has come across some issues. This PR resolves those.

## ✅ Issue 1 - Multiple ICS chains
When building multiple ICS chains, local-ic improperly only set the last consumer found, not all.

This is resolved as I forgot to iterate each consumer, as the parent for loop would not do that (just the providers). Leading to only the last consumer being setup.

## ✅ Issue 2 - Fast Chains
They want to run chains fast, and the relayer can not keep up. Added `MinLoopDuration` from upstream go rly -> here with env override for faster or slower (default 50ms). No userspace config changes required


local-interchain/chains/interchainsecurity.json
```diff
@@ -83,18 +83,67 @@
...
+            "gas_adjustment": 1.3,
+            "number_vals": 1,
+            "number_node": 0,
+            "ics_consumer_link": "localcosmos-1",
+            "debugging": true,
+            "block_time": "1s",
+            "genesis": {
+                "modify": [
+                    {
+                        "key": "consensus.params.block.max_gas",
```
Reecepbcups, line 133: SDK v50 moved this

agouin (approved):

PR #1237, opened 2024-08-26
feat: implement override flag when creating clients

boojamya:
In order to create two different clients on chains you already have configured, you need to use the `--override` flag in the relayer. 

This PR adds the ability to take advantage this `--override` flag with interchaintest.

Note that this is only implemented for the go relayer.


Reecepbcups (approved):

Reecepbcups:
If we need to expand this further, I almost feel we should just have an `ExtraFlags []string` to ensure it can be expanded easily without modification to the core type

PR #1257, opened 2024-09-14
refactor: remove blocksdb

Reecepbcups:
This is a test PR to see if we can / should remove this.

For the FindTxs() commands removed for thor/cosmos, if needed we should just query consensus at a given height, then query the tx events from a given txhash. if this is all we need it for we can reduce a TON of complexity here.


chain/cosmos/chain_node.go
```diff
@@ -427,82 +423,6 @@ func (tn *ChainNode) Height(ctx context.Context) (int64, error) {
 	return height, nil
 }
 
-// FindTxs implements blockdb.BlockSaver.
-func (tn *ChainNode) FindTxs(ctx context.Context, height int64) ([]blockdb.Tx, error) {
```
Reecepbcups, line 431: refactor this to query cometbft for a given block, get the Txs, base64 decode, and query each tx hash from consensus directly to get the EventAttributes. Then query the block to get the FinalizeBlockEvents.

While this is less efficient, it removes a TON of code for a function that does not seem to be called hardly ever

chain/thorchain/thornode.go
```diff
@@ -404,82 +400,6 @@ func (tn *ChainNode) Height(ctx context.Context) (int64, error) {
 	return height, nil
 }
 
-// FindTxs implements blockdb.BlockSaver.
-func (tn *ChainNode) FindTxs(ctx context.Context, height int64) ([]blockdb.Tx, error) {
```
Reecepbcups, line 408: Thorchain, ref chain_node: https://github.com/strangelove-ventures/interchaintest/pull/1257/files#r1759640964

agouin:
The blocks db was instrumental in development of relayer v2, among other things. It is a lightweight indexer and easily allows you to inspect transaction lifecycle on involved chains. There is also the terminal UI built on top of it for browsing the indexed data.

The usage is much more for developers to see sequence of events of blocks, transactions, and events, rather than any programmatic usage.

Issue #1238, opened 2024-08-30
ICT should give more descriptive errors when chain genesis validation fails


dowlandaiello:
It would be helpful if local-ic was able to report more detailed errors when, for example, a parameter in the chain genesis JSON is incorrectly spelled or misplaced. Currently, this results in a bizarre, ambiguous error unrelated to the genesis at all, and the only way to determine the source of the error is by looking through the killed docker container's logs.

Example incorrect chain genesis param:

```
{
          "key": "app_state.gov.params.deposit_period",
          "value": "3s
},
```

Example current error:
```
2024-08-30T17:03:43.787Z        FATAL   interchain/start.go:176 ic.Build        {"error": "failed to start chains: failed to start chain localosmosis-1: All attempts fail:\n#1: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused\n#2: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused\n#3: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused\n#4: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused\n#5: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused\n#6: post failed: Post \"http://0.0.0.0:26659\": ..................
```

This is not necessarily a bug or anything, but it would be nice if the error from inside the docker container could be bubbled up to local-ic somehow!

Reecepbcups:
Agree!

Currently Interchaintest does read / parse panics in a proper way. I had an issue on this, unsure where it went. Once a panic parser is in ICT this will be automatically added to local-ic as well due to how it is designed.

Reecepbcups:
PR is up and pending review from our side. It will parse out any panics on startup, return this error, show all the logs, and cleanup any containers hanging.

![Image](https://github.com/user-attachments/assets/a7ce2c80-3acc-4bbe-8030-2da9eb6c5117)

PR #30, opened 2022-04-20
feat: adds the skeleton of the 6 out of 9 partition

bmorphism:

PR #1258, opened 2024-09-16
feat: check for panic in logs on start

Reecepbcups:
closes https://github.com/strangelove-ventures/interchaintest/issues/1238

## Summary

This is a huge user experience improvement for ICT & local-ic upon a bad genesis param set (or some other failed startup option). 

Before, the network would hang and just state it failed to query
:
failed to start chains: failed to start chain localosmosis-1: All attempts fail:
`#1`: post failed: Post \"http://0.0.0.0:26659\": dial tcp 0.0.0.0:26659: connect: connection refused
...
`#30` ...

## Showcase

This chain uses a genesis modification of `{"key": "app_state.gov.params.voting_period", "value": "bad"},`. Both validators show this error in an easy to read format + the docker logs for the stacktrace. 

image

### DIFF
--- a/dockerutil/container_lifecycle.go
+++ b/dockerutil/container_lifecycle.go
@@ -130,7 +130,7 @@ func (c *ContainerLifecycle) StartContainer(ctx context.Context) error {
 		return err
 	}
 
-	if err := c.CheckForFailedStart(ctx, time.Second*1); err != nil {
+	if err := c.CheckForFailedStart(ctx, time.Second*3); err != nil {
 		return err
 	}
 
--- a/dockerutil/container_lifecycle.go
+++ b/dockerutil/container_lifecycle.go
@@ -3,8 +3,11 @@ package dockerutil
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
+	"regexp"
 	"strings"
+	"time"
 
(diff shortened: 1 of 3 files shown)


joelsmith-2019 (approved):

joelsmith-2019:
Will this slow down tests that spin up numerous containers?

Reecepbcups:
@joelsmith-2019 

> Will this slow down tests that spin up numerous containers? 
Test are only slowed down by the 1 second per container, and containers are spun up in parallel. So if you have 1 or 100 containers, the test time only slows down by 1 second

>  just want to double check we won't wait 1 second for each ephemeral container we spin up
No, ephemeral containers use their own job.Run() method which is outside of the containerLifecycle.  containerLifecycle is only used for long running containers (nodes & sidecars).

I had previously tested this with
```
	if err := c.CheckForFailedStart(ctx, time.Second*10); err != nil {
		return err
	}
```
Only time it hangs is post "Starting container" once.

Reecepbcups:
I also have a WIP speed up branch which will reduce test time by like 40%. Theoretical max is ~92% with state cache. Which would make this 1 second start wait the slowest part of ICT in the future

https://github.com/strangelove-ventures/interchaintest/pull/1208

PR #1269, opened 2024-09-26
feat(local-ic): stream interaction and container logs

Reecepbcups:
## Summary

For the Spawn UI platform, we need a way to get logs from the network -> the users UI. This PR adds support to stream the logs to both the console & a unique file. Then the user can query /logs in the REST api. It will tail the last N lines, and continue to return new data on new interaction.

Further, the container logs for actuals of a container are streamed too. You can use the container ID or the node name itself. much better than `docker ps; docker logs <id> -f` for API users.

Based on concept here: https://github.com/Reecepbcups/golang_stream

## TODO
- [x] stream docker container logs out as well? (i.e. /container_logs?id=<id>)
- [x] Remove dumb comments
- [x] Print tail X lines of logs endpoint (instead of trying to merge them)
- [x] Select logs by chain-id vs being the container ID (some mapping between)


local-interchain/interchain/handlers/container_log_stream.go
```diff
@@ -0,0 +1,153 @@
...
+	"strings"
+	"unicode"
+
+	dockertypes "github.com/docker/docker/api/types"
+	dockerclient "github.com/docker/docker/client"
+	"github.com/strangelove-ventures/interchaintest/v8/chain/cosmos"
+	"go.uber.org/zap"
+)
+
+var removeColorRegex = regexp.MustCompile("\x1b\\[[0-9;]*m")
```
jtieri, line 18: Is this necessary because certain operating systems or terminals cannot render the colors in the logs?

local-interchain/interchain/handlers/container_log_stream.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Flush ensures data is sent to the client immediately
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		http.Error(w, "Streaming unsupported", http.StatusInternalServerError)
+		return
+	}
+
+	for {
+		buf := make([]byte, 8*1024)
```
jtieri, line 105: What was the reasoning for instantiating the slice with this specific value? Does changing this value break anything? If so a comment regarding the reason for this specific size could be helpful.

local-interchain/interchain/handlers/container_log_stream.go
```diff
@@ -0,0 +1,153 @@
...
+	dockertypes "github.com/docker/docker/api/types"
+	dockerclient "github.com/docker/docker/client"
+	"github.com/strangelove-ventures/interchaintest/v8/chain/cosmos"
+	"go.uber.org/zap"
+)
+
+var removeColorRegex = regexp.MustCompile("\x1b\\[[0-9;]*m")
+
+type ContainerStream struct {
+	ctx      context.Context
```
jtieri, line 21: Is there a specific reason we have to use a Context as a field member of this struct vs. piping a Context into the `StreamContainer` method or wherever a Context is needed for the `ContainerStream`?

It's typically more [idiomatic to pass a Context around vs. storing one inside of a struct type](https://go.dev/blog/context-and-structs)

One of the exceptions to this rule are typically when you have an external request that starts a background process/operation. 

Edit: After reading the rest of the code it seems like that is what may be happening here when a user makes a request to `/container_logs` so this may be a correct usage of context in a struct but figured I would ask for clarity.

jtieri (commented):
Had a couple nits around logs and a couple questions but otherwise the changes look good.

Although it is out of scope of this PR one thing that jumped out to me is that the `StartChain` function in `start.go` is growing to be rather large. There are a few places where logic could be broken out into separate functions.

Dockerfile.local-interchain
```diff
@@ -21,8 +21,16 @@ RUN cd local-interchain && make build
...
+# Final stage
+FROM debian:bookworm-slim AS final
+
+# Install certificates and required libraries
+RUN apt-get update && \
+    apt-get install -y ca-certificates libc6 && \
+    update-ca-certificates && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
```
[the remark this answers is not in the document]
Reecepbcups: from ref: https://github.com/strangelove-ventures/interchaintest/pull/1284 / issue https://github.com/strangelove-ventures/interchaintest/issues/1282

> Local_IC in docker results in a tls; 509 cert error. The user must install ca-cetificates. If we get this error tls: failed > to verify certificate: x509: certificate signed by unknown authority, sudo apt-get install -y ca-certificates

The original docker instance had this. I fixed in spawn (which wraps local-ic) but forgot to add it here. Required for Galen to get the SpawnUI to work with the backend infra

local-interchain/interchain/start.go
```diff
@@ -53,11 +54,27 @@ func StartChain(installDir, chainCfgFile string, ac *types.AppStartConfig) {
...
+	}
+	defer func() {
+		if err := logFile.Close(); err != nil {
+			fmt.Println("Error closing log file: ", err)
+		}
+
+		if err := os.Remove(logFile.Name()); err != nil {
+			fmt.Println("Error deleting log file: ", err)
+		}
+	}()
```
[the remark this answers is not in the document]
Reecepbcups: This log file really is not useful except for when interacting at runtime. i.e. I send command, it logs some info. The actual information that would be valuable we don't log yet (docker container appd processes). In the future I may add this for the Spawn infra, but for now it's just runtime viewable.

So this is fine as is unless I add on more in the future

local-interchain/interchain/handlers/container_log_stream.go
Reecepbcups: Since this is being forwarded over a REST api, it is always the case. So most users parsing that data don't care about. We may in the future want to parse out the colors which is why it is possible to opt out and see those ASNI escape sequences

local-interchain/interchain/handlers/container_log_stream.go
Reecepbcups: 1024 is the standard but due to the size of some JSON blobs sent through it was exceeding that for Galen and he was unable to properly parse (as it came in multiple chunks). 

It was exceeding the size for some relayer JSON blobs for Galen. He was unable to parse it since it came in multiple chunks. I wanted to ensure we have enough of a buffer and 8192 felt like a good spot (2*1024 was still too small, I am sure 4* would have been fine but increased just to be sure since the data sent it relative small and controlled anyways)

local-interchain/interchain/handlers/container_log_stream.go
Reecepbcups: tldr; better Dev UX for the same relative solution

It's cleaner than having to write methods return `func(w http.ResponseWriter, req *http.Request) ` signatures just for ctx to be passed in in a wrapped type.

Current:
```go
r.HandleFunc("/container_logs", containerStream.StreamContainer)
```

Would be required change
```go
r.HandleFunc("/container_logs", containerStream.handleStreamContainer(ctx))

func (cs *ContainerStream) handleStreamContainer(ctx context) func(w http.ResponseWriter, req *http.Request) {

   return func(w http.ResponseWriter, req *http.Request) {
      // all the same logic here, just wrapped now
    }
}
```

ctx is only used for timeout limits and set on startup. So any runtime changes are unaffected in either approach - and the first approach is just net easier for a new contributor to wrap their head around.

local-interchain/interchain/start.go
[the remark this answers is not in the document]
jtieri: Makes sense to me!

local-interchain/interchain/start.go
```diff
@@ -98,7 +115,7 @@ func StartChain(installDir, chainCfgFile string, ac *types.AppStartConfig) {
 
 	chains, err := cf.Chains(testName)
 	if err != nil {
-		log.Fatal("cf.Chains", err)
+		logger.Fatal("cf.Chains", zap.Error(err))
```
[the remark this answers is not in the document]
jtieri: It would be awesome to find a way to enforce our logging style guide via the linter. I'm sure the codebase has many inconsistencies at this point from people not remembering about the style guide during code review but figured i would point it out since we have been putting some effort into consistency across the codebase.

local-interchain/interchain/handlers/container_log_stream.go
jtieri: Not necessary but this is the type of decision that can be helpful to document via a comment. You never know when someone is going to come across and naively change values without understanding what the implications are.

jtieri (approved):

PR #1270, opened 2024-09-29
chore: fix linter warnings

jtieri:
This PR updates the CI jobs for linting and running tests to use current deps. It also fixes all the linter warnings from `golangci-lint`.

PR #1281, opened 2024-10-02
fix(thorchain/utxo): fix race conditions

misko9:
Fixes a few race conditions on thorchain and utxo chains.

Thorchain: sidecar processes now get their own copy of env list so they can make runtime changes in parallel.

Utxo: reading and writing KeynameToWalletMap and AddrToKeynameMap is now protected by a mutex. The block producing go routine is now ran after faucet creation and a utxo stop method is added for the test framework to cancel its context.

Thorchain's setup_test.go is updated to call the utxo chain's stop method. This will also be reflected in the thornode v50 branch.

### DIFF
--- a/examples/utxo/start_test.go
+++ b/examples/utxo/start_test.go
@@ -56,6 +56,12 @@ func TestUtxo(t *testing.T) {
 	}))
 	t.Cleanup(func() {
 		_ = ic.Close()
+		for _, chain := range chains {
+			utxoChain, ok := chain.(*utxo.UtxoChain)
+			if ok {
+				utxoChain.Stop()
+			}
+		}
 	})
 
 	// Create and fund a user using GetAndFundTestUsers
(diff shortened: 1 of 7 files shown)


jtieri:
I wanted to give you a heads up, https://github.com/strangelove-ventures/interchaintest/pull/1270 introduces some changes that are going to affect the thorchain/utxo stuff you are working on. I just finished linting the codebase and re-adding the linter as a required status check in CI and one major change I needed to make was passing in a `context` to the Thorchain API calls so that context cancellation is respected

examples/utxo/start_test.go
```diff
@@ -56,6 +56,12 @@ func TestUtxo(t *testing.T) {
 	}))
 	t.Cleanup(func() {
 		_ = ic.Close()
+		for _, chain := range chains {
+			utxoChain, ok := chain.(*utxo.UtxoChain)
+			if ok {
+				utxoChain.Stop()
+			}
+		}
```
Reecepbcups, line 64: Not for this PR, but if this always needs to be called it may make sense to move into the ic.Close() func?

Reecepbcups (approved):

PR #1283, opened 2024-10-12
fix: do not require a directory for remote urls

### LINKED ISSUE #1282: [Unclassified]: fix local-ic tls error & required chains/ dir
## Error

Local_IC in docker results in a tls; 509 cert error. The user must install ca-cetificates. If we get this error `tls: failed to verify certificate: x509: certificate signed by unknown authority`, have them:

`sudo apt-get install -y ca-certificates`

## Chains dir

Even when using a link, local-ic still requires a valid chain dir.

### Desire
docker run -it local-interchain:local local-ic start https://gist.githubusercontent.com/Reecepbcups/70bf59c82c797ead9a5430b8b9d8d852/raw/cecc7be35bcec8b976a5d92e78fd6d56de2e1aa1/cosmoshub_localic_config.json

### Current
docker run -it -v $(pwd)/local-interchain/chains:/app/local-interchain/chains local-interchain:local local-ic start https://gist.githubusercontent.com/Reecepbcups/70bf59c82c797ead9a5430b8b9d8d852/raw/cecc7be35bcec8b976a5d92e78fd6d56de2e1aa1/cosmoshub_localic_config.json

### Reason

If using link, don't require any chains/ location for files. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Reecepbcups:
Closes #1282 

## Summary

If a URL is used instead of a base file (filename or filename.(json|yaml), it will not call GetDirectory() as it is not needed. This previously caused issues for the SpawnUI as we only use remote github links and it required us to mount a volume for an empty chains/ directory.

## logical change

If URL, don't get directory. Else do get parent directory


local-interchain/cmd/local-ic/start_chain.go
```diff
@@ -36,7 +36,11 @@ local-ic start https://pastebin.com/raw/Ummk4DTM
 	Args: cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		configPath := args[0]
-		parentDir := GetDirectory()
+		isURL := strings.HasPrefix(configPath, "http")
+
+		var parentDir string
+		var config *types.Config
+		var err error
```
jtieri, line 43:
```suggestion
		var (
		    parentDir string
		    config *types.Config
		    err error
	         )
```

jtieri (approved):
lgtm

PR #1284, opened 2024-10-16
fix: cors and sse for logs

frechg:
- Updates local-interchain Docker to provide ssl certificates allowing testnet config json to be loaded remotely
- Fixes logs server CORS errors
- Fixes Server Sent Events formatting errors


local-interchain/interchain/handlers/log_stream.go
```diff
@@ -79,7 +79,7 @@ func (ls *LogStream) StreamLogs(w http.ResponseWriter, r *http.Request) {
 			line, err := reader.ReadString('\n')
 			if err == nil {
 				// Send the log line to the client
-				fmt.Fprintf(w, "%s\n", line)
+				fmt.Fprintf(w, "data: %s\n\n", line)
```
Reecepbcups, line 82: assume this is for pretty parsing client side?

Reecepbcups (approved):

local-interchain/interchain/handlers/log_stream.go
frechg: Yeah, the data key and double new line at the end is proper server sent event format. You can get the event stream and manually handle it, but it won't work with the EventSource interface if it's not in that format. https://developer.mozilla.org/en-US/docs/Web/API/EventSource

PR #1306, opened 2024-11-29
feat: ics v6 support

Reecepbcups:
## Summary

Support ICS v6

## TODO:
- Rename the previous Addition to "Legacy", then create a new function for permissionless adds instead. This way it's not state breaking, just API


reecepbcups:
Technically works, but won't get it. using it upstream in Prysm

PR #1327, opened 2025-01-28
feat: bump ibc-go to v9.0.2

mmsqe:
bump ibc-go to v9.x to support e2e test in https://github.com/cosmos/relayer/pull/1530


reecepbcups:
We never planned to support v9, and since then the IBC team realized no one was integrating it and yanked it as well.

PR #1352, opened 2025-04-04
feat: add cardano chain

kocubinski:
Adds a cardano implementation for interchain tests.

For the go version bump, github.com/blinklabs-io/gouroboros (used for cardano node communication) requires go >= 1.23.6.


examples/cardano/start_test.go
```diff
@@ -0,0 +1,94 @@
...
+	ctx := context.Background()
+
+	cf := interchaintest.NewBuiltinChainFactory(zaptest.NewLogger(t), []*interchaintest.ChainSpec{
+		{ChainConfig: ada.DefaultConfig("cardano")},
+	})
+
+	chains, err := cf.Chains(t.Name())
+	require.NoError(t, err)
+	adaChain := chains[0].(*ada.AdaChain)
+	err = adaChain.SetFaucet(ctx)
```
reecepbcups, line 34: UX nit: this (SetFaucet) likely makes more sense within the Build(...), similar to what we do for [`CreatePenumbraClient` if chain is of that type](https://github.com/strangelove-ventures/interchaintest/blob/137503860844eded95712df7cb26577047f772b1/interchain.go#L354).

Else this step would be fairly easy to miss since integrator may expect the faucet to already be setup. not a blocker

reecepbcups (approved):
utAck, that ictest bash CI & markdown check CI do not need to pass but the lint should

PR #1353, opened 2025-04-16
Remove unused CrisisInvariantBroken method

VolodymyrBg:
Remove the unused CrisisInvariantBroken method from module_crisis.go

- The CrisisInvariantBroken method is not called anywhere in the codebase
- Keeping only the package declaration in module_crisis.go for backward compatibility
- This change reduces code maintenance burden without affecting functionality
- Improved code cleanliness by removing unused code


reecepbcups:
its a helper method for external codebases.

